PR #65, opened 2019-02-08
zkvm/vm: add witness fields and assignments for cloak variables

vickiniu:
Uses the `SignedInteger` type from spacesuit and fills in the witness and assignment fields for cloak variables.

### DIFF
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -1,10 +1,10 @@
 //! Core ZkVM stack types: data, variables, values, contracts etc.
 
 use bulletproofs::{r1cs, PedersenGens};
-use byteorder::{ByteOrder, LittleEndian};
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
 use merlin::Transcript;
+use spacesuit::SignedInteger;
 
 use crate::encoding::Subslice;
 use crate::errors::VMError;
@@ -29,19 +29,10 @@ pub enum PortableItem {
     Value(Value),
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum Data {
     Opaque(Vec<u8>),
     Witness(DataWitness),
-}
-
-impl Data {
-    pub fn to_bytes(self) -> Vec<u8> {
-        match self {
-            Data::Opaque(data) => data,
-            Data::Witness(_) => unimplemented!(),
-        }
-    }
 }
 
 #[derive(Debug)]
@@ -73,7 +64,7 @@ pub struct Variable {
 pub struct Expression {
     /// Terms of the expression
     pub(crate) terms: Vec<(r1cs::Variable, Scalar)>,
-    pub(crate) assignment: Option<u64>,
+    pub(crate) assignment: Option<SignedInteger>,
 }
 
 #[derive(Clone, Debug)]
@@ -85,7 +76,7 @@ pub enum Constraint {
     // this also allows us not to wrap this enum in a struct.
 }
 
-#[derive(Debug)]
+#[derive(Clone,Debug)]
 pub enum Predicate {
     Opaque(CompressedRistretto),
     Witness(Box<PredicateWitness>),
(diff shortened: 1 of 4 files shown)


zkvm/src/types.rs
```diff
@@ -140,6 +142,12 @@ pub struct FrozenValue {
     pub(crate) flv: Commitment,
 }
 
+#[derive(Copy, Clone, Debug)]
```
```suggestion
/// Represents a concrete kind of a number represented by a scalar:
/// `ScalarKind::Integer` represents a signed integer with 64-bit absolute value (aka i65)
/// `ScalarKind::Scalar` represents a scalar modulo group order.
#[derive(Copy, Clone, Debug)]
```

zkvm/src/types.rs
```diff
@@ -63,6 +64,7 @@ pub struct Variable {
 pub struct Expression {
     /// Terms of the expression
     pub(crate) terms: Vec<(r1cs::Variable, Scalar)>,
+    pub(crate) assignment: Option<SignedInteger>,
```
oleganza: Probably we need to allow Scalar assignments too - in case constraints are done over the whole group order |G|, not just integers. 

```suggestion
    pub(crate) assignment: Option<ScalarKind>,
```

zkvm/src/types.rs
```diff
@@ -140,6 +142,12 @@ pub struct FrozenValue {
     pub(crate) flv: Commitment,
 }
 
+#[derive(Copy, Clone, Debug)]
+pub enum ScalarKind {
```
oleganza: `ScalarKind` sounds meta to me: like, a type of a type. Maybe `ScalarWitness` is more apt and consistent with the rest of the code? It's a "real witness value for the given scalar representation".

```suggestion
pub enum ScalarWitness {
```

zkvm/src/vm.rs
```diff
@@ -537,18 +548,40 @@ where
...
+    fn variable_to_expression(&mut self, var: Variable) -> Result<Expression, VMError> {
         let (_, r1cs_var) = self.attach_variable(var);
-        Expression {
+        let expr = Expression {
             terms: vec![(r1cs_var, Scalar::one())],
+            assignment: match self.variable_assignment(var) {
+                None => None,
+                Some(v) => match v {
+                    ScalarKind::Integer(i) => Some(i),
+                    ScalarKind::Scalar(_) => return Err(VMError::FormatError),
```
oleganza: Seems like we'll be using `variable_to_expression` in other contexts, than range proofs, too and should respect scalars as well as signed  integers for assignments here.

Then, the whole method would never fail.

zkvm/src/vm.rs
```diff
@@ -644,8 +677,7 @@ where
         spacesuit::range_proof(
             self.delegate.cs(),
             r1cs::LinearCombination::from_iter(expr.terms),
-            // TBD: maintain the assignment for the expression and provide it here
-            None,
+            expr.assignment,
```
oleganza: If assignment is `Option<ScalarKind>` (aka `Number` aka `ScalarWitness`), then this can be changed using `Option.map`:

```suggestion
            match expr.assignment {
                None => None,
                Some(ScalarKind::Integer(i)) => Some(i),
                Some(ScalarKind::Scalar(_)) => return Err(VMError::InconsistentAssignment),
            }
```

zkvm/src/vm.rs
```diff
@@ -537,18 +548,40 @@ where
...
             Item::WideValue(w) => Ok(w),
             _ => Err(VMError::TypeNotWideValue),
         }
     }
 
-    fn variable_to_expression(&mut self, var: Variable) -> Expression {
+    fn variable_to_expression(&mut self, var: Variable) -> Result<Expression, VMError> {
         let (_, r1cs_var) = self.attach_variable(var);
-        Expression {
+        let expr = Expression {
```
oleganza: instead of `let expr = ...; Ok(expr)` you can simply wrap the whole thing in `Ok`:

```rust
Ok(Expression{
   ...
})
```

zkvm/src/vm.rs
```diff
@@ -537,18 +548,40 @@ where
...
                 r1cs_flv: self.attach_variable(value.flv).1,
-                // TBD: add witness for Value types where it exists.
-                witness: None,
+                witness: match (
+                    self.variable_assignment(value.qty),
+                    self.variable_assignment(value.flv),
+                ) {
+                    (Some(ScalarKind::Integer(q)), Some(ScalarKind::Scalar(f))) => Some((q, f)),
+                    (None, None) => None,
+                    (_, _) => return Err(VMError::FormatError),
```
oleganza: AFAIR, there are a few places where assignments may be inconsistent. Would it make sense to introduce VMError::InconsistentWitness?

PR #67, opened 2019-02-08
zkvm/prover: implement traits and build_tx for prover

vickiniu:
Implements the `Delegate` and `RunTrait` traits needed by the VM for the prover. Also implements the `build_tx` function, minus the code to actually encode a prover's `Instruction` into tx bytecode for the verifier.

### DIFF
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -1,3 +1,4 @@
+use core::borrow::Borrow;
 use core::mem;
 
 use crate::encoding::Subslice;
@@ -194,5 +195,15 @@ impl Instruction {
     /// to the program.
     pub fn encode(&self, program: &mut Vec<u8>) {
         unimplemented!()
+    }
+
+    pub fn encode_program<I>(iterator: I, program: &mut Vec<u8>)
+    where
+        I: IntoIterator,
+        I::Item: Borrow<Self>,
+    {
+        for i in iterator.into_iter() {
+            i.borrow().encode(program);
+        }
     }
 }
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -190,8 +190,9 @@ impl Instruction {
         }
     }
 
-    /// Returns the bytecode representation of an Instruction.
-    pub fn encode(&self) -> Vec<u8> {
+    /// Appends the bytecode representation of an Instruction
+    /// to the program.
+    pub fn encode(&self, program: &mut Vec<u8>) {
         unimplemented!()
     }
 }
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -107,11 +107,13 @@ impl Opcode {
 }
 
 impl Instruction {
-    /// Returns a parsed instruction with a size that it occupies in the program string.
-    /// E.g. a push instruction with 5-byte string occupies 1+4+5=10 bytes
-    /// (4 for the LE32 length prefix).
+    /// Returns a parsed instruction from a subslice of the program string, modifying
+    /// the subslice according to the bytes the instruction occupies
+    /// E.g. a push instruction with 5-byte string occupies 1+4+5=10 bytes,
+    /// (4 for the LE32 length prefix), advancing the program subslice by 10 bytes.
     ///
-    /// Return `VMError::FormatError` if there is not enough bytes to parse an instruction.
+    /// Return `VMError::FormatError` if there are not enough bytes to parse an
+    /// instruction.
     pub fn parse(program: &mut Subslice) -> Result<Self, VMError> {
         let byte = program.read_u8()?;
 
@@ -186,5 +188,10 @@ impl Instruction {
             Opcode::Right => Ok(Instruction::Right),
             Opcode::Delegate => Ok(Instruction::Delegate),
         }
+    }
+
+    /// Returns the bytecode representation of an Instruction.
+    pub fn encode(&self) -> Vec<u8> {
+        unimplemented!()
     }
 }
--- a/zkvm/src/prover.rs
+++ b/zkvm/src/prover.rs
@@ -16,12 +16,10 @@ use crate::vm::{Delegate, Tx, VM};
 pub struct Prover<'a, 'b> {
     signtx_keys: Vec<Scalar>,
     cs: r1cs::Prover<'a, 'b>,
-    bytecode: Vec<u8>,
 }
 
 pub struct ProverRun {
     program: VecDeque<Instruction>,
-    root: bool,
 }
 
 impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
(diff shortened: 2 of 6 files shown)


zkvm/src/ops.rs
```diff
@@ -187,4 +189,9 @@ impl Instruction {
             Opcode::Delegate => Ok(Instruction::Delegate),
         }
     }
+
+    /// Returns the bytecode representation of an Instruction.
+    pub fn encode(&self) -> Vec<u8> {
```
oleganza: Maybe make it like `write_` APIs, write by appending to the `&mut Vec<u8>` provided as an argument?

zkvm/src/prover.rs
```diff
@@ -1,25 +1,44 @@
...
-use crate::vm::{Delegate, RunTrait};
+use crate::vm::{Delegate, RunTrait, Tx, VM};
 
 pub struct Prover<'a, 'b> {
     signtx_keys: Vec<Scalar>,
     cs: r1cs::Prover<'a, 'b>,
 }
 
-pub struct RunProver {}
+pub struct RunProver {
```
oleganza: Maybe `RunProver`, `RunVerifier` -> `ProverRun`, `VerifierRun`?

zkvm/src/prover.rs
```diff
@@ -1,25 +1,44 @@
...
 
-    fn commit_variable(&mut self, com: &Commitment) -> (CompressedRistretto, r1cs::Variable) {
-        unimplemented!()
+    fn commit_variable(
+        &mut self,
+        com: &Commitment,
+    ) -> Result<(CompressedRistretto, r1cs::Variable), VMError> {
+        let (v, v_blinding) = match com {
+            Commitment::Open(w) => {
+                let val = match w.value {
```
oleganza: Can use `w.value.into()` here, since ScalarWitness already knows how to convert to Scalar.

zkvm/src/prover.rs
```diff
@@ -1,25 +1,38 @@
...
+        com: &Commitment,
+    ) -> Result<(CompressedRistretto, r1cs::Variable), VMError> {
+        let (v, v_blinding) = match com {
+            Commitment::Open(w) => (w.value.into(), w.blinding),
+            Commitment::Closed(_) => return Err(VMError::WitnessMissing),
+        };
+        Ok(self.cs.commit(v, v_blinding))
     }
 
     fn verify_point_op<F>(&mut self, point_op_fn: F)
```
oleganza: how about making this failable (and closure failable too), so that:

1. Verifier can decide whether to do batching or check and fail early (for diagnostics)
2. Prover can decide to double-check its consistency too.

We do not need to provide these choices now, but we can plug them later easily if this method is failable.

vickiniu: that makes sense to me! do we want to leave it as a no-op on the prover right now, or have the closure returnan `Option` and check on the prover if there is a PointOp

zkvm/src/prover.rs
```diff
@@ -44,8 +57,68 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
...
+            ProverRun::new(&program),
+            &mut prover,
+        );
+
+        let (txid, _) = vm.run()?;
+
+        // Sign txid
+        let mut signtx_transcript = Transcript::new(b"ZkVM.signtx");
+        signtx_transcript.commit_bytes(b"txid", &txid.0);
+        let signature = Signature::sign_aggregated(&mut signtx_transcript, &prover.signtx_keys);
```
oleganza: :mindblown:

zkvm/src/prover.rs
```diff
@@ -44,8 +57,68 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
...
+            proof,
+            program: bytecode,
+        })
+    }
+}
+
+impl ProverRun {
+    fn new(program: &Vec<Instruction>) -> Self {
+        ProverRun {
+            program: program.clone().into(),
```
oleganza: Can we design it, so that instead of having two full copies of programs: one for consumption by VM, another for converting into bytecode, we consume the stream of `Instruction` and simultaneously encode them into a bytecode buffer?

We can do so with a hook like `delegate.will_execute_instruction(&Instruction)`.

vickiniu: sounds good to me! would that fit best inside the call to `vm::´step` before we enter the big match block?

vickiniu: or I imagine we could serialize the instruction when we call `next_instruction` on the prover?

zkvm/src/prover.rs
oleganza: Let's have both the method and the closure return `Result<(), VMError>`. Prover would do nothing (`Ok(())`) and verifier will batch.

oleganza: Oh, wait. Probably the closure will never fail. It just returns a PointOp.

oleganza: (edited)

zkvm/src/verifier.rs
```diff
@@ -10,33 +10,40 @@ use crate::point_ops::PointOp;
...
-        let point = com.to_point();
-        let var = self.cs.commit(point);
-        (point, var)
+    type RunType = VerifierRun;
+
+    fn commit_variable(
+        &mut self,
+        com: &Commitment,
+    ) -> Result<(CompressedRistretto, r1cs::Variable), VMError> {
+        let point = match com {
```
oleganza: don't we have `to_point` method for commitment?

vickiniu: we do! but to mirror the prover, I wanted to fail if the commitment is `Open`

zkvm/src/prover.rs
```diff
@@ -39,13 +54,83 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
...
+            maxtime,
+            signature,
+            proof,
+            program: prover.bytecode,
+        })
+    }
+}
+
+impl ProverRun {
+    fn root(program: Vec<Instruction>) -> Self {
```
oleganza: maybe s/root/from_txprogram/ and s/subprogram/from_subprogram/?

zkvm/src/prover.rs
```diff
@@ -39,13 +54,83 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
...
+
+        let vm = VM::new(
+            version,
+            mintime,
+            maxtime,
+            ProverRun::from_txprogram(program),
+            &mut prover,
+        );
+
+        let (txid, _) = vm.run()?;
```
oleganza: Maybe we should return txlog together with `Tx`, for convenience? Then the prover can avoiding passing the Tx through the Verifier to get it.

`build_tx() -> Result<(Tx,Vec<txlog::Entry>), VMError>`

Issue #61, opened 2019-02-07
ZkVM: API layering proposal


oleganza:
### Low-level API

* Takes a stream of structured instructions bearing _witness data_ (secret values, secret blinding factors, private keys, etc.).
* Passes that stream through a generic VM implementation that performs the same stack operations as the verifier, and constructs a transaction log and a constraint system.
* Generates a `Tx` object that contains:
    * serialized program bytecode,
    * aggregated signature over txid (see [`signtx`](https://github.com/interstellar/slingshot/blob/main/zkvm/spec/ZkVM.md#signtx) instruction)
    * R1CS proof for the constraint system

### High-level API

This is defined in terms of particular _actions_: issue funds, spend funds, etc. This itself may be layered, like key derivation and utxo selection for _accounts_ could be a separate layer on top of more agnostic spend/send API.

Here's a sketch of things we want to implement at this level:

1. Asset issuance API (issue/retire)
2. Account API (spend/send)
3. Pegging (import/export)
4. Vault protocols (multi-key storage + timelock)
5. Payment channels and payment routing (Lightning)

oleganza:
This plan is largely implemented, so i'm closing this issue.

PR #69, opened 2019-02-09
zkvm/prover: serialize instructions to bytecode

vickiniu:
Implements part of #61, to serialize the Prover's vec of Instructions into bytecode for the Verifier.

WIP: a few pieces missing around how best to implement Data `to_bytes` and `len`, as well as how to get the previous txID needed for `Input`

### DIFF
--- a/zkvm/src/encoding.rs
+++ b/zkvm/src/encoding.rs
@@ -124,6 +124,13 @@ pub fn write_u32<'a>(x: u32, target: &mut Vec<u8>) {
     target.extend_from_slice(&buf);
 }
 
+// Writes a LE64-encoded integer
+pub fn write_u64<'a>(x: u64, target: &mut Vec<u8>) {
+    let mut buf = [0u8; 8];
+    LittleEndian::write_u64(&mut buf, x);
+    target.extend_from_slice(&buf);
+}
+
 /// Reads a 32-byte array and returns the subsequent slice
 pub fn write_bytes(x: &[u8], target: &mut Vec<u8>) {
     target.extend_from_slice(&x);
--- a/zkvm/src/encoding.rs
+++ b/zkvm/src/encoding.rs
@@ -124,13 +124,6 @@ pub fn write_u32<'a>(x: u32, target: &mut Vec<u8>) {
     target.extend_from_slice(&buf);
 }
 
-// Writes a LE64-encoded integer
-pub fn write_u64<'a>(x: u64, target: &mut Vec<u8>) {
-    let mut buf = [0u8; 8];
-    LittleEndian::write_u64(&mut buf, x);
-    target.extend_from_slice(&buf);
-}
-
 /// Reads a 32-byte array and returns the subsequent slice
 pub fn write_bytes(x: &[u8], target: &mut Vec<u8>) {
     target.extend_from_slice(&x);
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -1,6 +1,7 @@
 use core::borrow::Borrow;
 use core::mem;
 
+use crate::encoding;
 use crate::encoding::Subslice;
 use crate::errors::VMError;
 use crate::types::Data;
@@ -194,7 +195,69 @@ impl Instruction {
     /// Appends the bytecode representation of an Instruction
     /// to the program.
     pub fn encode(&self, program: &mut Vec<u8>) {
-        unimplemented!()
+        match self {
+            Instruction::Push(data) => {
+                program.push(Opcode::Push.to_u8());
+                data.encode(program);
+            }
+            Instruction::Drop => program.push(Opcode::Drop.to_u8()),
+            Instruction::Dup(idx) => {
+                program.push(Opcode::Dup.to_u8());
+                encoding::write_u32(*idx as u32, program);
+            }
+            Instruction::Roll(idx) => {
+                program.push(Opcode::Roll.to_u8());
+                encoding::write_u32(*idx as u32, program);
+            }
+            Instruction::Const => program.push(Opcode::Const.to_u8()),
+            Instruction::Var => program.push(Opcode::Var.to_u8()),
+            Instruction::Alloc => program.push(Opcode::Alloc.to_u8()),
+            Instruction::Mintime => program.push(Opcode::Mintime.to_u8()),
+            Instruction::Maxtime => program.push(Opcode::Maxtime.to_u8()),
+            Instruction::Neg => program.push(Opcode::Neg.to_u8()),
+            Instruction::Add => program.push(Opcode::Add.to_u8()),
+            Instruction::Mul => program.push(Opcode::Mul.to_u8()),
+            Instruction::Eq => program.push(Opcode::Eq.to_u8()),
+            Instruction::Range(bit_width) => {
+                program.push(Opcode::Range.to_u8());
+                program.push(*bit_width);
+            }
+            Instruction::And => program.push(Opcode::And.to_u8()),
+            Instruction::Or => program.push(Opcode::Or.to_u8()),
+            Instruction::Verify => program.push(Opcode::Verify.to_u8()),
+            Instruction::Blind => program.push(Opcode::Blind.to_u8()),
+            Instruction::Reblind => program.push(Opcode::Reblind.to_u8()),
+            Instruction::Unblind => program.push(Opcode::Unblind.to_u8()),
+            Instruction::Issue => program.push(Opcode::Issue.to_u8()),
+            Instruction::Borrow => program.push(Opcode::Borrow.to_u8()),
+            Instruction::Retire => program.push(Opcode::Retire.to_u8()),
+            Instruction::Qty => program.push(Opcode::Qty.to_u8()),
+            Instruction::Flavor => program.push(Opcode::Flavor.to_u8()),
+            Instruction::Cloak(m, n) => {
+                program.push(Opcode::Cloak.to_u8());
+                encoding::write_u32(*m as u32, program);
+                encoding::write_u32(*n as u32, program);
+            }
+            Instruction::Import => program.push(Opcode::Import.to_u8()),
+            Instruction::Export => program.push(Opcode::Export.to_u8()),
+            Instruction::Input => program.push(Opcode::Input.to_u8()),
+            Instruction::Output(k) => {
+                program.push(Opcode::Output.to_u8());
+                encoding::write_u32(*k as u32, program);
+            }
+            Instruction::Contract(k) => {
+                program.push(Opcode::Contract.to_u8());
+                encoding::write_u32(*k as u32, program);
+            }
+            Instruction::Nonce => program.push(Opcode::Nonce.to_u8()),
+            Instruction::Log => program.push(Opcode::Log.to_u8()),
+            Instruction::Signtx => program.push(Opcode::Signtx.to_u8()),
+            Instruction::Call => program.push(Opcode::Call.to_u8()),
+            Instruction::Left => program.push(Opcode::Left.to_u8()),
+            Instruction::Right => program.push(Opcode::Right.to_u8()),
+            Instruction::Delegate => program.push(Opcode::Delegate.to_u8()),
+            Instruction::Ext(x) => program.push(*x),
+        };
     }
 
     pub fn encode_program<I>(iterator: I, program: &mut Vec<u8>)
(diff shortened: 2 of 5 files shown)


zkvm/src/encoding.rs
```diff
@@ -124,6 +124,13 @@ pub fn write_u32<'a>(x: u32, target: &mut Vec<u8>) {
     target.extend_from_slice(&buf);
 }
 
+// Writes a LE64-encoded integer
+pub fn write_u64<'a>(x: u64, target: &mut Vec<u8>) {
```
oleganza: do we use this method?

zkvm/src/ops.rs
```diff
@@ -194,7 +195,69 @@ impl Instruction {
     /// Appends the bytecode representation of an Instruction
     /// to the program.
     pub fn encode(&self, program: &mut Vec<u8>) {
-        unimplemented!()
+        match self {
+            Instruction::Push(data) => {
+                program.push(Opcode::Push.to_u8());
+                data.encode(program);
+            }
+            Instruction::Drop => program.push(Opcode::Drop.to_u8()),
```
oleganza: Suggestion: define a closure at the beginning of this method that captures program and does .to_u8() inside:

```rust
let write = |op: Opcode| { program.push(op.to_u8()) };
```

and then use it as:

```suggestion
            Instruction::Drop => write(Opcode::Drop),
```

zkvm/src/types.rs
```diff
@@ -213,7 +222,34 @@ impl Predicate {
 
 impl PredicateWitness {
     pub fn to_point(&self) -> CompressedRistretto {
-        unimplemented!()
+        self.to_commitment().compress()
+    }
+
+    fn to_commitment(&self) -> RistrettoPoint {
```
oleganza: to avoid confusion with pedersen commitments, i'd call this method `to_uncompressed_point()`.

oleganza: Also, it seems like `predicate.rs` is better place for this definition, where we could also factor out all the hashing in a single private function.

zkvm/src/types.rs
```diff
@@ -281,15 +317,25 @@ impl Data {
     pub fn len(&self) -> usize {
         match self {
             Data::Opaque(data) => data.len(),
+            // TBD: calculate length of witness data
+            // Is there a way to do this without just serializing the
+            // data and taking its length?
```
oleganza: We can have `serialized_len()` method on all types that can be captured by `Data` and call them recursively to add them up. Not sure how tedious it would be.

Alternative is to rename this method to `minimum_serialized_length()` which guarantees to give a lower bound, which is what we need for the Vec allocation. And then return 0 for Data::Witness, which means the prover would have to re-allocate the Vec as it builds it, but it's good for now and probably will never matter.

oleganza: Although we need to check whether the .len() is used where exact length is expected. Then we have to scratch our heads more.

zkvm/src/types.rs
```diff
@@ -281,15 +317,25 @@ impl Data {
...
+    pub fn to_bytes(&self) -> Vec<u8> {
         match self {
-            Data::Opaque(data) => data,
-            Data::Witness(_) => unimplemented!(),
+            Data::Opaque(data) => data.clone(),
+            Data::Witness(w) => {
+                let mut bytes: Vec<u8> = Vec::new();
+                // TBD: right now the encode/to_bytes flow
+                // seems a little backwards. Might be better
+                // to do it the other way around.
```
oleganza: yeah, lets have `encode(buf: &mut Vec<u8>)` as the primary interface for writing data, and in `to_bytes(self)` (still consuming self) use it to encode opaque data efficiently by moving Vec and for witness, simply using encode() the way it happens in this patch.

zkvm/src/types.rs
```diff
@@ -329,6 +375,36 @@ impl Data {
             },
         }
     }
+
+    /// Encodes blinded Data values for txprogram bytecode.
+    pub fn encode(&self, program: &mut Vec<u8>) {
+        match self {
+            Data::Opaque(x) => {
+                program.append(&mut x.clone());
```
oleganza: We don't need to clone the opaque data vector here:
```suggestion
                program.extend_from_slice(&x);
```

zkvm/src/types.rs
```diff
@@ -329,6 +375,36 @@ impl Data {
...
+                return;
+            }
+            Data::Witness(w) => w.encode(program),
+        };
+    }
+}
+
+impl DataWitness {
+    fn encode(&self, program: &mut Vec<u8>) {
```
oleganza: I'd rename all `encode` methods from using `program` to `buf` to be a bit more agnostic of where the bytes go.

zkvm/src/types.rs
```diff
@@ -329,6 +375,36 @@ impl Data {
...
+        match self {
+            DataWitness::Program(instr) => Instruction::encode_program(instr.iter(), program),
+            DataWitness::Predicate(pw) => pw.encode(program),
+            DataWitness::Commitment(cw) => cw.encode(program),
+            DataWitness::Scalar(s) => program.extend_from_slice(&s.to_bytes()),
+            DataWitness::Input(b) => {
+                // Input = PreviousTxID || PreviousOutput
+                let (contract, _) = (&b.0, b.1);
+                // TBD: get prev_txid
```
oleganza: for the incorrect code it's better to slap `unimplemented!()` where the value is incorrect, so it's not sounding like "this is good-to-have optimization, can do it ~later~ never"

zkvm/src/types.rs
```diff
@@ -344,6 +420,31 @@ impl Contract {
...
+    fn encode(&self, program: &mut Vec<u8>) {
+        program.extend_from_slice(&self.predicate.to_point().to_bytes());
+        for p in self.payload.iter() {
+            match p {
+                // Data = 0x00 || LE32(len) || <bytes>
+                FrozenItem::Data(d) => {
+                    program.push(0u8);
+                    let bytes = d.to_bytes();
+                    // If we can't get the length before serializing, might have
+                    // to write a filler value and then go backwards
```
oleganza: this extra allocation on the prover side probably isn't going to matter anyway, so this is good enough, imho

zkvm/src/types.rs
```diff
@@ -344,6 +420,31 @@ impl Contract {
...
+                    encoding::write_u32(bytes.len() as u32, program);
+                    program.extend_from_slice(&bytes);
+                }
+                // Value = 0x01 || <32 bytes> || <32 bytes>
+                FrozenItem::Value(v) => {
+                    program.push(1u8);
+                    program.extend_from_slice(&v.qty.to_point().to_bytes());
+                    program.extend_from_slice(&v.qty.to_point().to_bytes());
```
```suggestion
                    program.extend_from_slice(&v.flv.to_point().to_bytes());
```

oleganza: Another reason to have encode and decode methods together somewhere.

zkvm/src/types.rs
oleganza: Seems like it's conceptually cleaner to keep `Input` have the same structure as what we need to encode:

```
pub enum Input {
    Opaque(Vec<u8>),
    Witness(Box<(FrozenContract, TxID)>),
}
```

Or even a triplet:  contract, utxo, txid to avoid recomputing utxo by serializing a contract.

zkvm/src/types.rs
```diff
@@ -101,7 +102,7 @@ pub enum DataWitness {
     Predicate(Box<PredicateWitness>), // maybe having Predicate and one more indirection would be cleaner - lets see how it plays out
     Commitment(Box<CommitmentWitness>),
     Scalar(Box<Scalar>),
-    Input(Box<(FrozenContract, UTXO)>),
+    Input(Box<(FrozenContract, UTXO, TxID)>),
```
oleganza: Now we have a slight problem: as all variants in an `enum` are necessarily public (unlike struct fields that can be selectively private), someone can construct an inconsistent set of this triplet. To prevent this, it makes sense to have this triplet made into a struct, with private fields `contract, utxo, txid` and a method `.destructure(self) -> (FC, UTXO, TxID)` that consumes self and returns these fields. And all the initializer methods that we might need will ensure the consistency of this thing.

oleganza: (We can get back to what these initializers are in a follow-up PR where we'll attempt to compose a tx)

zkvm/src/types.rs
```diff
@@ -329,21 +337,79 @@ impl Data {
...
+                buf.extend_from_slice(x);
+                return;
+            }
+            Data::Witness(w) => w.encode(buf),
+        };
+    }
+
+    /// Encodes blinded Data values with length.
+    /// LE32(len) || <bytes>
+    fn encode_with_len(&self, buf: &mut Vec<u8>) {
```
oleganza: I'd avoid factoring this out to keep code clearer - this length-prefix stuff is very specific to the frozencontract encoding, so unless we need to do this in 3 more places, i'd avoid having a separate method for this. WDYT?

zkvm/src/types.rs
```diff
@@ -277,19 +280,24 @@ impl Item {
...
         }
     }
 
     /// Converts the Data into a vector of bytes
     pub fn to_bytes(self) -> Vec<u8> {
         match self {
             Data::Opaque(data) => data,
-            Data::Witness(_) => unimplemented!(),
+            Data::Witness(w) => {
+                let mut bytes: Vec<u8> = Vec::new();
```
oleganza: maybe, so later we only have to improve this in one place?

```suggestion
                let mut bytes: Vec<u8> = Vec::with_capacity(self.min_serialized_length());
```

zkvm/src/predicate.rs
```diff
@@ -71,6 +73,39 @@ impl Predicate {
...
+impl PredicateWitness {
+    pub fn to_point(&self) -> CompressedRistretto {
+        self.to_uncompressed_point().compress()
+    }
+
+    fn to_uncompressed_point(&self) -> RistrettoPoint {
+        let gens = PedersenGens::default();
+        match self {
+            PredicateWitness::Key(s) => s * gens.B,
+            PredicateWitness::Or(b) => {
```
oleganza: why is it `b`? :-) Also: you can match immediately on a nested tuple:

```suggestion
            PredicateWitness::Or((l, r)) => {
```

...

zkvm/src/predicate.rs
```diff
@@ -71,6 +73,39 @@ impl Predicate {
...
+                let mut t = Transcript::new(b"ZkVM.predicate");
+                let (left, right) = (&b.0.to_uncompressed_point(), &b.1.to_uncompressed_point());
```
oleganza: ... and then do:
```suggestion
                let (left, right) = (&l.to_uncompressed_point(), &r.to_uncompressed_point());
```

zkvm/src/types.rs
vickiniu: sounds good to me! I went with this route since we want `to_bytes` to consume `self` but the encode functions all take `&self` — it would be an easy switch to go back to having `to_bytes` take a reference, though!

zkvm/src/predicate.rs
oleganza: That works if you change `Box<(l,r)>` into `(Box<l>, Box<r>)`

oleganza: which i think fits better with all the code - boxing specific thing, not a group of things.

zkvm/src/predicate.rs
oleganza: (now i see why it's `b` - it's "boxed" pair - see the comment below)

zkvm/src/vm.rs
```diff
@@ -645,7 +644,7 @@ where
     }
 
     fn encode_output(&mut self, contract: Contract) -> Vec<u8> {
-        let mut output = Vec::with_capacity(contract.exact_output_size());
+        let mut output = Vec::with_capacity(contract.min_output_size());
```
oleganza: maybe use `min_serialized_length` for contracts too, for consistency?

zkvm/src/types.rs
```diff
@@ -165,13 +173,28 @@ impl Commitment {
...
     }
+
+    fn encode(&self, buf: &mut Vec<u8>) {
+        buf.extend_from_slice(&self.to_point().to_bytes());
+    }
+}
+
+impl InputWitness {
+    // TBD: need initializer to construct InputWitness
+    fn destructure(&self) -> (&FrozenContract, UTXO, TxID) {
```
oleganza: maybe we'll need a consuming self version of this later, but now we can replace this method with `encode`:

```rust
pub fn encode(&self, buf: &mut Vec<u8>) {
    buf.extend_from_slice(&self.txid.0);
    self.contract.encode(buf);
}
```

PR #74, opened 2019-02-12
ZkVM: refactor predicate logic [WIP]

oleganza:
# DO NOT USE THIS PATCH

### DIFF
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -194,12 +194,12 @@ impl Instruction {
 
     /// Appends the bytecode representation of an Instruction
     /// to the program.
-    pub fn encode(&self, program: &mut Vec<u8>) {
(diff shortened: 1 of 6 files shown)


oleganza:
@vickiniu This patch poisons a bunch of APIs with failability only due to Prover that could encounter invalid compressed point down the hierarchy of data structures. I don't think this is a good way forward. If the prover cannot construct a good witness, it's better to detect this at the construction time, not during VM execution time.

Let's try another approach (will open another PR and abandon this one): 

Predicate is a _structure_, not enum, that holds compressed point which is always ready for `to_point` conversion, and an additional `Option<PredicateWitness>`. This way encoding of the predicate tree never fails and does not poison the APIs. Maybe we'll need to adopt the same `Option<Witness>` style for other structs as a result.

oleganza:
Solved in a nicer way by @vickiniu in #75

PR #75, opened 2019-02-12
zkvm/predicate: move point into Predicate struct

vickiniu:
First pass at the suggestion in https://github.com/interstellar/slingshot/pull/74, to move the `CompressedRistretto` representation of a predicate into the struct

### DIFF
--- a/zkvm/src/point_ops.rs
+++ b/zkvm/src/point_ops.rs
@@ -44,7 +44,7 @@ impl PointOp {
             .ok_or(VMError::PointOperationFailed)
     }
 
-    /// Non-batched verificatin of an individual point operation
+    /// Non-batched verification of an individual point operation
     pub fn verify(self) -> Result<(), VMError> {
         if !self.compute()?.is_identity() {
             return Err(VMError::PointOperationFailed);
--- a/zkvm/src/point_ops.rs
+++ b/zkvm/src/point_ops.rs
@@ -41,7 +41,7 @@ impl PointOp {
         }
 
         RistrettoPoint::optional_multiscalar_mul(weights, points)
-            .ok_or_else(|| VMError::PointOperationFailed)
+            .ok_or(VMError::PointOperationFailed)
     }
 
     /// Non-batched verificatin of an individual point operation
@@ -93,7 +93,7 @@ impl PointOp {
         }
 
         let check = RistrettoPoint::optional_multiscalar_mul(weights, points)
-            .ok_or_else(|| VMError::PointOperationFailed)?;
+            .ok_or(VMError::PointOperationFailed)?;
         if !check.is_identity() {
             return Err(VMError::PointOperationFailed);
         }
--- a/zkvm/src/point_ops.rs
+++ b/zkvm/src/point_ops.rs
@@ -1,7 +1,7 @@
 use bulletproofs::PedersenGens;
 use curve25519_dalek::ristretto::{CompressedRistretto, RistrettoPoint};
 use curve25519_dalek::scalar::Scalar;
-use curve25519_dalek::traits::{IsIdentity, VartimeMultiscalarMul};
+use curve25519_dalek::traits::{Identity, IsIdentity, VartimeMultiscalarMul};
 
 use super::errors::VMError;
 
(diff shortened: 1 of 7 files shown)


zkvm/src/types.rs
```diff
@@ -320,10 +327,10 @@ impl Data {
         match self {
             Data::Opaque(data) => {
                 let point = Subslice::new(&data).read_point()?;
-                Ok(Predicate::Opaque(point))
+                Ok(Predicate::new(point))
             }
             Data::Witness(witness) => match witness {
-                DataWitness::Predicate(w) => Ok(Predicate::Witness(w)),
+                DataWitness::Predicate(w) => Ok(Predicate::new_witness(w.to_point(), *w)),
```
oleganza: This witness-to-point conversion should be failable, and its computation from witness is better inside Predicate::from_witness method (failable) that ensures that the point is in sync with the witness. I’d also keep Predicate.to_point method (non failable) to make witness and point private fields. We should probably use pub(crate) very rarely only when we can make it pub, but not sure if we should, while keeping it public within crate only.

zkvm/src/point_ops.rs
```diff
@@ -37,16 +37,18 @@ impl PointOp {
...
         }
 
-        let check = RistrettoPoint::optional_multiscalar_mul(weights, points)
-            .ok_or_else(|| VMError::PointOperationFailed)?;
+        RistrettoPoint::optional_multiscalar_mul(weights, points)
+            .ok_or_else(|| VMError::PointOperationFailed)
+    }
 
-        if !check.is_identity() {
+    /// Non-batched verificatin of an individual point operation
```
oleganza:
```suggestion
    /// Non-batched verification of an individual point operation
```

zkvm/src/predicate.rs
```diff
@@ -54,12 +46,33 @@ impl Predicate {
         let gens = PedersenGens::default();
         t.commit_bytes(b"prog", &prog);
         let h = t.challenge_scalar(b"h");
-        Predicate::Opaque((h * gens.B_blinding).compress())
+        Predicate::new((h * gens.B_blinding).compress())
```
oleganza: Note for later: this method as-is does not make much sense: it takes in a bytecode `[u8]` instead of `[Instruction]`, and creates an opaque witness-less predicate. We'll probably need this to take `Vec<Instruction>` and retain the witness. But we don't need to worry about this in this PR.

zkvm/src/types.rs
```diff
@@ -226,10 +226,28 @@ impl Into<Scalar> for ScalarWitness {
 }
 
 impl Predicate {
-    pub fn to_point(&self) -> CompressedRistretto {
-        match self {
-            Predicate::Opaque(point) => *point,
-            Predicate::Witness(witness) => witness.to_point(),
+    pub fn new(point: CompressedRistretto) -> Self {
```
oleganza: maybe call it `::opaque` to avoid confusion with it being a "default" initializer. It's strictly less powerful than `from_witness` and the code that uses it belongs to a verifier-only path.

```suggestion
    pub fn opaque(point: CompressedRistretto) -> Self {
```

zkvm/src/types.rs
```diff
@@ -226,10 +226,28 @@ impl Into<Scalar> for ScalarWitness {
...
+            point: witness.to_point()?,
+            witness: Some(witness),
+        })
+    }
+
+    pub fn point(&self) -> CompressedRistretto {
+        self.point
+    }
+
+    pub fn witness(&self) -> Option<PredicateWitness> {
```
oleganza: I'd return `&Option<PredicateWitness>` to avoid unnecessary cloning. Or, if that's more ergonomic, a `Option<&PredicateWitness>`. Then, the user can clone if they really need to, but if they encode the witness, they don't need to consume it.

zkvm/src/predicate.rs
vickiniu: would we use this method to construct the Predicate object on the verifier's side?

vickiniu: ah, I guess in that instance we would just read the `CompressedRistretto` point and use that, nvm !

zkvm/src/types.rs
vickiniu: got it, that makes sense!

zkvm/src/predicate.rs
```diff
@@ -68,62 +117,51 @@ impl Predicate {
...
         }
     }
 }
 
 impl PredicateWitness {
-    pub fn to_point(&self) -> CompressedRistretto {
-        self.to_uncompressed_point().compress()
-    }
-
-    fn to_uncompressed_point(&self) -> RistrettoPoint {
```
oleganza: we don't use it yet and VerificationKey API does not provide uncompressed point. So lets refactor this when we actually need uncompressed points.

PR #76, opened 2019-02-12
slidechain/mockhorizon: use mock horizon in unit tests

vickiniu:
Adds the `mockhorizon` package, defining a mock Horizon interface to use in place of making real calls to Horizon for unit tests (currently, `TestPegOut`). The integration test, `TestEndToEnd`, still calls out to a real Horizon server.

### DIFF
--- a/slidechain/mockhorizon/mockhorizon.go
+++ b/slidechain/mockhorizon/mockhorizon.go
@@ -19,17 +19,17 @@ func New() *Client {
 	}
 }
 
+// Client is a mock Horizon client, implemengint the Horizon
+// client interface.
+//
+// Our mock implementation assumes that the calling functions
+// want to submit transactions and then stream to see if they
+// have been successfully included in the ledger.
 type Client struct {
 	txs       []string
 	mu        *sync.Mutex
 	submitted *sync.Cond
 }
-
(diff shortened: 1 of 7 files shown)


slidechain/mockhorizon/mockhorizon.go
```diff
@@ -0,0 +1,146 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ch:
+		}
+
+		// figure out how we exactly want to "stream" txs
+		// stream all the ones in the array
+		// and then wait for new ones?
```
debnil: This approach does make sense to me. My only concern is that this could end up storing more things in memory than we'd like, and something like a queue could avoid that. I don't think that matters in the context of unit tests.

vickiniu: oh man sorry for leaving this comment to myself in there! I was thinking about whether or not we wanted to remove txs from the queue once we mock-stream them, but thought it might end up causing more problems if we have two stream threads concurrently

vickiniu: thanks for catching this! let me clean up the comment :)

PR #79, opened 2019-02-13
ZkVM: decode input string into Input object instead of directly into Contract

### LINKED ISSUE #72: ZkVM: decode input string into Input object
Currently we decode input string into a Contract object directly, and have to do it inside the `VM` because the payload Values need variables to be allocated.

However, we can split this work in two parts: 
1. decode input string into Input object (containing FrozenContract, FrozenValues) outside VM.
2. convert FrozenContract into Contract and FrozenValue into Value inside VM.

This would allow us to group encoding/decoding of contracts (cf. #69) together in one place, while reusing the code that converts FrozenContract to Contract for both the prover and verifier.

oleganza:
This allows decoupling parsing of the input from constructing a Contract instance and allocating necessary variables for the stored Values. The single method to convert FrozenContract into Contract becomes shared between Prover and Verifier.

* Contract-related code is moved to `contract.rs`
* `Input` and `InputWitness` are merged in one type `Input` which is constructed both by the Prover and Verifier when parsing the input string.
* Serialization code for inputs and outputs is grouped together in `contract.rs`

Closes #72.

### DIFF
--- a/zkvm/src/contract.rs
+++ b/zkvm/src/contract.rs
@@ -115,7 +115,7 @@ impl PortableItem {
             PortableItem::Value(v) => {
                 let flv = commitments[v.flv.index].closed_commitment();
                 let qty = commitments[v.qty.index].closed_commitment();
-                FrozenItem::Value(FrozenValue{ flv, qty })
+                FrozenItem::Value(FrozenValue { flv, qty })
             }
         }
     }
--- a/zkvm/src/contract.rs
+++ b/zkvm/src/contract.rs
@@ -70,7 +70,13 @@ impl Contract {
 
 impl Input {
     pub fn from_bytes(data: Vec<u8>) -> Result<Self, VMError> {
-        Self::decode(Subslice::new(&data))
+        let mut reader = Subslice::new(&data);
+        let contract = Self::decode(&mut reader)?;
+        // Check that reader bytes have been fully consumed
+        if reader.len() != 0 {
+            return Err(VMError::FormatError);
+        }
+        Ok(contract)
     }
 
     pub fn encode(&self, buf: &mut Vec<u8>) {
@@ -78,7 +84,7 @@ impl Input {
         self.contract.encode(buf);
     }
 
-    fn decode<'a>(mut reader: Subslice<'a>) -> Result<Self, VMError> {
+    fn decode<'a>(reader: &mut Subslice<'a>) -> Result<Self, VMError> {
         // Input  =  PreviousTxID || PreviousOutput
         // PreviousTxID  =  <32 bytes>
         let txid = TxID(reader.read_u8x32()?);
@@ -118,7 +124,7 @@ impl FrozenContract {
         }
     }
 
-    fn decode<'a>(mut output: Subslice<'a>) -> Result<Self, VMError> {
+    fn decode<'a>(output: &mut Subslice<'a>) -> Result<Self, VMError> {
         //    Output  =  Predicate  ||  LE32(k)  ||  Item[0]  || ... ||  Item[k-1]
         // Predicate  =  <32 bytes>
         //      Item  =  enum { Data, Value }
--- a/zkvm/src/contract.rs
+++ b/zkvm/src/contract.rs
@@ -1,4 +1,5 @@
 use crate::encoding;
+use crate::errors::VMError;
 use crate::predicate::Predicate;
 use crate::txlog::{TxID, UTXO};
 use crate::types::{Commitment, Data, Value};
@@ -18,13 +19,7 @@ pub enum PortableItem {
 }
 
 #[derive(Clone, Debug)]
-pub enum Input {
-    Opaque(Vec<u8>),
-    Witness(Box<InputWitness>),
-}
-
-#[derive(Clone, Debug)]
-pub struct InputWitness {
+pub struct Input {
     contract: FrozenContract,
     utxo: UTXO,
     txid: TxID,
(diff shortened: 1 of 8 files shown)


zkvm/src/contract.rs
```diff
@@ -0,0 +1,185 @@
...
+            match item {
+                PortableItem::Data(d) => size += 1 + 4 + d.min_serialized_length(),
+                PortableItem::Value(_) => size += 1 + 64,
+            }
+        }
+        size
+    }
+
+    /// Half-way to encoding the contract
+    pub fn to_frozen(self, commitments: &Vec<VariableCommitment>) -> FrozenContract {
```
oleganza: Maybe it will be clearer if the whole freezing/unfreezing logic is kept in two thick methods in VM (for freezing and unfreezing), without smaller methods that receive pieces of vm state from outside.

zkvm/src/contract.rs
```diff
@@ -0,0 +1,192 @@
...
+    pub fn encode(&self, buf: &mut Vec<u8>) {
+        buf.extend_from_slice(&self.txid.0);
+        self.contract.encode(buf);
+    }
+
+    fn decode<'a>(mut input: Subslice<'a>) -> Result<Self, VMError> {
+        // Input  =  PreviousTxID || PreviousOutput
+        // PreviousTxID  =  <32 bytes>
+        let txid = TxID(input.read_u8x32()?);
+        let output_slice = &input;
```
oleganza: I remember suggesting this, but now i see how this line is confusing. For safety/clarity we should probably remove Deref impl from Subslice to make this logic more explicit. And maybe move the line computing utxo up here before the slice is mutated by contract decoder.

oleganza: By removing Deref i mean adding an explicit `to_slice` method.

zkvm/src/vm.rs
```diff
@@ -361,11 +355,31 @@ where
...
     }
 
+    fn freeze_contract(&mut self, contract: Contract) -> FrozenContract {
+        let frozen_items = contract
+            .payload
+            .into_iter()
+            .map(|i| match i {
+                PortableItem::Data(d) => FrozenItem::Data(d),
+                PortableItem::Value(v) => {
+                    let flv = self.variable_commitments[v.flv.index].closed_commitment();
```
oleganza: do we really need to close the commitment? can we just take `.commitment` as is?

zkvm/src/contract.rs
```diff
@@ -0,0 +1,158 @@
...
+                PortableItem::Value(_) => size += 1 + 64,
+            }
+        }
+        size
+    }
+}
+
+impl Input {
+    pub fn from_bytes(data: Vec<u8>) -> Result<Self, VMError> {
+        Self::decode(Subslice::new(&data))
```
oleganza: let's create a subslice, pass it in by reference (&mut) and after the decode, check if its len() == 0 ("all bytes consumed"). If not, return Err(VMError::FormatError). We'll factor this into SliceReader API later.

zkvm/src/vm.rs
```diff
@@ -682,3 +634,9 @@ where
         .map_err(|_| VMError::R1CSInconsistency)
     }
 }
+
+impl VariableCommitment {
+    pub fn closed_commitment(&self) -> Commitment {
```
oleganza: probably don't need this method.

zkvm/src/vm.rs
```diff
@@ -586,90 +600,28 @@ where
...
+        let payload = contract
+            .payload
+            .into_iter()
+            .map(|p| match p {
+                FrozenItem::Data(d) => PortableItem::Data(d),
+                FrozenItem::Value(v) => {
+                    let qty = self.make_variable(v.qty);
+                    let flv = self.make_variable(v.flv);
+                    let val = Value { qty, flv };
+                    PortableItem::Value(val)
```
oleganza: alternative this can be written with less intermediate variables for clarity:

```rust
PortableItem::Value(Value {
    qty: self.make_variable(v.qty),
    flv: self.make_variable(flv),
})
```

zkvm/src/errors.rs
```diff
@@ -111,4 +111,7 @@ pub enum VMError {
 
     #[fail(display = "Predicate item must be opaque")]
     PredicateNotOpaque,
+
+    #[fail(display = "Variable commitment out of range.")]
+    CommitmentOutOfRange,
```
oleganza: no need for this error anymore?

zkvm/src/vm.rs
```diff
@@ -586,90 +599,26 @@ where
...
-                },
-                PortableItem::Value(v) => {
-                    encoding::write_u8(VALUE_TYPE, &mut output);
-                    let qty = self.get_variable_commitment(v.qty);
-                    let flv = self.get_variable_commitment(v.flv);
-                    encoding::write_point(&qty, &mut output);
-                    encoding::write_point(&flv, &mut output);
-                }
-            }
+        let contract = self.unfreeze_contract(input.contract);
```
oleganza: Maybe to make it symmetrical with `output`, we can inline this logic inside `input`? So there's only one helper method - for freezing and unfreezing.

```rust
    fn input(&mut self) -> Result<(), VMError> {
        let input = self.pop_item()?.to_data()?.to_input()?;
        let contract = self.unfreeze_contract(input.contract);
        self.push_item(contract);
        self.txlog.push(Entry::Input(input.utxo));
        self.unique = true;
        Ok(())
    }
```

zkvm/src/contract.rs
```diff
@@ -0,0 +1,164 @@
...
+#[derive(Debug)]
+pub enum PortableItem {
+    Data(Data),
+    Value(Value),
+}
+
+#[derive(Clone, Debug)]
+pub struct Input {
+    pub contract: FrozenContract,
+    pub utxo: UTXO,
```
oleganza: lets make all of these at least `pub(crate)` because we don't want to allow constructing an Input manually and making inconsistent utxo ID w.r.t. txid and contract.

Issue #80, opened 2019-02-14
zkvm: safer parsing API with subslices


oleganza:
### Problem

We currently have a few places where we are supposed to parse a vector of bytes, but do not check whether we consumed _all the bytes_:

1. decoding a point:
https://github.com/interstellar/slingshot/blob/2ed9fc02d722c205e619c96947c6df0e57b0d953/zkvm/src/types.rs#L300
2. decoding another point: 
https://github.com/interstellar/slingshot/blob/2ed9fc02d722c205e619c96947c6df0e57b0d953/zkvm/src/types.rs#L313
3. parsing an input:
https://github.com/interstellar/slingshot/blob/ad114c0a4ddc8318b91ee167411ee6431772ccd5/zkvm/src/contract.rs#L73-L78 (checking, but it's not enforced by the API)

We also allow arbitrary conversion of a subslice state into a `&[u8]` which, when the line is reordered, completely changes the content (because the subslice is mutable):
https://github.com/interstellar/slingshot/blob/ad114c0a4ddc8318b91ee167411ee6431772ccd5/zkvm/src/contract.rs#L91-L93

### Suggestion

1. Rename `Subslice` into `SliceReader`
2. Remove `Deref` and `to_vec` to avoid accidental conversion to a slice of whatever-state the reader is in.
3. Add `.slice(|r| { ... })` method that remembers the offset before/after the closure and returns both the result of the closure and the `&[u8]` that was parsed in it. Forwards the failure. We'll use it to compute UTXO after parsing the contract, but reusing the bytes that we parsed without wasting CPU on re-serializing.
4. Add `Reader::parse(&[u8]){|r| ...->T } -> Result<T, VMError>` that checks that the whole input buffer is "consumed" (the reader ends with len()==0). This replaces `Subslice::new(&[u8])`.
5. To support use in Verifier where the program is not fully parsed in a static scope, add 
   ```
   reader.skip_trailing_bytes() -> usize // returns the offset of the trailing bytes; and advances the reader to the end
   ```
   this allows removing `range()` api and be compatible with checking logic inside `::parse` which is now the _only_ way to construct a reader. We can also make this method consume the Reader - we only need it in places where we own the Reader because we own the backing bytearray.
6. Make sure the parsing APIs take `&mut` reader instead of moving it in (to make them compatible with the checking `::parse`)
7. Remove Copy/Clone traits and see if we can live w/o them - it's kinda dangerous for the same reason as converting the reader to a slice.

[...]

PR #81, opened 2019-02-14
ZkVM: add instructions const, var, min/maxtime, left, right

decentralisedkev:
Re-opening https://github.com/interstellar/zkvm/pull/30

### DIFF
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -305,9 +305,12 @@ impl Value {
 
 impl Expression {
     pub fn constant<S: Into<Scalar>>(a: S) -> Self {
+
+        let a:Scalar = a.into();
+
         Expression {
-            terms: vec![(r1cs::Variable::One(), a.into())],
-            assignment: Some(ScalarWitness::Scalar(a.into())),
+            terms: vec![(r1cs::Variable::One(), a)],
+            assignment: Some(ScalarWitness::Scalar(a)),
         }
     }
 }
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -307,7 +307,7 @@ impl Expression {
     pub fn constant<S: Into<Scalar>>(a: S) -> Self {
         Expression {
             terms: vec![(r1cs::Variable::One(), a)],
-            assignment: None,
+            assignment: Some(ScalarWitness::Scalar(a.into())),
         }
     }
 }
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -304,7 +304,7 @@ impl Value {
 }
 
 impl Expression {
-    pub fn constant(a: Scalar) -> Self {
+    pub fn constant<S: Into<Scalar>>(a: S) -> Self {
         Expression {
             terms: vec![(r1cs::Variable::One(), a)],
             assignment: None,
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -306,7 +306,7 @@ impl Value {
 impl Expression {
     pub fn constant<S: Into<Scalar>>(a: S) -> Self {
         Expression {
-            terms: vec![(r1cs::Variable::One(), a)],
+            terms: vec![(r1cs::Variable::One(), a.into())],
             assignment: Some(ScalarWitness::Scalar(a.into())),
         }
     }
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -305,8 +305,7 @@ impl Value {
 
 impl Expression {
     pub fn constant<S: Into<Scalar>>(a: S) -> Self {
-
-        let a:Scalar = a.into();
+        let a: Scalar = a.into();
 
         Expression {
             terms: vec![(r1cs::Variable::One(), a)],
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -399,6 +399,15 @@ impl Value {
     }
 }
 
+impl Expression {
+    pub fn constant(a: Scalar) -> Self {
+        Expression {
+            terms: vec![(r1cs::Variable::One(), a)],
+            assignment: None,
+        }
+    }
+}
+
 // Upcasting all types to Item
 
 impl From<Data> for Item {
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -6,13 +6,12 @@ use curve25519_dalek::scalar::Scalar;
 use merlin::Transcript;
 use spacesuit::SignedInteger;
 
-use crate::encoding;
+use crate::contract::{Contract, Input, PortableItem};
 use crate::encoding::Subslice;
 use crate::errors::VMError;
 use crate::ops::Instruction;
 use crate::predicate::Predicate;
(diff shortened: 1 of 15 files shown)


zkvm/src/vm.rs
```diff
@@ -204,11 +204,11 @@ where
                 Instruction::Drop => self.drop()?,
                 Instruction::Dup(i) => self.dup(i)?,
                 Instruction::Roll(i) => self.roll(i)?,
-                Instruction::Const => unimplemented!(),
-                Instruction::Var => unimplemented!(),
+                Instruction::Const => self.const_instr()?,
```
oleganza: would it be better or awful to use raw identifier syntax?

```suggestion
                Instruction::Const => self.r#const()?,
```

zkvm/src/vm.rs
```diff
@@ -287,6 +287,31 @@ where
         Ok(())
     }
 
+    fn const_instr(&mut self) -> Result<(), VMError> {
```
oleganza: would it be better or awful to use raw identifier syntax?

```suggestion
    fn r#const(&mut self) -> Result<(), VMError> {
```

https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/raw-identifiers.html#raw-identifiers

zkvm/src/vm.rs
```diff
@@ -287,6 +287,31 @@ where
...
+
+    fn mintime(&mut self) -> Result<(), VMError> {
+        let a: Scalar = self.mintime.into();
+        self.push_item(Expression::from(a));
+        Ok(())
+    }
+
+    fn maxtime(&mut self) -> Result<(), VMError> {
+        let a: Scalar = self.maxtime.into();
+        self.push_item(Expression::from(a));
```
oleganza: seems like this PR does not define `Expression::from`. And should it be more explicit, like `Expression::constant(Scalar)`? Spec uses "constant expressions" as a guaranteed optimization, so it's good to have explicit API for these.

zkvm/src/vm.rs
```diff
@@ -461,6 +486,36 @@ where
...
+        let mut contract = self.pop_item()?.to_contract()?;
+        let p = contract.predicate;
+
+        self.delegate.verify_point_op(|| p.prove_or(&l, &r));
+
+        contract.predicate = r;
+
+        self.push_item(contract);
+        Ok(())
+    }
```
oleganza: Maybe refactor this like:

```rust

fn left_or_right(&mut self, assign: F) -> Result<(), VMError> 
where F: FnOnce(&mut Contract, Predicate, Predicate)->()
{
    let r = self.pop_item()?.to_data()?.to_predicate()?;
    let l = self.pop_item()?.to_data()?.to_predicate()?;

    let mut contract = self.pop_item()?.to_contract()?;
    let p = contract.predicate;

    self.delegate.verify_point_op(|| p.prove_or(&l, &r));

    assign(&mut contract, l, r);

    self.push_item(contract);
    Ok(())
}

fn left(&mut self) -> Result<(), VMError> {
    left_or_right(|contract, left, _| {
        contract.predicate = left;
    })
}

fn right(&mut self) -> Result<(), VMError> {
    left_or_right(|contract, _, right| {
        contract.predicate = right;
    })
}
```

zkvm/src/vm.rs
decentralisedkev: Yep I agree

zkvm/src/types.rs
```diff
@@ -303,6 +303,15 @@ impl Value {
     }
 }
 
+impl Expression {
+    pub fn constant(a: Scalar) -> Self {
```
oleganza: if this is generic over `Into<Scalar>`, then we can have simpler one-liners in mintime/maxtime:

```suggestion
    pub fn constant<S: Into<Scalar>>(a: S) -> Self {
```

zkvm/src/vm.rs
```diff
@@ -281,6 +281,33 @@ where
...
+    fn var(&mut self) -> Result<(), VMError> {
+        let comm = self.pop_item()?.to_data()?.to_commitment()?;
+        let v = self.make_variable(comm);
+        self.push_item(v);
+        Ok(())
+    }
+
+    fn mintime(&mut self) -> Result<(), VMError> {
+        let a: Scalar = self.mintime.into();
+        self.push_item(Expression::constant(a));
```
oleganza: If `Expression::constant` is generic over `S: Into<Scalar>`:
```suggestion
        self.push_item(Expression::constant(self.mintime));
```

zkvm/src/types.rs
```diff
@@ -303,6 +303,15 @@ impl Value {
...
+        Expression {
+            terms: vec![(r1cs::Variable::One(), a)],
+            assignment: None,
```
oleganza: The VM types are reused by both Prover and Verifier. So we need to keep in mind whether a specific type is used by both, or only by one. Here, we need to keep the assignment.

```suggestion
            assignment: Some(ScalarWitness::Scalar(a.into())),
```

We'd also want to retain `ScalarWitness::Integer` if we compose integers and wish to send them into a rangeproof later, but we can leave this refactoring for later PR. It would involve writing Add/Neg traits for ScalarWitness etc.

zkvm/src/vm.rs
```diff
@@ -281,6 +281,33 @@ where
...
+
+    fn var(&mut self) -> Result<(), VMError> {
+        let comm = self.pop_item()?.to_data()?.to_commitment()?;
+        let v = self.make_variable(comm);
+        self.push_item(v);
+        Ok(())
+    }
+
+    fn mintime(&mut self) -> Result<(), VMError> {
+        let a: Scalar = self.mintime.into();
```
oleganza: unnecessary line now :-)

zkvm/src/vm.rs
```diff
@@ -281,6 +281,33 @@ where
...
+        self.push_item(Expression::constant(self.mintime));
+        Ok(())
+    }
+
+    fn maxtime(&mut self) -> Result<(), VMError> {
+        let a: Scalar = self.maxtime.into();
+        self.push_item(Expression::constant(a));
```
```suggestion
        self.push_item(Expression::constant(self.maxtime));
```

zkvm/src/types.rs
oleganza: ...and also changing `S: Into<Scalar>` for `S: Into<ScalarWitness>`.

zkvm/src/vm.rs
```diff
@@ -281,6 +282,31 @@ where
         Ok(())
     }
 
+    fn r#const(&mut self) -> Result<(), VMError> {
+        let data = self.pop_item()?.to_data()?.to_bytes();
+        let mut slice = Subslice::new(&data);
+        let scalar = slice.read_scalar()?;
```
```suggestion
        let scalar = slice.read_scalar()?;
        if slice.len() != 0 { return Err(VMError::FormatError) }
```

We should not allow garbage in the trailing bytes.

oleganza: This will be simplified and made safer later: https://github.com/interstellar/slingshot/issues/80

oleganza:
One more nitpick, `cargo fmt` and it's great!

PR #83, opened 2019-02-14
slidechain: add peg-out smart contract

debnil:
Implements a smart contract for peg-out. The better peg-out protocol (#49) suggests that the txvm-side export transaction lock the funds in a smart contract. If the peg-out Stellar transaction succeeds, that contract retires the funds; else, it repays them to the exporter. This PR accordingly updates the export transaction; implements the aforementioned post-export contract; adds a go-routine to watch for the Stellar transaction; and updates tests and associated binaries as needed.

https://trello.com/c/boButuVu


slidechain/export.go
```diff
@@ -276,67 +342,81 @@ func BuildExportTx(ctx context.Context, asset xdr.Asset, amount, inputAmt int64,
...
-	err = tpl.Sign(ctx, func(_ context.Context, msg []byte, prv []byte, path [][]byte) ([]byte, error) {
-		return ed25519.Sign(prv, msg), nil
+	retireRefdataHex := i10rjson.HexBytes(retireRefdata)
+	b := new(txvmutil.Builder)
+	b.PushdataBytes(inputRefdataHex)                                                                                     // con stack: json
+	b.Op(op.Put)                                                                                                         // arg stack: json
+	standard.SpendMultisig(b, 1, []ed25519.PublicKey{pubkey}, inputAmt, assetID, anchor, standard.PayToMultisigSeed1[:]) // arg stack: inputval, sigcheck
+	b.Op(op.Get).Op(op.Get)                                                                                              // con stack: sigcheck, inputval
+	b.PushdataInt64(amount).Op(op.Split)                                                                                 // con stack: sigcheck, changeval, retireval
+	b.PushdataInt64(1).Op(op.Roll).Op(op.Drop)                                                                           // con stack: sigcheck, retireval
```
bobg: You can only `drop` a value if it's zero! Which is only true if `amount == inputAmt`. If it's non-zero, you need to pay it as change (e.g. using `standard.PayToMultisigProg1`) to the owner of `pubkey`.

slidechain/postexport.go
```diff
@@ -0,0 +1,169 @@
...
+			tup.PushdataBytes(refdataHex)
+		})
+	})
+	b.PushdataInt64(peggedOut).Op(op.Put) // arg stack: selector
+	b.Op(op.Input).Op(op.Call)
+	b.PushdataBytes(bcid)
+	b.PushdataInt64(int64(bc.Millis(time.Now().Add(10 * time.Minute))))
+	b.Op(op.Nonce).Op(op.Finalize)
+
+	// TODO(debnil): Should there be a signature check here?
```
bobg: Yes! 😄

slidechain/cmd/export/export.go
```diff
@@ -117,82 +132,108 @@ func main() {
...
+
+func mustDecode(src string) []byte {
+	bytes, err := hex.DecodeString(src)
+	if err != nil {
+		log.Fatalf("error decoding %s: %s", src, err)
+	}
+	return bytes
+}
+
+func submitTxAndWait(ctx context.Context, tx *bc.Tx, slidechaind string, fn checkFn, asset xdr.Asset, amount int64, temp, exporter string, seqnum int64, anchor, pubkey []byte) error {
```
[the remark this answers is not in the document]
debnil: Thanks for the voice of simplification -- `?wait=1` should be more than sufficient here.

slidechain/export.go
```diff
@@ -90,15 +134,37 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
...
+						log.Fatalf("getting error codes from failed submission of tx %s", txid)
+					}
+					if resultCodes.TransactionCode != xdr.TransactionResultCodeTxBadSeq.String() { // non-retriable error
+						break
+					}
+				} else {
+					break
+				}
+				if i == 0 {
+					_, err = c.DB.ExecContext(ctx, `UPDATE exports SET exported=1, pegged_out=2 WHERE txid=$1`, txid)
```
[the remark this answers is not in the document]
debnil: Settled over DM, but for posterity: We can replace `exported` with four values of `pegged_out`: `PegOutNotYet`, set upon inserting the row in `exports`; `PegOutOK`, when the peg-out is successful; `PegOutRetry`, when the peg-out fails but is retriable; and `PegOutFail`, when the peg-out fails and is not retriable.

slidechain/custodian.go
```diff
@@ -106,6 +107,7 @@ func newCustodian(ctx context.Context, db *sql.DB, hclient horizon.ClientInterfa
 		hclient:       hclient,
 		imports:       sync.NewCond(new(sync.Mutex)),
 		exports:       sync.NewCond(new(sync.Mutex)),
+		pegouts:       sync.NewCond(new(sync.Mutex)),
```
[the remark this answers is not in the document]
debnil: This was super helpful - I think most of this should be covered now. Within `watchPegOuts`, a go routine polls periodically using `time.NewTicker`, and there's also a loop over values from a channel as you described.

A concurrency problem did arise, and I'm curious if I approached it correctly. The new approach could `SELECT` some rows within each routine. That becomes especially problematic if they're also post-processed twice, since that would attempt to delete an already deleted row. As a result, I defined a new mutex `exportmux` on the `Custodian` to avoid double-processing exported rows. I treated the entire iteration of a for-loop within each routine as a critical section, and the new mutex locks and unlocks accordingly.

I tried to approach this using channels, but they seemed like overkill - a mutex would be sufficient. Would channels have been a better approach?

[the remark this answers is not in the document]
debnil: > the post-pegout tx can fail separately from the peg-out tx.

I thought about this as well, but I'm not certain what the right protection or behavior is for this. Currently, if the post-pegout tx fails, `doPegOutTx` will return an error, which will result in a `log.Fatal` in `watchPegOuts`. What behavior would we want instead?

slidechain/export.go
```diff
@@ -51,15 +106,12 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
 		case <-ch:
 		}
 
-		const q = `SELECT txid, amount, asset_xdr, exporter, temp, seqnum FROM exports WHERE exported=0`
+		const q = `SELECT txid, amount, asset_xdr, exporter, temp, seqnum FROM exports WHERE pegged_out=0`
```
bobg: Use your `PegOutState` constants, not integer literals.

slidechain/export.go
```diff
@@ -90,15 +142,36 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
...
 			log.Printf("pegging out export %x: %d of %s to %s", txid, amounts[i], asset.String(), exporters[i])
-			// TODO(vniu): flag txs that fail with unretriable errors in the db
-			err = c.pegOut(ctx, exporter, asset, int64(amounts[i]), tempID, xdr.SequenceNumber(seqnums[i]))
-			if err != nil {
-				log.Fatalf("pegging out tx: %s", err)
+			var peggedOut PegOutState
+			for i := 0; i < 5; i++ {
+				err = c.pegOut(ctx, exporter, asset, int64(amounts[i]), tempID, xdr.SequenceNumber(seqnums[i]))
+				if err == nil { // successful peg-out
+					peggedOut = PegOutOK
```
bobg: Nit: handle errors first, and avoid indenting non-error logic. See https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow.

slidechain/export.go
```diff
@@ -248,12 +321,13 @@ func SubmitPreExportTx(hclient horizon.ClientInterface, kp *keypair.Full, custod
...
-// BuildExportTx builds a txvm retirement tx for an asset issued
-// onto slidechain. It will retire `amount` of the asset, and the
-// remaining input will be output back to the original account.
-func BuildExportTx(ctx context.Context, asset xdr.Asset, amount, inputAmt int64, temp string, anchor []byte, prv ed25519.PrivateKey, seqnum xdr.SequenceNumber) (*bc.Tx, error) {
-	if inputAmt < amount {
-		return nil, fmt.Errorf("cannot have input amount %d less than export amount %d", inputAmt, amount)
+// BuildExportTx locks money to be retired in a TxVM smart contract.
+// If the peg-out transaction succeeds, it retires `exportAmt` of this asset,
+// and `inputAmt` - `exportAmt` will be output back to the original account.
+// Else, it returns `exportAmt` to the exporter address.
```
bobg: Nit: the wording needs a tweak here to make it clearer that you're describing what the smart contract does (when invoked separately later), not what this function does.

slidechain/export.go
```diff
@@ -271,117 +345,140 @@ func BuildExportTx(ctx context.Context, asset xdr.Asset, amount, inputAmt int64,
 		return nil, err
 	}
 	pubkey := prv.Public().(ed25519.PublicKey)
+
+	// We first split off the difference between inputAmt and amt.
```
bobg: Update the comment to reflect the new name of `exportAmt`.

slidechain/export.go
```diff
@@ -271,117 +345,140 @@ func BuildExportTx(ctx context.Context, asset xdr.Asset, amount, inputAmt int64,
...
+	}
+	sigProg := standard.VerifyTxID(vm.TxID)
+	msg := append(sigProg, anchor...)
+	sig := ed25519.Sign(prv, msg)
+	b.PushdataBytes(sig).Op(op.Put)
+	b.PushdataBytes(sigProg).Op(op.Put)
+	b.Op(op.Call)
+
+	prog2 := b.Build()
+	log.Printf("%x", prog2)
```
bobg: Remove this debugging line.

slidechain/export.go
```diff
@@ -271,117 +345,140 @@ func BuildExportTx(ctx context.Context, asset xdr.Asset, amount, inputAmt int64,
...
+	tx, err := bc.NewTx(prog2, 3, math.MaxInt64)
 	if err != nil {
-		return nil, errors.Wrap(err, "building tx")
-	}
-	if inputAmt > amount {
-		txresult := txresult.New(tx)
-		output := txresult.Outputs[0].Value
-		log.Printf("output: assetid %x amount %x anchor %x", output.AssetID.Bytes(), output.Amount, output.Anchor)
+		return nil, errors.Wrap(err, "making pre-export tx")
```
bobg: This is the export tx, not the pre-export tx.

slidechain/export.go
```diff
@@ -271,117 +345,140 @@ func BuildExportTx(ctx context.Context, asset xdr.Asset, amount, inputAmt int64,
...
+// {"O", ...}
 // {"O", caller, outputid}
-// {"R", ...}
-// {"L", ...}
 // {"F", ...}
-func IsExportTx(tx *bc.Tx, asset xdr.Asset, inputAmt int64, temp, exporter string, seqnum int64) bool {
-	// The export transaction when we export the full input amount has seven operations, and when we export
-	// part of the input and output the rest back to the exporter, it has ten operations
-	if len(tx.Log) != 7 && len(tx.Log) != 10 {
+func IsExportTx(tx *bc.Tx, asset xdr.Asset, inputAmt int64, temp, exporter string, seqnum int64, anchor, pubkey []byte) bool {
```
bobg: It looks like this function isn't needed - it's only called in a test.

(If you want to keep calling it from that test, you could move its definition into the test file.)

slidechain/postexport.go
```diff
@@ -0,0 +1,167 @@
...
+	err := asset.UnmarshalBinary(assetXDR)
+	if err != nil {
+		return errors.Wrap(err, "unmarshaling asset xdr")
+	}
+	assetBytes, err := asset.MarshalBinary()
+	if err != nil {
+		return errors.Wrap(err, "marshaling asset bytes")
+	}
+	assetID := bc.NewHash(txvm.AssetID(importIssuanceSeed[:], assetBytes))
+	ref := struct {
```
bobg: This struct definition should not be repeated in multiple places, which could drift out of sync with each other. It should have its own `type`.

slidechain/postexport.go
```diff
@@ -0,0 +1,167 @@
...
+	i10rjson "github.com/chain/txvm/encoding/json"
+	"github.com/chain/txvm/errors"
+	"github.com/chain/txvm/protocol/bc"
+	"github.com/chain/txvm/protocol/txvm"
+	"github.com/chain/txvm/protocol/txvm/op"
+	"github.com/chain/txvm/protocol/txvm/txvmutil"
+	"github.com/stellar/go/xdr"
+)
+
+func (c *Custodian) doPostExport(ctx context.Context, assetXDR, anchor, txid []byte, amount, seqnum, peggedOut int64, exporter, temp string, pubkey []byte) error {
```
bobg: "Post-export" is the wrong name for this function/this file/the other things here. This is _post-pegout_ logic.

slidechain/postexport.go
```diff
@@ -0,0 +1,167 @@
...
+// IsPostExportTx returns whether or not a txvm transaction matches the slidechain post-export tx format.
+//
+// Expected log is
+// {"I", ...}
+// {"X", ...}
+// {"L", ...}
+// {"N", ...}
+// {"R", ...}
+// {"F", ...}
+func IsPostExportTx(tx *bc.Tx, asset xdr.Asset, amount int64, temp, exporter string, seqnum int64, anchor, pubkey []byte) bool {
```
bobg: This function doesn't seem to be needed.

slidechain/watch.go
```diff
@@ -117,62 +118,153 @@ func (c *Custodian) watchExports(ctx context.Context) {
...
 
-				c.exports.Broadcast()
+	// Poll the database every minute for unprocessed exports.
+	go func(ctx context.Context) {
+		defer log.Print("export table poll exiting")
+		for t := time.Tick(1 * time.Minute); ; {
+			// Lock access to the export database.
+			// We do this to avoid double-processing any peg-out tx, via selecting it in both go-routines.
+			// We unlock at the end of an iteration of this go-routine or before any early return.
+			c.exportmux.Lock()
```
bobg: When using a mutex, try to organize the code so that you can `defer` a call to `Unlock` as soon as you `Lock` it, and then return from that function as soon as possible to release the lock.

One thing you definitely want to avoid is trying to `Unlock` the mutex "manually" in every possible exit path from the function. (Plus, to be thorough, you'd need a `recover` function to handle non-local exits, so you can't really avoid `defer`.)

slidechain/watch.go
```diff
@@ -117,62 +118,153 @@ func (c *Custodian) watchExports(ctx context.Context) {
...
+	}
+}
 
-				log.Printf("recorded export: %d of txvm asset %x (Stellar %x) for %s", retiredAmount, retiredAssetIDBytes, info.AssetXDR, exporter.Address())
+// Runs as a goroutine
+func (c *Custodian) watchPegOuts(ctx context.Context) {
+	defer log.Print("watchPegOuts exiting")
 
-				c.exports.Broadcast()
+	// Poll the database every minute for unprocessed exports.
```
bobg: A better way to do this that doesn't require a separate goroutine or a mutex:

```
ticker := time.NewTicker(time.Minute)
defer ticker.Stop()

for {
  var pegout pegOut
  select {
    case <-ctx.Done(): ...
    case <-ticker.C: ...
    case pegout = <-c.pegouts: ...
  }
}
```

slidechain/watch.go
```diff
@@ -117,62 +118,153 @@ func (c *Custodian) watchExports(ctx context.Context) {
...
+			return
+		case pegOut = <-c.pegouts:
+		}
+		const q = `SELECT amount, asset_xdr, exporter, temp, seqnum, anchor, pubkey FROM exports WHERE txid=$1 AND pegged_out=$2`
+		var (
+			assetXDR, anchor, pubkey []byte
+			amount, seqnum           int64
+			exporter, temp           string
+		)
+		err := sqlutil.ForQueryRows(ctx, c.DB, q, pegOut.txid, pegOut.state, func(qAmount int64, qAssetXDR []byte, qExporter, qTemp string, qSeqnum int64, qAnchor, qPubkey []byte) {
```
bobg: Is it necessary to get this information from the database? Doesn't `pegOutFromExports` have it all, and can't it send it over the `c.pegouts` channel?

Also, it's never possible to have more than one row matching the given txid, so you want `c.DB.QueryRowContext` here, not `ForQueryRows`. (Although it is possible to have zero rows, which should perhaps be treated as an error.)

slidechain/export.go
```diff
@@ -90,15 +154,47 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
...
+							log.Fatalf("getting error codes from failed submission of tx %s", txid)
+						}
+						if resultCodes.TransactionCode != xdr.TransactionResultCodeTxBadSeq.String() { // non-retriable error
+							peggedOut = pegOutFail
+							break
+						}
+					} else {
+						log.Fatalf("could not get error code from failed submission of tx %s", txid)
+					}
+					_, err = c.DB.ExecContext(ctx, `UPDATE exports SET pegged_out=$1 WHERE txid=$2`, pegOutRetry, txid)
```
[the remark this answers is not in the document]
bobg: Also, shouldn't `peggedOut` be set to `pegOutRetry` in case the loop exits now?

slidechain/export.go
```diff
@@ -90,15 +154,47 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
...
+				peggedOut = pegOutOK
+				break
 			}
-			_, err = c.DB.ExecContext(ctx, `UPDATE exports SET exported=1 WHERE txid=$1`, txid)
+			_, err = c.DB.ExecContext(ctx, `UPDATE exports SET pegged_out=$1 WHERE txid=$2`, peggedOut, txid)
 			if err != nil {
 				log.Fatalf("updating export table: %s", err)
 			}
+			// Send peg-out info to goroutine for post-peg-out txvm txs
+			pegouts <- pegOut{
```
bobg: Shouldn't this happen only if `peggedOut` is `pegOutOK`?

slidechain/export.go
```diff
@@ -90,15 +154,47 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
 			}
 
 			log.Printf("pegging out export %x: %d of %s to %s", txid, amounts[i], asset.String(), exporters[i])
-			// TODO(vniu): flag txs that fail with unretriable errors in the db
-			err = c.pegOut(ctx, exporter, asset, int64(amounts[i]), tempID, xdr.SequenceNumber(seqnums[i]))
-			if err != nil {
-				log.Fatalf("pegging out tx: %s", err)
+			var peggedOut pegOutState
+			for i := 0; i < 5; i++ {
```
bobg: Danger: this `i` shadows the `i` in the outer loop.

With the `break`s and `continue`s in this loop, and the other logic errors I've pointed out, I think you're better off factoring the retry logic out as a separate function.

Better still, don't retry in a loop here at all. Instead, if you encounter a retriable error, flag it as such in the database and then move on to the next `exports` record. Come back around to the retriable record later via the db query, which should produce both `pegOutNotYet` and `pegOutRetry` rows.

slidechain/export.go
```diff
@@ -90,15 +154,47 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
...
+					if err != nil {
+						log.Fatalf("updating export table for retriable tx: %s", err)
+					}
+					continue
+				}
+				peggedOut = pegOutOK
+				break
 			}
-			_, err = c.DB.ExecContext(ctx, `UPDATE exports SET exported=1 WHERE txid=$1`, txid)
+			_, err = c.DB.ExecContext(ctx, `UPDATE exports SET pegged_out=$1 WHERE txid=$2`, peggedOut, txid)
```
bobg: See comment above about checking the number of affected rows.

slidechain/export.go
```diff
@@ -271,117 +368,64 @@ func BuildExportTx(ctx context.Context, asset xdr.Asset, amount, inputAmt int64,
...
-		exporter,
-	}
-	refdata, err := json.Marshal(ref)
-	if !bytes.Equal(refdata, tx.Log[3][2].(txvm.Bytes)) {
-		return false
-	}
-	// Beyond this, the two transactions diverge but must either finalize
-	// or output the remaining unconsumed input
-	return true
+	return tx, nil
```
bobg: This and the preceding three lines can be collapsed to `return tx, errors.Wrap(err, ...)`.

slidechain/export_test.go
```diff
@@ -59,7 +60,8 @@ func TestPegOut(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	_, err = c.DB.Exec("INSERT INTO exports (txid, amount, asset_xdr, temp, seqnum, exporter) VALUES ($1, $2, $3, $4, $5, $6)", "", amount, lumenXDR, temp, seqnum, kp.Address())
+	var anchor, pubkey [32]byte // anchor and pubkey do not matter to test this functionality
```
bobg: Style nit: I would just declare a single thing called `zero32` and pass it for both anchor and pubkey below.

slidechain/postpegout.go
```diff
@@ -0,0 +1,97 @@
...
+			tup.PushdataInt64(amount)
+			tup.PushdataBytes(assetID.Bytes())
+			tup.PushdataBytes(anchor)
+		})
+		contract.Tuple(func(tup *txvmutil.TupleBuilder) { // {'S', refdata}
+			tup.PushdataByte(txvm.BytesCode)
+			tup.PushdataBytes(refdata)
+		})
+	})
+	b.PushdataInt64(int64(peggedOut)).Op(op.Put)                        // con stack: snapshot; arg stack: selector
```
bobg: The selector for phase 2 of the export contract has to be zero (to refund the exported funds) or non-zero (to retire them). But AFAICT `peggedOut` is always either `pegOutOK` (1) or `pegOutFail` (3), so this will always retire.

slidechain/schema.go
```diff
@@ -27,10 +27,12 @@ CREATE TABLE IF NOT EXISTS exports (
   txid BLOB NOT NULL PRIMARY KEY,
   exporter TEXT NOT NULL,
   amount INTEGER NOT NULL,
-  asset_xdr BLOB NOT NULL,
+  asset_xdr TEXT NOT NULL,
```
bobg: Why change this?

slidechain/export.go
```diff
@@ -25,10 +26,74 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-const baseFee = 100
+type pegOut struct {
+	Txid     []byte      `json:"txid,omitempty"`
+	AssetXDR string      `json:"asset"`
```
bobg: Let's stay consistent about XDR-named variables having `[]byte` type and _not_ being base64-encoded.

slidechain/watch.go
```diff
@@ -117,61 +117,101 @@ func (c *Custodian) watchExports(ctx context.Context) {
...
-				const q = `
-					INSERT INTO exports 
-					(txid, exporter, amount, asset_xdr, temp, seqnum)
-					VALUES ($1, $2, $3, $4, $5, $6)`
-				_, err = c.DB.ExecContext(ctx, q, tx.ID.Bytes(), exporter.Address(), retiredAmount, info.AssetXDR, info.Temp, info.Seqnum)
+	ticker := time.NewTicker(time.Minute)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
```
bobg: Isn't this where you want to return from the goroutine?

slidechain/watch.go
```diff
@@ -117,61 +117,101 @@ func (c *Custodian) watchExports(ctx context.Context) {
...
-				err = exporter.SetAddress(info.Exporter)
-				if err != nil {
-					continue
-				}
+			c.exports.Broadcast()
+		}
+	}
+}
+
+// Runs as a goroutine
```
bobg: Nit: full sentences.

slidechain/export.go
bobg: (Oops, sorry: `pegOutOK` or `pegOutFail`.)

slidechain/postpegout.go
```diff
@@ -0,0 +1,111 @@
...
+	r, err := c.S.submitTx(ctx, tx)
+	if err != nil {
+		return errors.Wrap(err, "submitting post-peg-out tx")
+	}
+	err = c.S.waitOnTx(ctx, tx.ID, r)
+	if err != nil {
+		return errors.Wrap(err, "waiting on post-peg-out tx to hit txvm")
+	}
+	// Delete relevant row from exports table.
+	result, err := c.DB.ExecContext(ctx, `DELETE FROM exports WHERE txid=$1`, txid)
```
bobg: What if a crash occurs right before this line? We'll have retired or refunded the txvm funds, but not updated the db. How should slidechaind recover the next time it runs?

slidechain/watch.go
```diff
@@ -117,56 +117,97 @@ func (c *Custodian) watchExports(ctx context.Context) {
...
 
-			for i := 0; i < len(tx.Log)-2; i++ {
-				item := tx.Log[i]
-				if item[0].(txvm.Bytes)[0] != txvm.RetireCode {
-					continue
-				}
-				retiredAmount := int64(item[2].(txvm.Int))
-				retiredAssetIDBytes := []byte(item[3].(txvm.Bytes))
+			outputIndex := len(tx.Log) - 2
+			if tx.Log[outputIndex][0].(txvm.Bytes)[0] != txvm.OutputCode {
```
bobg: To be sure this is an export tx (and not a tx masquerading as one), it would be good to check that the export-contract seed appears in the log. The easiest way to do that is to have the export contract do a `log` instruction and then checking the seed in the extra `{"L", seed, item}` entry against `exportContract1Seed`.

This can be for a future PR, but should be at least a TODO in this one.

slidechain/watch.go
```diff
@@ -117,56 +117,97 @@ func (c *Custodian) watchExports(ctx context.Context) {
...
-				// Record the export in the db,
-				// then wake up a goroutine that executes peg-outs on the main chain.
-				const q = `
-					INSERT INTO exports 
-					(txid, exporter, amount, asset_xdr, temp_addr, seqnum)
-					VALUES ($1, $2, $3, $4, $5, $6)`
-				_, err = c.DB.ExecContext(ctx, q, tx.ID.Bytes(), exporter.Address(), retiredAmount, info.AssetXDR, info.TempAddr, info.Seqnum)
+			}
+		case p, ok := <-pegouts:
+			if ok {
```
bobg: Change this to `if !ok { log.Fatalf(...) }` and then don't indent the normal-flow logic. (This is the [indent error flow](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow) style guideline we've discussed before.)

slidechain/export.go
```diff
@@ -106,28 +162,43 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
...
 			peggedOut := pegOutOK
-			err = c.pegOut(ctx, exporter, asset, int64(amounts[i]), tempID, xdr.SequenceNumber(seqnums[i]))
+			err = c.pegOut(ctx, exporter, asset, amounts[i], tempID, xdr.SequenceNumber(seqnums[i]))
 			if err != nil {
 				peggedOut = pegOutFail
 				if herr, ok := errors.Root(err).(*horizon.Error); ok {
 					resultCodes, err := herr.ResultCodes()
 					if err != nil {
-						log.Fatalf("getting error codes from failed submission of tx %s", txid)
+						log.Fatalf("getting error codes from failed submission of tx %x", txid)
```
bobg: Include `herr` and/or `err` in this log line. And what about the outermost `err` (which is being shadowed)?

slidechain/postpegout.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return errors.Wrap(err, "computing transaction ID")
+	}
+	sig := ed25519.Sign(c.privkey, vm.TxID[:])
+	b.Op(op.Get).PushdataBytes(sig).Op(op.Put) // con stack: sigchecker; arg stack: sig
+	b.Op(op.Call)
+
+	// Build, submit, and wait for the post-peg-out tx to hit txvm.
+	prog2 := b.Build()
+	tx, err := bc.NewTx(prog2, 3, math.MaxInt64)
```
bobg: Create the tx with the proper runlimit.

slidechain/export.go
```diff
@@ -166,9 +166,9 @@ func (c *Custodian) pegOutFromExports(ctx context.Context, pegouts chan<- pegOut
 			if err != nil {
 				peggedOut = pegOutFail
 				if herr, ok := errors.Root(err).(*horizon.Error); ok {
-					resultCodes, err := herr.ResultCodes()
-					if err != nil {
-						log.Fatalf("getting error codes from failed submission of tx %x", txid)
+					resultCodes, rerr := herr.ResultCodes()
+					if rerr != nil {
+						log.Fatalf("getting error codes from failed submission of tx %x (with horizon err '%s' from original error '%s'): %s", txid, herr, err, rerr)
```
bobg: My apologies: on further reflection, including both `herr` and `err` here is redundant, because `herr` _is_ `err`.

PR #84, opened 2019-02-14
zkvm: convert variables to expressions explicitly

### LINKED ISSUE #82: ZkVM: inconsistent terminology around Expressions/Variables
@decentralisedkev pointed this out in a private conversation:

>Neg instruction states that:
>
>>2.If the expression is a *detached* variable, attaches it to the constraint system.
>
>Does this conflict with the part in Attached and Detached Variables, which states:
>
>>When an expression is formed using detached variables, all of them transition to an attached state?
>
>If it does then, I may need to edit the part on Expressions which states that:
>
>>Expression is a linear combination of attached variables with cleartext scalar weights.?

Sources: [`neg`](https://github.com/interstellar/slingshot/blob/main/zkvm/spec/ZkVM.md#neg), [Attached and detached variables](https://github.com/interstellar/slingshot/blob/main/zkvm/spec/ZkVM.md#attached-and-detached-variables), [Expression](https://github.com/interstellar/slingshot/blob/main/zkvm/spec/ZkVM.md#expression-type).

The short answer is: "Expression" is accidentally used both as an object representing a linear combination of low-level and high-level variables, as well as a "supertype" of a Variable (since any variable can be turned into a linear combination of itself at no cost), which allows using variables where expressions are expected w/o explicit conversion.

How do we make this more clear?

### Option 1: explicit name for the supertype

Rewrite all the parts of the spec where `Expression | Variable` is expected in terms of this supertype, and reflect this in implementation with explicit conversion to expression with attaching the vars as necessary.

### Option 2: explicit conversion instruction for Variable->Expression

Instruction `attach` (or `expr`) would turn the `Variable` into `Expression` and attach it into R1CS. The spec can be made simpler, and the user is more explicitly aware of the variable being attached. But we expose this extra instruction which seems like a bit of redundancy. But maybe things would be overall more clear and less footgunny.

PR: #84 

### Else

Other ideas?

oleganza:
This gets rid of attempts to have a second meaning for `Expression` as a "super type" of Variables and linear combinations.

* New `expr` instruction that attaches the variable and turns it into an expression.
* All operations that take expressions do not need to also accept Variable and attach it. It's done by `expr`.

Closes #82


oleganza:
PTAL @vickiniu @bobg

zkvm/spec/ZkVM.md
```diff
@@ -1248,14 +1252,23 @@ Pushes an [expression](#expression-type) `expr` corresponding to the [maximum ti
 
 The one-term expression represents time bound as a weight on the R1CS constant `1` (see [`const`](#const)).
 
+#### expr
+
+_var_ **expr** → _ex_
+
+1. Pops a [variable](#variable-type) `var`.
+2. If it is [detached](#attached-and-detached-variables), attaches it to the constraint system.
```
vickiniu: If the variable is already attached, does `expr` do nothing?

decentralisedkev:
Does this mean we need to get rid of `variable_to_expression` function in vm.rs?

zkvm/spec/ZkVM.md
```diff
@@ -204,11 +204,11 @@ bound to its [Pedersen commitment](#pedersen-commitment).
 
 A [point](#point) that represents a commitment to a secret scalar can be turned into a variable using the [`var`](#var) instruction.
 
-A cleartext [scalar](#scalar) can be turned into a single-term [expression](#expression-type) using the [`const`](#const) instruction (which does not allocate a variable). Since we do not need to hide their values, a Variable is not needed.
+A cleartext [scalar](#scalar) can be turned into a single-term [expression](#expression-type) using the [`const`](#const) instruction (which does not allocate a variable). Since we do not need to hide their values, a Variable is not needed to represent the cleartext constant.
 
 Variables can be copied and dropped at will, but cannot be ported across transactions via [outputs](#output-structure).
 
-Examples of variables: [value quantities](#value-type) and [time bounds](#time-bounds).
+[Value quantities and flavors](#value-type) are represented as variables.
```
decentralisedkev: question: In the definition of `Value type` we have "Both quantity and flavor are represented as variables." Is this fine?

zkvm/spec/ZkVM.md
```diff
@@ -1278,14 +1290,19 @@ _ex1 ex2_ **mul** → _ex3_
 
 Multiplies two [expressions](#expression-type) producing another [expression](#expression-type) representing the result of multiplication.
 
-This performs an optimization: if one of the expressions `ex1` or `ex2` contains only one term and this term is for the variable representing the R1CS constant `1` (in other words, the statement is a cleartext constant), then the other expression is multiplied by that constant in-place without allocating a multiplier in the [constraint system](#constraint-system).
+This performs a _guaranteed optimization_: if one of the expressions `ex1` or `ex2` contains
```
decentralisedkev: nit: Could we remove the explanation for constant expression?

It would then read as: "This performs a guaranteed optimization if one of the expressions, ex1 or ex2 is a cleartext constant. The other expression is multiplied by that constant in-place without allocating a multiplier in the constraint system."

WDYT?

zkvm/spec/ZkVM.md
oleganza: `expr` creates Expression object, consuming Variable object regardless. But it's not really much of a side-effect - one can drop and dup Variables at will.

oleganza:
>Does this mean we need to get rid of `variable_to_expression` function in vm.rs and replace all occurrences with `expr`?

Nope, `expr` is an instruction in the VM, and is only supposed to be invoked by the bytecode interpreter. The `variable_to_expression` is used in a few places like `issue` and `import` where we need a variable, but also we need its commitment, so we cannot just ask for an expression.

zkvm/spec/ZkVM.md
oleganza: Do you mean it's redundant to mention this here? I just wanted to show an example of variables.

zkvm/spec/ZkVM.md
oleganza: cleartext constant != constant expression. The first is a Scalar, the second is Expression. The idea was to not have a separate instruction for mul-by-scalar distinct from mul-by-expression.

Issue #78, opened 2019-02-13
zkvm end-to-end unit tests


oleganza:
As we are approaching complete Prover/Verifier set of APIs (cf. #72), here are the unit tests that we want to make for the low-level API (#61):

1. Issue $10 (with nonce) into a pubkey.
2. Spend two utxos with $6+$4 into $9 and $1 outputs.
3. Spend 2 utxos into 1 output.
4. Spend 1 utxo into 2 outputs.
5. Spend 1 utxo into 1 output.
6. Issue (w/o nonce) $4 while spending $6 from a utxo into $9 and $1.
7. Have variants of these tests for inconsistent amounts - these should fail at verification.

cathieyun:
PR for tests 2-7 here: https://github.com/interstellar/slingshot/pull/100

cathieyun:
https://github.com/interstellar/slingshot/pull/100 merged, closing this issue!

PR #85, opened 2019-02-14
zkvm: add issuance unit test

vickiniu:
Adds issuance unit test as part of #78


zkvm/src/vm.rs
```diff
@@ -626,3 +626,40 @@ where
...
+
+        // pushdata qty
+        // var
+        // pushdata flv
+        // var
+        // pushdata pred <- pubkey to issue to
+        // issue
+        let mut program = vec![(
+            // TBD: replace with real scalars
+            Instruction::Push(Data::Witness(DataWitness::Scalar(Box::new(Scalar::one())))),
```
oleganza: Probably would be good to do these in one place: 

`impl From<Scalar> for Data`
`impl From<Predicate> for Data`
`impl From<Commitment> for Data`
etc.

oleganza: I've added these conversions here: https://github.com/interstellar/slingshot/pull/86

oleganza: Merged #86 in `main`. Please rebase and you could use these convenience methods.

zkvm/src/prover.rs
```diff
@@ -34,14 +34,16 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
...
     }
 
-    fn verify_point_op<F>(&mut self, _point_op_fn: F) -> Result<(), VMError>
+    fn verify_point_op<F>(&mut self, point_op_fn: F) -> Result<(), VMError>
     where
         F: FnOnce() -> PointOp,
     {
-        Ok(())
+        let op = point_op_fn();
+        op.verify()
```
cathieyun: I think it's better to take this out, as it impacts prover performance and should just be a debugging tool

zkvm/src/predicate.rs
```diff
@@ -121,6 +121,15 @@ impl Predicate {
     }
 }
 
+impl From<Scalar> for Predicate {
```
oleganza: Contrary to other conversions, this seems to me a little bit off. Scalar is not a _representation_ of a Predicate. It's an input of a specific subkind of a predicate. At the call site it may not be super clear why and where some scalar is being converted to with `x.into()`. 

How about something like `Predicate::from_secret_key(Scalar)` to make it more clear what's the meaning of the scalar and where it's being converted?

zkvm/src/prover.rs
```diff
@@ -34,14 +34,16 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
             Commitment::Open(w) => (w.value.into(), w.blinding),
             Commitment::Closed(_) => return Err(VMError::WitnessMissing),
         };
-        Ok(self.cs.commit(v, v_blinding))
+        let var = self.cs.commit(v, v_blinding).1;
+        Ok((com.to_point(), var))
```
cathieyun: what's the motivation for using `com.to_point()` instead of the first value of `self.cs.commit(v, v_blinding)`?

zkvm/src/types.rs
```diff
@@ -115,6 +116,26 @@ impl CommitmentWitness {
...
+    where
+        T: Into<ScalarWitness>,
+    {
+        CommitmentWitness {
+            blinding: Scalar::zero(),
+            value: x.into(),
+        }
+    }
+
+    pub fn from_secret<T>(x: T) -> Self
```
oleganza: maybe `blinded` for symmetry with `unblinded`?

zkvm/src/predicate.rs
vickiniu: I like that!

zkvm/src/prover.rs
vickiniu: ah, good catch! this is left over from debugging the blinding factor business :)

zkvm/src/types.rs
```diff
@@ -264,13 +285,20 @@ impl Data {
     }
 
     /// Encodes blinded Data values.
```
oleganza: Why is it "blinded data values"? BTW, this length prefix is only used in the output serialization, why don't we just use `data.to_bytes` there and add a length prefix where we encode the output? This method in its current shape can be confused with `to_bytes`.

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,72 @@
...
+use curve25519_dalek::scalar::Scalar;
+use merlin::Transcript;
+use std::collections::VecDeque;
+
+use zkvm::*;
+
+fn issue_contract(qty: u64, flv: Scalar, pred: &Predicate) -> Vec<Instruction> {
+    vec![
+        // pushdata qty [qty]
+        Instruction::Push(Commitment::from(CommitmentWitness::from_secret(qty)).into()),
```
oleganza: We can skip `Commitment::from(...).into()` entirely if we change this definition from:

```rust
impl From<Commitment> for DataWitness {
    fn from(x: Commitment) -> Self {
        DataWitness::Commitment(Box::new(x))
    }
}
```

into:

```rust
impl<T> From<T> for DataWitness
where
    T: Into<Commitment>,
{
    fn from(x: T) -> Self {
        DataWitness::Commitment(Box::new(x.into()))
    }
}
```

Then anything convertible into `Commitment` (including itself) will be convertible into DataWitness, and per similar definition for Data, it will be convertible to Data.

zkvm/src/types.rs
vickiniu: I think we use "blinded data values" here to refer to the fact that we only encode the blinded/opaque versions of data? open to a more descriptive comment though!

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,72 @@
...
+        // stack: qty-var
+        Instruction::Var,
+        // stack: flv, qty-var
+        Instruction::Push(Commitment::from(CommitmentWitness::unblinded(flv)).into()),
+        // stack: flv-var, qty-var
+        Instruction::Var,
+        // stack: pred, flv-var, qty-var
```
oleganza: These comments use right-to-left notation while the stack diagrams in the spec use left-to-right (top is on the right). Also would it be clearer if these comments were added on the same line as the instruction (since they denote the stack _after_ the instruction is executed)?

zkvm/tests/zkvm.rs
vickiniu: it seems like we can only have one such conversion, since this I get this error message:
```
conflicting implementations of trait `std::convert::From<_>` for type `types::DataWitness`:
```
where the definition for `Commitment` conflicts with the definition for `ScalarWitness`

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,72 @@
...
+        Instruction::Push(pred.clone().into()),
+        // stack: issue-contract
+        Instruction::Issue,
+        // stack: pred, issue-contract
+        Instruction::Push(pred.clone().into()),
```
oleganza: reusing the predicate would work for the unit test, but it's not going to be workable in real life: the nonces must be unique and by default would use random predicate keys. I'd keep that explicit in the tests to avoid confusion later. We can pass `issuance_pred` and `nonce_pred` as arguments to this method.

zkvm/tests/zkvm.rs
vickiniu: ah, got it! I changed it to using left-is-top since that seemed to be how txvm worked (cc @cathieyun ), but can modify to have the top be on the right to align with the spec

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,72 @@
...
+        // stack: nonce-contract, issue-contract
+        Instruction::Nonce,
+        // stack: issue-contract
+        Instruction::Signtx,
+        // stack: issue-payload...
```
oleganza: "issue-payload" is always going to be the "issued-value", so we can document it as such explicitly

zkvm/tests/zkvm.rs
oleganza: Oh, i probably didn't pay enough attention to internal txvm documentation. AFAIR, both TxVM and ZkVM specs use the same L-to-R stack diagram style, so i'd keep it the same in the code.

zkvm/src/types.rs
cathieyun: 👍 I was thinking that too!

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        // Generate flavor scalar
+        let mut t = Transcript::new(b"ZkVM.issue");
+        t.commit_bytes(b"predicate", pred.point().as_bytes());
+        let flavor = t.challenge_scalar(b"flavor");
+
+        let program = issue_contract(1u64, flavor, &pred);
+
+        let bp_gens = BulletproofGens::new(64, 1);
+        let txresult = Prover::build_tx(program, 0u64, 0u64, 0u64, &bp_gens);
```
oleganza: Note for later: this `0u64, 0u64, 0u64` is completely unreadable because we don't have TxHeader type. Once we have it, then `Entry::Header(u64, u64, u64)` can  be redefined to `Header(TxHeader)` and things will be cleaner. We don't have to do this in this PR.

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,72 @@
...
+            Ok(x) => x,
+        };
+        (tx, txid)
+    };
+
+    // Verify tx
+    let bp_gens = BulletproofGens::new(64, 1);
+    match Verifier::verify_tx(tx, &bp_gens) {
+        Err(err) => return assert!(false, err.to_string()),
+        Ok(_) => return,
```
oleganza: if you want to simply do nothing in a `match` branch (and maybe continue execution after it) - then you can use empty tuple:

```suggestion
        Ok(_) => (),
```

zkvm/tests/zkvm.rs
vickiniu: will leave a TBD comment!

zkvm/tests/zkvm.rs
cathieyun: oh i went back to the txvm docs and they are L-to-R, I misremembered! My bad, sorry vicki!

zkvm/src/types.rs
oleganza: To avoid confusion, i'd use "blinded" only to places where we blind Pedersen commitments. For opaque "serialized" representations we could use the "opaque" term.

zkvm/tests/zkvm.rs
oleganza: OMG. Does this test work?

vickiniu: yes!

zkvm/src/ops.rs
```diff
@@ -202,7 +202,9 @@ impl Instruction {
         match self {
             Instruction::Push(data) => {
                 write(Opcode::Push);
-                data.encode(program);
+                let mut bytes = data.to_bytes();
+                encoding::write_u32(bytes.len() as u32, program);
```
oleganza: yeah, it's good to have it here, since it's the same module where this length prefix is parsed.

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,72 @@
...
+        Instruction::Output(1),                          // stack: empty
+    ]
+}
+
+#[test]
+fn issue() {
+    let (tx, txid) = {
+        let issuance_pred = match Predicate::from_witness(PredicateWitness::Key(Scalar::random(
+            &mut rand::thread_rng(),
+        ))) {
```
oleganza: instead of `match`, you could simply do `.unwrap()` to panic on Err() and unwrap the Ok() - it's not supposed to fail ever anyway.

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,72 @@
...
+        Instruction::Var, // stack: qty-var
+        Instruction::Push(Commitment::from(CommitmentWitness::unblinded(flv)).into()), // stack: qty-var, flv
+        Instruction::Var,                                // stack: qty-var, flv-var
+        Instruction::Push(issuance_pred.clone().into()), // stack: qty-var, flv-var, pred
+        Instruction::Issue,                              // stack: issue-contract
+        Instruction::Push(nonce_pred.clone().into()),    // stack: issue-contract, pred
+        Instruction::Nonce,                              // stack: issue-contract, nonce-contract
+        Instruction::Signtx,                             // stack: issue-contract
+        Instruction::Signtx,                             // stack: issued-value
+        Instruction::Push(issuance_pred.clone().into()), // stack: issued-value, pred
```
oleganza: The issuance predicate is reused as an "account" for holding the issued token. That's totally fine, but to make the example more relevant to real usage, maybe add a third predicate: `recipient_pred`? To make it clear that we are sending funds somewhere.

oleganza:
re: https://github.com/interstellar/slingshot/pull/85#discussion_r258255260

>it seems like we can only have one such conversion, since this I get this error message:
>```
>conflicting implementations of trait `std::convert::From<_>` for type `types::DataWitness`:
>```
>where the definition for `Commitment` conflicts with the definition for `ScalarWitness`

Maybe we can instead implement `Into<DataWitness>` for ScalarWitness and CommitmentWitness, to make it work?

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,74 @@
...
+        Instruction::Signtx,                              // stack: issued-value
+        Instruction::Push(recipient_pred.clone().into()), // stack: issued-value, pred
+        Instruction::Output(1),                           // stack: empty
+    ]
+}
+
+#[test]
+fn issue() {
+    let (tx, txid) = {
+        let issuance_pred = Predicate::from_witness(PredicateWitness::Key(Scalar::random(
```
oleganza: It's a PITA that we cannot convert PredicateWitness to a point or Predicate w/o errors and streamline this code This is a "design smell" that is probably worth looking into (in a separate PR.)

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,74 @@
...
+            Ok(x) => x,
+        };
+        (tx, txid)
+    };
+
+    // Verify tx
+    let bp_gens = BulletproofGens::new(64, 1);
+    match Verifier::verify_tx(tx, &bp_gens) {
+        Err(err) => return assert!(false, err.to_string()),
+        Ok(_) => (),
```
oleganza: maybe check that txlog and txid are the same as on the prover's side?

vickiniu: we can check explicitly (esp for txlog), but txid should be guaranteed to be the same, otherwise signature verification will fail

oleganza: i'd also suggest hardcoding a hex-encoded string to compare with txid (`assert_eq`) to detect when we intentionally or accidentally break binary compatibility.

oleganza: If you need a `hex` dependency, you can do that in `dev-dependencies` so it's only available in tests. Cf. https://github.com/dalek-cryptography/ed25519-dalek/blob/master/Cargo.toml#L41-L42

zkvm/src/types.rs
```diff
@@ -115,6 +116,26 @@ impl CommitmentWitness {
         let gens = PedersenGens::default();
         gens.commit(self.value.into(), self.blinding).compress()
     }
+
+    pub fn unblinded<T>(x: T) -> Self
```
oleganza: FYI: for short trait bounds you can put them inside angle brackets directly, if it looks better. E.g.:

```suggestion
    pub fn unblinded<T: Into<ScalarWitness>>(x: T) -> Self
```

In these short methods it may be easier to read.

oleganza: (that's not a nitpick worth addressing in this PR)

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,81 @@
...
+use curve25519_dalek::scalar::Scalar;
+use hex;
+use merlin::Transcript;
+
+use zkvm::*;
+
+fn issue_contract(
+    qty: u64,
+    flv: Scalar,
+    issuance_pred: &Predicate,
```
oleganza: You are passing borrow of Predicate, but making a clone of it below. It's recommended to do the opposite: _move_ the value in and not clone. Then the user has a choice: if they do not need to reuse the instance, they can move it it without extra cloning.

https://rust-lang-nursery.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control

image

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,81 @@
...
+        let recipient_pred =
+            Predicate::from_witness(PredicateWitness::Key(Scalar::from(2u64))).unwrap();
+
+        // Generate flavor scalar
+        let mut t = Transcript::new(b"ZkVM.issue");
+        t.commit_bytes(b"predicate", issuance_pred.point().as_bytes());
+        let flavor = t.challenge_scalar(b"flavor");
+
+        // Build program
+        let program = issue_contract(1u64, flavor, &issuance_pred, &nonce_pred, &recipient_pred);
```
oleganza: Addition to the above comment: these predicates are not used again after this call, so if we make them _moved_ into the `issue_contract`, then we wouldn't need to clone them.

oleganza:
@vickiniu amazing that finally we have a full-path ZkVM test!

zkvm/tests/zkvm.rs
```diff
@@ -0,0 +1,81 @@
...
+            Commitment::from(CommitmentWitness {
+                value: qty.into(),
+                blinding: Scalar::from(1u64),
+            })
+            .into(),
+        ), // stack: qty
+        Instruction::Var, // stack: qty-var
+        Instruction::Push(Commitment::from(CommitmentWitness::unblinded(flv)).into()), // stack: qty-var, flv
+        Instruction::Var,                         // stack: qty-var, flv-var
+        Instruction::Push(issuance_pred.into()),  // stack: qty-var, flv-var, pred
```
oleganza: Fun idea: have an `ops::Builder` type so we can write `ops::build().push(issuance_pred).issue().push(nonce_pred).nonce().signtx()...` and make `push<T: Into<Data>>` so we don't even have to type `.into()` all the time.

oleganza: (Note for later)

PR #86, opened 2019-02-14
zkvm: complete data witness conversion methods

oleganza:
This adds conversion from specific witness types into Data and DataWitness and makes DataWitness support integers and closed commitments (which is valid for reblinding scenarios). Also removes `ensure_closed` as unnecessary.

### DIFF
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -284,7 +284,7 @@ impl DataWitness {
             DataWitness::Scalar(s) => {
                 let s: Scalar = (*s.clone()).into();
                 buf.extend_from_slice(&s.to_bytes())
-            },
+            }
             DataWitness::Input(b) => b.encode(buf),
         }
     }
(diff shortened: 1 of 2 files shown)


zkvm/src/types.rs
```diff
@@ -76,8 +76,8 @@ pub enum Commitment {
 pub enum DataWitness {
     Program(Vec<Instruction>),
     Predicate(Box<Predicate>),
-    Commitment(Box<CommitmentWitness>),
-    Scalar(Box<Scalar>),
+    Commitment(Box<Commitment>),
+    Scalar(Box<ScalarWitness>),
```
vickiniu: (just for my understanding) what's the situation in which we want the `SignedInteger` representation of a Scalar in DataWitness?

oleganza: I'm trying to preserve integer-ness when passing witness data to rangeproofs. So it probably makes sense to define all arithmetic ops in a way that preserve integer witnesses until they overflow. Imagine you do a bunch of arithmetic constraints on `Value.qty` and then make a payment with such quantity. The output will slap a rangeproof and if the integer was preserved all the way, then the rangeproof will receive an integer and be happy. Without trying to parse out 64 bits out of a scalar and checking if the rest is zeroes.

PR #89, opened 2019-02-15
spacesuit: fix the trait bound for the RngCore

oleganza:
This fixes the issue triggered by the recent curve25519-dalek update to v1.1.

This is the issue:
image

```rust
error[E0382]: use of moved value: `rng`
   --> src/value.rs:161:70
    |
160 |         let (q_commit, q_var) = prover.commit(self.q.into(), Scalar::random(rng));
    |                                                                             --- value moved here
161 |         let (f_commit, f_var) = prover.commit(self.f, Scalar::random(rng));
    |                                                                      ^^^ value used here after move
    |
    = note: move occurs because `rng` has type `&mut R`, which does not implement the `Copy` trait
```

An issue on curve25519-dalek: https://github.com/dalek-cryptography/curve25519-dalek/issues/232

### DIFF
--- a/spacesuit/benches/spacesuit.rs
+++ b/spacesuit/benches/spacesuit.rs
@@ -24,7 +24,7 @@ fn prove<R: Rng + CryptoRng>(
     pc_gens: &PedersenGens,
     inputs: &Vec<Value>,
     outputs: &Vec<Value>,
-    rng: &mut R,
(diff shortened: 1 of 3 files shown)


oleganza:
Henry published a v1.1.3 that reverts the API change. We should review the rng-using APIs in a more principled manner.

PR #97, opened 2019-02-20
zkvm: revamp the predicate type

oleganza:
1. Makes conversion from predicate witness to predicate non-failable, so we can use From/Into traits.
2. Clarifies distinction between the opaque predicate (as viewed by the Verifier) and opaque witness for a predicate (an opaque branch in a predicate tree as viewed by the Prover).

### DIFF
--- a/zkvm/src/predicate.rs
+++ b/zkvm/src/predicate.rs
@@ -3,7 +3,7 @@
 //! Operations:
 //! - disjunction: P = L + f(L,R)*B
 //! - program_commitment: P = h(prog)*B2
-use curve25519_dalek::ristretto::CompressedRistretto;
+use curve25519_dalek::ristretto::{RistrettoPoint,CompressedRistretto};
 use curve25519_dalek::scalar::Scalar;
 use merlin::Transcript;
 
@@ -15,17 +15,18 @@ use crate::transcript::TranscriptProtocol;
 
 /// Represents a ZkVM predicate with its optional witness data.
 #[derive(Clone, Debug)]
-pub struct Predicate {
-    point: CompressedRistretto,
-    witness: Option<PredicateWitness>,
+pub enum Predicate {
+    Compressed(CompressedRistretto),
+    Decompressed(PredicateWitness),
 }
 
 /// Prover's representation of the predicate tree with all the secret witness data.
 #[derive(Clone, Debug)]
 pub enum PredicateWitness {
+    Opaque(RistrettoPoint),
     Key(Scalar),
     Program(Vec<Instruction>),
-    Or(Box<Predicate>, Box<Predicate>),
+    Or(Box<PredicateWitness>, Box<PredicateWitness>),
 }
 
 impl Predicate {
--- a/zkvm/src/predicate.rs
+++ b/zkvm/src/predicate.rs
@@ -82,14 +82,14 @@ impl Predicate {
         }
     }
 
-    /// Creates an opaque predicate witness from a compressed point
-    pub fn opaque_witness(point: CompressedRistretto) -> Result<Self, VMError> {
+    /// Creates a predicate with witness being an opaque branch of the tree.
+    pub fn opaque_branch(point: CompressedRistretto) -> Result<Self, VMError> {
         Ok(Predicate::Witness(PredicateWitness::OpaqueBranch(
             point.decompress().ok_or(VMError::FormatError)?,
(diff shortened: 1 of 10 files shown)


zkvm/src/predicate.rs
```diff
@@ -15,125 +16,151 @@ use crate::transcript::TranscriptProtocol;
...
 
-    /// Computes a disjunction of two predicates.
-    /// TBD: push this code into to_point() impl for the witness
-    pub fn or(self, right: Predicate) -> Result<Predicate, VMError> {
-        Ok(Predicate::from_witness(PredicateWitness::Or(
-            Box::new(self),
-            Box::new(right),
-        ))?)
+    /// Creates a predicate with a signing key witness
+    pub fn from_signing_key(secret_key: Scalar) -> Self {
```
vickiniu: Similarly here, it might be more clearer to have the function name indicate that we're creating a `Witness`-type predicate (or to use `From<Scalar>` for `PredicateWitness` instead)

zkvm/src/predicate.rs
```diff
@@ -15,125 +16,151 @@ use crate::transcript::TranscriptProtocol;
...
+    Opaque(CompressedRistretto),
+
+    /// Prover's view on the predicate that has valid representation.
+    Witness(PredicateWitness),
 }
 
 /// Prover's representation of the predicate tree with all the secret witness data.
 #[derive(Clone, Debug)]
 pub enum PredicateWitness {
+    // Externally provided opaque predicate, but known to be a valid Ristretto point.
```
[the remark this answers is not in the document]
oleganza: i meant, the prover does not know the witness for the branch of the predicate tree, and it's simply provided by some counter-party as-is.

edited as:

```
    /// Representation of an opaque branch of the predicate tree
    /// (but known to be a valid Ristretto point).
```

zkvm/src/signature.rs
```diff
@@ -144,8 +144,13 @@ impl Signature {
 impl VerificationKey {
     // Constructs a VerificationKey from the private key.
     pub fn from_secret(privkey: &Scalar) -> Self {
+        VerificationKey(Self::from_secret_uncompressed(privkey).compress())
+    }
+
+    // Constructs a VerificationKey from the private key.
```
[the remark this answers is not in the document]
oleganza: +1

zkvm/src/predicate.rs
```diff
@@ -15,125 +16,151 @@ use crate::transcript::TranscriptProtocol;
...
+            primary: None,
+            secondary: Some(h),
+            arbitrary: vec![(-Scalar::one(), self.to_point())],
         }
     }
 
-    /// Encodes the Predicate in program bytecode.
-    pub fn encode(&self, prog: &mut Vec<u8>) {
-        prog.extend_from_slice(&self.point().to_bytes())
+    /// Creates an opaque predicate witness from a compressed point
```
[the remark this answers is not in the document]
oleganza: `From` trait is not failable, and we want to decompress the point if we want to make an _opaque witness_ instead of an opaque predicate.

zkvm/src/predicate.rs
oleganza: My goal was to deemphasize the PredicateWitness in order to minimize amount of entities to think about (can't make it private since it's used in the public enum Predicate, but i removed all methods from it).

oleganza: renamed PredicateWitness::Opaque into PW::OpaqueBranch

zkvm/src/predicate.rs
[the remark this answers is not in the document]
oleganza: renamed method to `Predicate::opaque_branch(point)`

zkvm/src/predicate.rs
```diff
@@ -170,8 +204,8 @@ mod tests {
         let gens = PedersenGens::default();
 
         // dummy predicates
-        let left = Predicate::opaque(gens.B.compress());
-        let right = Predicate::opaque(gens.B_blinding.compress());
+        let left = Predicate::opaque_witness(gens.B.compress()).unwrap();
```
vickiniu: `s/opaque_witness/opaque_branch`?

vickiniu: (for all of these)

oleganza: oops

Issue #90, opened 2019-02-15
zkvm: enum for const expressions


oleganza:
I think it makes sense to make operations on [constant expressions](https://github.com/interstellar/slingshot/blob/main/zkvm/spec/ZkVM.md#constant-expression) formalized with an enum: 

```rust
pub type ExpressionTerm = (r1cs::Variable, Scalar);
pub enum Expression { 
  Constant(ScalarWitness), 
  Terms(Vec<ExpressionTerm>, Option<ScalarWitness>)
}
```

Then, we'd need to define `Neg`, `Add` traits for these and use them in `neg` and `add` instructions. The idea is to preserve Expression::Constant if possible. `mul` instruction is stateful for non-constant case, so it can only be implemented directly in VM.

See also #91.

Issue #96, opened 2019-02-19
zkvm: API guide


oleganza:
The ZkVM spec covers verification of the ZkVM transactions. However, the implementation also implements a _Prover API_ that reuses VM logic in order to construct a transaction signature, program bytecode and R1CS proof. Prover uses a lot of structured data types such as FrozenContract, CommitmentWitness and others that remain opaque on the Verifier's side. We need a document that provides and overview of this architecture, in addition to the core specification.

PR #98, opened 2019-02-20
zkvm: formalize constant expressions with an enum

decentralisedkev:
Closes https://github.com/interstellar/slingshot/issues/90


oleganza:
meta-comment: these new "draft PRs" have a distinct color coding, so we no longer need "[WIP]" in the title

zkvm/src/types.rs
```diff
@@ -310,6 +323,25 @@ impl Expression {
...
+
+    fn neg(self) -> Expression {
+        match self{
+            Expression::Constant(a) => {
+                Expression::Constant(-a)
+            }, 
+            Expression::Terms(terms,assignment)=> {
+
+                for (_, n) in terms.iter_mut() {
+                    -*n;
```
oleganza: you sure it changes n in place? Maybe `*n = -n`.?

zkvm/src/types.rs
```diff
@@ -310,6 +323,25 @@ impl Expression {
...
+                }
+                Expression::Terms(terms, assignment)
```
oleganza: assignment is not negated

zkvm/src/types.rs
decentralisedkev: I impl Neg for ScalarWitness so I think it should. *n = -n is much clearer though, let me change

zkvm/src/types.rs
decentralisedkev: Would we need to add that in the `neg` section? https://github.com/interstellar/slingshot/blob/main/zkvm/spec/ZkVM.md#neg

oleganza: Spec is focused on verification only. The witness types are implementation detail. See also 
https://github.com/interstellar/slingshot/issues/96

decentralisedkev:
@oleganza Could you PTAL at the summary?

impl Add for Expression:

- Takes exp1 and exp2

If both are constant, then add them together: Constant(a+b)

If both are non-constant, concatenate terms in ex2 onto ex1. Then add assignments together.

If exp1 is constant and exp2 is non-constant, then concatenate `r1cs::Variable::One(), exp1_constant.into())` onto the terms in exp2. Then add the assignments together, where the assignment for the constant is the constant itself.

If any of the assignments in the non-constant term are None, then we return None as an assignment. Initially it was:

```
                let ass = match (ass1, ass2) {
                    (Some(a), Some(b)) => Some(a + b),
                    (Some(a), None) => Some(a),
                    (None, Some(b)) => Some(b),
                    (None, None) => None,
                };
```

zkvm/src/types.rs
```diff
@@ -136,6 +139,43 @@ impl ScalarWitness {
...
+}
+impl Add for ScalarWitness {
+    type Output = ScalarWitness;
+
+    fn add(self, rhs: ScalarWitness) -> ScalarWitness {
+        match (self, rhs) {
+            (ScalarWitness::Integer(a), ScalarWitness::Integer(b)) => {
+                let res = a + b;
+
+                if res > spacesuit::SignedInteger::from(std::u64::MAX) {
```
oleganza: oh, that's clever use of capped i128. But the result can also be lower than negated u64::MAX, so we need to check that too:

```suggestion
                let max = spacesuit::SignedInteger::from(std::u64::MAX);
                let min = -max;
                if res > max || res < min {
```

zkvm/src/types.rs
```diff
@@ -136,6 +139,43 @@ impl ScalarWitness {
...
+                    return ScalarWitness::Scalar(res.into());
```
oleganza: instead of early return, i'd use if-else:

```rust
if res > max || res < min {
    ScalarWitness::Scalar(res.into())
} else {
    ScalarWitness::Integer(res)
}
```

zkvm/src/types.rs
```diff
@@ -136,6 +139,43 @@ impl ScalarWitness {
...
+                }
+
+                ScalarWitness::Integer(res)
+            }
+            (ScalarWitness::Scalar(a), ScalarWitness::Scalar(b)) => ScalarWitness::Scalar(a + b),
+            (ScalarWitness::Integer(a), ScalarWitness::Scalar(b)) => {
+                let x: Scalar = a.into();
+                ScalarWitness::Scalar(x + b)
```
oleganza: let's keep a one-liner (similarly for another branch below):

```rust
 (ScalarWitness::Integer(a), ScalarWitness::Scalar(b)) => ScalarWitness::Scalar(Scalar::from(a) + b)
```

zkvm/src/types.rs
```diff
@@ -301,11 +341,73 @@ impl Value {
 
 impl Expression {
     pub fn constant<S: Into<Scalar>>(a: S) -> Self {
-        let a: Scalar = a.into();
+        Expression::Constant(ScalarWitness::Scalar(a.into()))
```
oleganza: u64 is `Into<Scalar>`, and its integerness will be destroyed when you do `Expression::constant(1u64)`. Maybe use `Into<ScalarWitness>` instead of `Into<Scalar>` here?

zkvm/src/types.rs
```diff
@@ -136,6 +139,43 @@ impl ScalarWitness {
...
+        }
+    }
+}
+impl Add for ScalarWitness {
+    type Output = ScalarWitness;
+
+    fn add(self, rhs: ScalarWitness) -> ScalarWitness {
+        match (self, rhs) {
+            (ScalarWitness::Integer(a), ScalarWitness::Integer(b)) => {
+                let res = a + b;
```
oleganza: This abuses `SignedInteger`'s not implementing overflow checks internally inside `Add` implementation. We'll need to fix it, but we can do that separately from this PR and get back to this line later.

zkvm/src/types.rs
```diff
@@ -301,11 +341,73 @@ impl Value {
...
+
+impl Neg for Expression {
+    type Output = Expression;
+
+    fn neg(self) -> Expression {
+        match self {
+            Expression::Constant(a) => Expression::Constant(-a),
+            Expression::Terms(mut terms, assignment) => {
+                for (_, n) in terms.iter_mut() {
+                    *n = -*n;
```
oleganza: After thinking a bit more about `map` vs `iter_mut`, IMHO it's better to do `iter_mut()` to avoid reallocations.

zkvm/src/types.rs
```diff
@@ -301,11 +341,73 @@ impl Value {
...
+        match (self, rhs) {
+            (Expression::Constant(a), Expression::Constant(b)) => Expression::Constant(a + b),
+            (Expression::Constant(a), Expression::Terms(mut terms, assignment)) => {
+                // concatenate constant term to term vector in non-constant expression
+                terms.push((r1cs::Variable::One(), a.into()));
+
+                // Add assignments
+                let ass = match assignment {
+                    Some(b) => Some(a + b),
+                    _ => None,
```
oleganza: yes, it makes sense that unknown assignment is "poisonous", making the result also unknown.

zkvm/src/types.rs
```diff
@@ -301,11 +341,73 @@ impl Value {
...
+
+    fn add(self, rhs: Expression) -> Expression {
+        match (self, rhs) {
+            (Expression::Constant(a), Expression::Constant(b)) => Expression::Constant(a + b),
+            (Expression::Constant(a), Expression::Terms(mut terms, assignment)) => {
+                // concatenate constant term to term vector in non-constant expression
+                terms.push((r1cs::Variable::One(), a.into()));
+
+                // Add assignments
+                let ass = match assignment {
```
oleganza: instead of having an intermediate variable `ass`, we can simply return this:

```rust
Expression::Terms(terms, assignment.map(|b| a+b))
```

zkvm/src/types.rs
```diff
@@ -301,11 +341,73 @@ impl Value {
...
+    }
+}
+
+impl Add for Expression {
+    type Output = Expression;
+
+    fn add(self, rhs: Expression) -> Expression {
+        match (self, rhs) {
+            (Expression::Constant(a), Expression::Constant(b)) => Expression::Constant(a + b),
+            (Expression::Constant(a), Expression::Terms(mut terms, assignment)) => {
```
oleganza: let's make it clearer (and similarly for the alternative branch):
```suggestion
            (Expression::Constant(left), Expression::Terms(mut right_terms, right_assignment)) => {
```

zkvm/src/types.rs
```diff
@@ -301,11 +341,73 @@ impl Value {
...
+                terms.push((r1cs::Variable::One(), a.into()));
+
+                // Add assignments
+                let ass = match assignment {
+                    Some(b) => Some(a + b),
+                    _ => None,
+                };
+                Expression::Terms(terms, ass)
+            }
+            (Expression::Terms(mut terms, assignment), Expression::Constant(b)) => {
```
oleganza: similar comment as above:
```suggestion
            (Expression::Terms(mut left_terms, left_assignment), Expression::Constant(right)) => {
```

zkvm/src/vm.rs
```diff
@@ -693,12 +691,21 @@ where
...
 
     fn add_range_proof(&mut self, bitrange: usize, expr: Expression) -> Result<(), VMError> {
-        spacesuit::range_proof(
-            self.delegate.cs(),
-            r1cs::LinearCombination::from_iter(expr.terms),
-            ScalarWitness::option_to_integer(expr.assignment)?,
-            bitrange,
-        )
-        .map_err(|_| VMError::R1CSInconsistency)
+        match expr {
```
oleganza: maybe factor out linear combination and assignment?

```rust
let (lc, assignment) = match expr {
    Expression::Constant(x) => (r1cs::LinearCombination::from(x), Some(x)),
    Expression::Terms(terms, assignment) => (r1cs::LinearCombination::from_iter(terms), assignment),
};
spacesuit::range_proof(
    self.delegate.cs(),
    lc, 
    ScalarWitness::option_to_integer(assignment)?,
    bitrange
).map_err(|_| VMError::R1CSInconsistency)
```

zkvm/src/types.rs
```diff
@@ -301,11 +341,73 @@ impl Value {
...
+                };
 
-        Expression {
-            terms: vec![(r1cs::Variable::One(), a)],
-            assignment: Some(ScalarWitness::Scalar(a)),
+                Expression::Terms(terms, ass)
+            }
+            (Expression::Terms(t1, a1), Expression::Terms(t2, a2)) => {
+                // concatenate terms from both non-constant expressions
+                let mut terms: Vec<ExpressionTerm> = Vec::new();
```
oleganza: you can take t1 and extend it with t2 instead of allocating a third vec.

zkvm/src/types.rs
```diff
@@ -301,11 +341,73 @@ impl Value {
...
+                terms.extend(t1);
+                terms.extend(t2);
+
+                // Add assignments
+                let ass = match (a1, a2) {
```
oleganza: Suggestion: 

```rust
Expression::Terms(terms, left.and_then(|a| right.map(|b| a+b)))
```

zkvm/src/types.rs
decentralisedkev: I'm not sure how to do it with a one liner. If I understand correctly `impl Into<Scalar> for SignedInteger` would not give me the `From` trait for free, but an `impl From` will give me the `Into` trait`?

zkvm/src/types.rs
decentralisedkev: Yep, thank you

zkvm/src/types.rs
decentralisedkev: Maybe we could make it more readable?

- `terms.iter_mut().for_each(|x| x.1 = -x.1);`

- `for x in &mut a { x.1 = -x.1 }`

-  `terms.iter_mut().for_each(|(_, n)| *n = -*n)`

The below code should test that they work as expected:

```
fn main() {
    let mut a : Vec<(u64,i32)> = vec![(1,2),(2,3),(4,5),(6,7),(8,9)];
    a.iter_mut().for_each(|x| x.1 = -x.1);
    println!("{:?}",a);
    
    let mut a : Vec<(u64,i32)> = vec![(1,12),(2,13),(4,15),(6,17),(8,19)];
    for x in &mut a { x.1 = -x.1 };
    println!("{:?}",a);
    
    let mut a : Vec<(u64,i32)> = vec![(1,12),(2,13),(4,15),(6,17),(8,19)];
    a.iter_mut().for_each(|(_, n)| *n = -*n);
    println!("{:?}",a);
}
```

WDYT should we keep the same or change?

zkvm/src/vm.rs
decentralisedkev: This is really nice +1

zkvm/src/types.rs
oleganza: Let's merge this checked arithmetic for `SignedInteger` first and then simplify this code as these checks will be done inside SignedInteger Add/Mul implementations. https://github.com/interstellar/slingshot/pull/106

zkvm/src/types.rs
oleganza: Oh, you can also try `Into<Scalar>::into(a)` if `a.into()` is ambiguous to the compiler. 

Cf. [Fully Qualified Syntax for Disambiguation](https://doc.rust-lang.org/1.22.1/book/second-edition/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation)

zkvm/src/types.rs
```diff
@@ -300,12 +342,61 @@ impl Value {
...
+    type Output = Expression;
+
+    fn add(self, rhs: Expression) -> Expression {
+        match (self, rhs) {
+            (Expression::Constant(left), Expression::Constant(right)) => {
+                Expression::Constant(left + right)
+            }
+            (Expression::Constant(left), Expression::Terms(mut right_terms, right_assignment)) => {
+                // prepend constant term to `term vector` in non-constant expression
+                right_terms.insert(0, (r1cs::Variable::One(), left.into()));
```
oleganza: We double-checked that the order of terms in the linear combination does not affect the proof, only the order of constraints themselves and order of allocated variables does. So you can do `push()` instead of `insert()` here.

oleganza: However, it's probably cleaner+safer to keep it as is — this way the result is locally stable, which is easier to reason about.

zkvm/src/types.rs
```diff
@@ -300,12 +342,61 @@ impl Value {
...
+                Expression::Terms(right_terms, right_assignment.map(|b| left + b))
```
```suggestion
                Expression::Terms(right_terms, right_assignment.map(|right| left + right))
```

zkvm/src/types.rs
```diff
@@ -300,12 +342,61 @@ impl Value {
...
+            }
+            (Expression::Terms(mut left_terms, left_assignment), Expression::Constant(right)) => {
+                // append constant term to term vector in non-constant expression
+                left_terms.push((r1cs::Variable::One(), right.into()));
+                Expression::Terms(left_terms, left_assignment.map(|a| a + right))
```
```suggestion
                Expression::Terms(left_terms, left_assignment.map(|left| left + right))
```

zkvm/src/types.rs
```diff
@@ -300,12 +342,61 @@ impl Value {
...
+            }
+            (
+                Expression::Terms(mut left_terms, left_assignment),
+                Expression::Terms(right_terms, right_assignment),
+            ) => {
+                // append right terms to left terms in non-constant expression
+                left_terms.extend(right_terms);
+                Expression::Terms(
+                    left_terms,
+                    left_assignment.and_then(|a| right_assignment.map(|b| a + b)),
```
```suggestion
                    left_assignment.and_then(|l| right_assignment.map(|r| l + r)),
```

sorry for a bikesheddy nitpick

zkvm/src/types.rs
oleganza: We can leave `for (_, n) in terms.iter_mut() { *n = -*n }` imho.

zkvm/src/types.rs
oleganza: In fact, we can probably do 2 arms in this match statement: one is for `Integer,Integer`, and the other is for all other combinations: `(a,b) => ScalarWitness::Scalar(a.into() + b.into())`.

oleganza:
#106 is merged, so we can remove unnecessary checks about `±u64::MAX`.

decentralisedkev:
> #106 is merged, so we can remove unnecessary checks about `±u64::MAX`.

refactored wrt to #106

zkvm/src/types.rs
```diff
@@ -13,6 +13,9 @@ use crate::ops::Instruction;
 use crate::predicate::Predicate;
 use crate::transcript::TranscriptProtocol;
 
+use std::ops::Add;
```
oleganza: can do these in one line:
```suggestion
use std::ops::{Neg,Add};
```

zkvm/src/types.rs
```diff
@@ -156,6 +159,35 @@ impl ScalarWitness {
...
+}
+impl Add for ScalarWitness {
+    type Output = ScalarWitness;
+
+    fn add(self, rhs: ScalarWitness) -> ScalarWitness {
+        match (self, rhs) {
+            (ScalarWitness::Integer(a), ScalarWitness::Integer(b)) => match a + b {
+                Some(res) => ScalarWitness::Integer(res),
+                None => ScalarWitness::Scalar(
+                    <SignedInteger as Into<Scalar>>::into(a)
```
oleganza: Seems like it'd be worth it defining `to_scalar(self) -> Scalar` and using it inside Into<Scalar> impl and here to make things much clearer.

zkvm/src/types.rs
```diff
@@ -322,12 +354,61 @@ impl Value {
...
+            Expression::Terms(mut terms, assignment) => {
+                for (_, n) in terms.iter_mut() {
+                    *n = -*n;
+                }
+
+                let x = match assignment {
+                    Some(a) => Some(-a),
+                    None => None,
+                };
+                Expression::Terms(terms, x)
```
oleganza: this one-liner would remove the need for the match statement:
```suggestion
                Expression::Terms(terms, assignment.map(|a| -a))
```

zkvm/src/vm.rs
```diff
@@ -292,16 +292,16 @@ where
...
-        let neg_expr = Expression {
-            terms: expr
-                .terms
-                .iter()
-                .map(|t| (t.0, t.1.neg()))
-                .collect::<Vec<_>>(),
-            assignment: expr.assignment,
-        };
-        self.push_item(neg_expr);
+        let mut expr = self.pop_item()?.to_expression()?;
```
oleganza: `Neg::neg` consumes the value, is `mut` necessary here?
 
```suggestion
        let expr = self.pop_item()?.to_expression()?;
```

zkvm/src/types.rs
```diff
@@ -49,11 +52,11 @@ pub struct Variable {
...
 
+pub type ExpressionTerm = (r1cs::Variable, Scalar);
 #[derive(Clone, Debug)]
-pub struct Expression {
-    /// Terms of the expression
-    pub(crate) terms: Vec<(r1cs::Variable, Scalar)>,
-    pub(crate) assignment: Option<ScalarWitness>,
+pub enum Expression {
+    Constant(ScalarWitness),
+    Terms(Vec<ExpressionTerm>, Option<ScalarWitness>),
```
oleganza: Maybe rename Terms to LinearCombination — seems more correct to me?

```suggestion
    LinearCombination(Vec<ExpressionTerm>, Option<ScalarWitness>),
```

zkvm/src/types.rs
```diff
@@ -322,12 +354,61 @@ impl Value {
...
-            assignment: Some(ScalarWitness::Scalar(a)),
+    fn neg(self) -> Expression {
+        match self {
+            Expression::Constant(a) => Expression::Constant(-a),
+            Expression::Terms(mut terms, assignment) => {
+                for (_, n) in terms.iter_mut() {
+                    *n = -*n;
+                }
+
+                let x = match assignment {
```
oleganza: this is dead code now

zkvm/src/types.rs
```diff
@@ -13,6 +13,9 @@ use crate::ops::Instruction;
 use crate::predicate::Predicate;
 use crate::transcript::TranscriptProtocol;
 
+use std::ops::{Neg,Add};
+use std::ops::Neg;
```
oleganza: this is unnecessary line now

spacesuit/src/value.rs
```diff
@@ -107,6 +107,10 @@ impl SignedInteger {
             Some(self.0 as u64)
         }
     }
+
+    pub fn to_scalar(self) -> Scalar {
+        <SignedInteger as Into<Scalar>>::into(self)
```
oleganza: instead of defining this in terms of Into, lets do the reverse: put the `match{...}` here and call `to_scalar` inside impl of `Into<Scalar>`.

zkvm/src/types.rs
```diff
@@ -146,6 +148,13 @@ impl ScalarWitness {
         }
     }
 
+    pub fn to_scalar(self) -> Scalar {
+        match self {
+            ScalarWitness::Integer(i) => <SignedInteger as Into<Scalar>>::into(i),
```
oleganza: `i.into()` should be unambiguous, i think, since compiler sees that the intended result is Scalar.

spacesuit/src/value.rs
```diff
@@ -107,6 +107,10 @@ impl SignedInteger {
             Some(self.0 as u64)
         }
     }
+
+    pub fn to_scalar(self) -> Scalar {
```
oleganza: do we need this method?

oleganza: I don't think we need this method. In the places where we convert SignedInteger to Scalar, the destination type is known and there is no ambiguity. So we can simply do `signed_integer.into()` there.

oleganza: _i'm checking out this branch in order to learn some Rust myself and stop adding nonsensical comments to this PR_

PR #100, opened 2019-02-20
zkvm: add spending tests

cathieyun:
https://github.com/interstellar/slingshot/issues/78 tests 2-6, in progress

### DIFF
--- a/zkvm/src/contract.rs
+++ b/zkvm/src/contract.rs
@@ -55,19 +55,6 @@ pub struct FrozenValue {
     pub(crate) flv: Commitment,
 }
 
-impl Contract {
-    pub fn min_serialized_length(&self) -> usize {
-        let mut size = 32 + 4;
-        for item in self.payload.iter() {
-            match item {
-                PortableItem::Data(d) => size += 1 + 4 + d.min_serialized_length(),
-                PortableItem::Value(_) => size += 1 + 64,
-            }
-        }
-        size
-    }
-}
-
 impl Input {
     pub fn from_bytes(data: Vec<u8>) -> Result<Self, VMError> {
         let contract = SliceReader::parse(&data, |r| Self::decode(r))?;
@@ -91,9 +78,45 @@ impl Input {
             txid,
         })
     }
+
+    // Creates a FrozenContract from payload (which is a vector of (qty, flv)) and pred.
+    // Serializes the contract, and uses the serialized contract and txid to generate a utxo.
+    // Returns an Input with the contract, txid, and utxo.
+    pub fn new(payload: Vec<(Commitment, Commitment)>, pred: Predicate, txid: TxID) -> Self {
+        let frozen_values: Vec<FrozenItem> = payload
+            .into_iter()
+            .map(|(qty, flv)| FrozenItem::Value(FrozenValue { qty: qty, flv: flv }))
+            .collect();
+
+        let frozen_contract = FrozenContract {
+            payload: frozen_values,
+            predicate: pred,
+        };
+
+        let mut contract_buf = Vec::with_capacity(frozen_contract.min_serialized_length());
+        frozen_contract.encode(&mut contract_buf);
+        let utxo = UTXO::from_output(&contract_buf, &txid);
+
+        Input {
+            contract: frozen_contract,
+            utxo: utxo,
+            txid: txid,
+        }
+    }
 }
 
 impl FrozenContract {
+    pub fn min_serialized_length(&self) -> usize {
+        let mut size = 32 + 4;
+        for item in self.payload.iter() {
+            match item {
+                FrozenItem::Data(d) => size += 1 + 4 + d.min_serialized_length(),
+                FrozenItem::Value(_) => size += 1 + 64,
+            }
+        }
+        size
+    }
+
     pub fn encode(&self, buf: &mut Vec<u8>) {
         buf.extend_from_slice(&self.predicate.to_point().to_bytes());
         encoding::write_u32(self.payload.len() as u32, buf);
--- a/zkvm/src/contract.rs
+++ b/zkvm/src/contract.rs
@@ -82,25 +82,25 @@ impl Input {
     // Creates a FrozenContract from payload (which is a vector of (qty, flv)) and pred.
     // Serializes the contract, and uses the serialized contract and txid to generate a utxo.
     // Returns an Input with the contract, txid, and utxo.
-    pub fn new(payload: Vec<(Commitment, Commitment)>, pred: Predicate, txid: TxID) -> Self {
-        let frozen_values: Vec<FrozenItem> = payload
+    pub fn new<I>(payload: I, predicate: Predicate, txid: TxID) -> Self
+    where
+        I: IntoIterator<Item = (Commitment, Commitment)>,
+    {
+        let payload: Vec<FrozenItem> = payload
             .into_iter()
-            .map(|(qty, flv)| FrozenItem::Value(FrozenValue { qty: qty, flv: flv }))
+            .map(|(qty, flv)| FrozenItem::Value(FrozenValue { qty, flv }))
             .collect();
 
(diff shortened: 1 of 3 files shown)


zkvm/tests/zkvm.rs
```diff
@@ -79,3 +89,202 @@ fn issue() {
         }
     };
 }
+
+// questions:
+// - do we have to use separate nonce predicates for the two inputs (to ensure uniqueness if they're same qty?)
```
vickiniu: My understanding is that we do need the nonce predicates to be unique to ensure uniqueness!

zkvm/tests/zkvm.rs
```diff
@@ -79,3 +89,202 @@ fn issue() {
...
+// - is it enough (for testing) to get the value from the issue contract? (how do we create a "utxo" in one tx?)
```
vickiniu: to fully test spending UTXOs, we might need two txs — one that issues the UTXOs and one that inputs them and spends them.

zkvm/tests/zkvm.rs
```diff
@@ -79,3 +89,202 @@ fn issue() {
...
+        let recipient_2_pred =
+            Predicate::from_witness(PredicateWitness::Key(Scalar::from(3u64))).unwrap();
+
+        // Generate flavor scalar
+        let mut t = Transcript::new(b"ZkVM.issue");
+        t.commit_bytes(b"predicate", issuance_pred.point().as_bytes());
+        let flavor = t.challenge_scalar(b"flavor");
+
+        // Build program
+        let program = spend_2_2_contract(
```
vickiniu: I think if we were to call issue_helper here in the test, and then build that instruction & pass the input from txlog into spend_2_2_contract that might be a good way to simulate spending UTXOs?

zkvm/tests/zkvm.rs
oleganza: if you use at least one `input`, no need for any `nonce`s. This is because inputs are identified by their txid (where their utxos were created), and txid is guaranteed to be unique recursively.

oleganza: `nonce` entry exists only for two reasons: 
1. make issuance possible when no `input`s are available 
2. have a tool to create singleton flavors (assets that can be issued exactly once with some fixed quantity)

oleganza: FYI, in Bitcoin, the "nonce" is implemented by placing block height into the coinbase tx (the "issuance transaction"):

* [BIP30](https://github.com/bitcoin/bips/blob/master/bip-0030.mediawiki) identified two cases of duplicate txids that shadowed coinbase outputs and introduced a rule to reject duplicates.
* [BIP34](https://github.com/bitcoin/bips/blob/master/bip-0034.mediawiki) introduced mandatory block height into the coinbase tx to make sure that such txs are unique w/o relying on explicit checking for duplicate txids (which does not scale).

zkvm/tests/zkvm.rs
oleganza: That'd be a nicer test, but we can as well "make up" an `Input` like if it was already created and stored in the blockchain state machine. Since we don't have this state machine to police us, we can make up any UTXOs we want and pretend they were created by prior transactions.

oleganza:
@cathieyun FYI: the merged https://github.com/interstellar/slingshot/pull/97 does a few tweaks to the Predicate API, so the tests will need to be updated accordingly. Let me know if there's any issue - i'd be glad to help!

zkvm/tests/zkvm.rs
cathieyun: thanks for the explanation, that makes sense! So that is why in the end-to-end unit tests (https://github.com/interstellar/slingshot/issues/78), the "issue w/o nonce" test is also spending from utxo, right?

oleganza: yup!

zkvm/src/contract.rs
```diff
@@ -91,9 +78,45 @@ impl Input {
...
         })
     }
+
+    // Creates a FrozenContract from payload (which is a vector of (qty, flv)) and pred.
+    // Serializes the contract, and uses the serialized contract and txid to generate a utxo.
+    // Returns an Input with the contract, txid, and utxo.
+    pub fn new(payload: Vec<(Commitment, Commitment)>, pred: Predicate, txid: TxID) -> Self {
+        let frozen_values: Vec<FrozenItem> = payload
+            .into_iter()
+            .map(|(qty, flv)| FrozenItem::Value(FrozenValue { qty: qty, flv: flv }))
```
oleganza:
```suggestion
            .map(|(qty, flv)| FrozenItem::Value(FrozenValue { qty, flv }))
```

zkvm/src/contract.rs
```diff
@@ -91,9 +78,45 @@ impl Input {
             txid,
         })
     }
+
+    // Creates a FrozenContract from payload (which is a vector of (qty, flv)) and pred.
+    // Serializes the contract, and uses the serialized contract and txid to generate a utxo.
+    // Returns an Input with the contract, txid, and utxo.
+    pub fn new(payload: Vec<(Commitment, Commitment)>, pred: Predicate, txid: TxID) -> Self {
```
oleganza: let the caller decide how to provide the payload:
```suggestion
    pub fn new<I>(payload: I, pred: Predicate, txid: TxID) -> Self 
    where I: IntoIterator<Item=(Commitment, Commitment)>
    {
```

zkvm/src/contract.rs
```diff
@@ -91,9 +78,45 @@ impl Input {
...
+        let frozen_values: Vec<FrozenItem> = payload
```
oleganza: if you shadow `payload` variable, the next line will be shorter...

```suggestion
        let payload: Vec<FrozenItem> = payload
```

zkvm/src/contract.rs
```diff
@@ -91,9 +78,45 @@ impl Input {
...
+            .into_iter()
+            .map(|(qty, flv)| FrozenItem::Value(FrozenValue { qty: qty, flv: flv }))
+            .collect();
+
+        let frozen_contract = FrozenContract {
```
oleganza: ...here:

```suggestion
        let frozen_contract = FrozenContract { payload, predicate };
```

zkvm/src/contract.rs
```diff
@@ -91,9 +78,45 @@ impl Input {
...
+            predicate: pred,
+        };
+
+        let mut contract_buf = Vec::with_capacity(frozen_contract.min_serialized_length());
+        frozen_contract.encode(&mut contract_buf);
+        let utxo = UTXO::from_output(&contract_buf, &txid);
+
+        Input {
+            contract: frozen_contract,
+            utxo: utxo,
```
oleganza:
```suggestion
            utxo,
```

zkvm/src/contract.rs
```diff
@@ -91,9 +78,45 @@ impl Input {
...
+            txid: txid,
```
oleganza:
```suggestion
            txid,
```

zkvm/tests/zkvm.rs
```diff
@@ -5,12 +5,47 @@ use merlin::Transcript;
...
+fn predicate_helper(pred_num: usize) -> (Vec<Predicate>, Scalar, Predicate) {
+    let predicates = (0..pred_num)
+        .into_iter()
+        .map(|n| Predicate::from_signing_key(Scalar::from(n as u64)))
+        .collect();
+
+    // Generate issuance predicate
+    let issuance_pred = Predicate::from_signing_key(Scalar::from(100u64));
+    // Generate flavor scalar
+    let mut t = Transcript::new(b"ZkVM.issue");
```
oleganza: why not using the existing method? `Value::issue_flavor(&issuance_pred)`

cathieyun: oh I forgot about the function, nice! updated

PR #101, opened 2019-02-20
zkvm/src: fix neg

bobg:


zkvm/src/vm.rs
```diff
@@ -291,11 +292,16 @@ where
...
-        }
-        self.push_item(expr);
+        let expr = self.pop_item()?.to_expression()?;
+        let neg_expr = Expression {
+            terms: expr
+                .terms
+                .iter()
+                .map(|t| (t.0, t.1.neg()))
+                .collect::<Vec<_>>(),
+            assignment: expr.assignment,
```
bobg: Rust newbie sez: does this move ownership of `expr.assignment`? Do I need a different way to do this?

cathieyun: my understanding is that yes, the field would take ownership of `expr.assignment`. but I don't see a problem with that - it doesn't seem like `expr.assignment` or `expr` are being used anywhere later in the code, and you already have ownership of `expr`.

oleganza: If you do into_iter, iterator will consume the storage and collect() may reuse it.

Issue #59, opened 2019-02-07
ZkVM checklist


oleganza:
This is a tracking issue for ZkVM implementation.

- [x] vm: Predicate tree: disjunction and program commitment: https://github.com/interstellar/zkvm/pull/4
- [x] vm: aggregated signature protocol: https://github.com/interstellar/zkvm/pull/4, https://github.com/interstellar/zkvm/pull/8
- [x] vm: batched verification of point operations: https://github.com/interstellar/zkvm/pull/16
- [x] vm: VM state with basic input+signtx+cloak+output flow: https://github.com/interstellar/zkvm/pull/4
- [x] vm: encoding inputs/outputs: https://github.com/interstellar/zkvm/pull/4
- [x] vm: merkle tree for txid: https://github.com/interstellar/zkvm/pull/4
- [x] api: prover: low-level API: convert a stream of `Instruction`s into a program bytecode, tx sig and r1cs proof. PR: #69, #79
- [x] api: unit tests for end-to-end tx creation/verification using low-level API: https://github.com/interstellar/slingshot/pull/100
- [ ] api: prover: high-level input-output txbuilder on top of the low-level API.
- [ ] api: prover: issuance txbuilder: add optional nonce if no inputs are present
- [ ] api: prover: retirement (add an API for destroying a value)
- [ ] api: prover: make txid-inclusion proof for chosen outputs
- [ ] vm: `left`/`right` instructions
- [ ] vm: `delegate`/`call` instructions
- [ ] vm: `contract` instruction
- [x] vm: `log` instruction https://github.com/interstellar/zkvm/pull/20
- [ ] api: txbuilder: add data-only (`log`) entry API
- [ ] api: payment channel protocol
- [ ] api: vault protocol
- [ ] vm: `mintime`/`maxtime` instructions
- [ ] vm: `qty`/`flavor` instructions
- [ ] vm: `var`/`const` instructions
- [x] vm: `neg`/`add`/`mul` instructions: https://github.com/interstellar/slingshot/pull/88, https://github.com/interstellar/slingshot/pull/101, https://github.com/interstellar/slingshot/pull/102
- [x] vm: `eq`/`and`/`or`/`verify` instructions: https://github.com/interstellar/slingshot/pull/102, https://github.com/interstellar/slingshot/pull/147
- [ ] vm: `alloc` instruction
- [ ] vm: `import`/`export` instructions
- [ ] vm: `blind`/`reblind`/`unblind` instructions

[...]

PR #102, opened 2019-02-21
zkvm: add `mul` and `eq` instructions

cathieyun:
from checklist here: https://github.com/interstellar/slingshot/issues/59


vickiniu:
drive-by comment: @oleganza added an issue for a potential change make expressions an `enum` for constant expressions & define the `Add`/`Neg` traits, which seems relevant to what you're working on here :)  https://github.com/interstellar/slingshot/issues/90

cathieyun:
Thanks for the heads up, @vickiniu - yeah that is totally relevant! Also kev might already be working on an `add` implementation here: https://github.com/interstellar/slingshot/pull/98
So I'll hold off on working on `add` and `mul` until those are resolved/merged in.

zkvm/src/vm.rs
```diff
@@ -306,6 +306,51 @@ where
...
     }
 
+    fn mul(&mut self) -> Result<(), VMError> {
+        let expr2 = self.pop_item()?.to_expression()?;
+        let expr1 = self.pop_item()?.to_expression()?;
+
+        // Note: we cannot implement `mul` for `Expression` because we need to have access
+        // to the `ConstraintSystem` API in order to make a multiplier in the
+        // `LinearCombination * LinearCombination` case.
+        let expr3 = match (expr1, expr2) {
```
oleganza: You can factor this into `Expression::multiply<CS: ConstraintSystem>(self, rhs, cs: &mut CS)` since you don't need anything else but `CS::multiply`, as far as i can see.

cathieyun: done! thanks, that's much better / more readable now

zkvm/src/constraints.rs
```diff
@@ -159,6 +208,14 @@ impl Add for Expression {
     }
 }
 
+impl Sub for Expression {
```
oleganza: I don't think we need impl of `Sub`. For `eq` instruction we need a different type `Constraint::Eq(Expr, Expr)`.

zkvm/src/vm.rs
```diff
@@ -306,6 +306,22 @@ where
...
+        let expr1 = self.pop_item()?.to_expression()?;
+        let expr3 = expr1.multiply(expr2, self.delegate.cs());
+        self.push_item(expr3);
+        Ok(())
+    }
+
+    fn eq(&mut self) -> Result<(), VMError> {
+        let expr2 = self.pop_item()?.to_expression()?;
+        let expr1 = self.pop_item()?.to_expression()?;
+        let expr3 = expr1 - expr2;
```
oleganza: See `Constraint::Eq` type. We don't need to compute difference between `Expression`s. This will happen at r1cs level within the implementation of `Constraint` (inside `verify` instruction).

oleganza:
Looks amazing. See the note on the `Constraint` type.

PR #103, opened 2019-02-21
zkvm/encoding: safer subslice parsing

### LINKED ISSUE #80: zkvm: safer parsing API with subslices

vickiniu:
Rename `Subslice` to `SliceReader` and change API for safer subslice parsing.

Fixes #80

### DIFF
--- a/zkvm/src/contract.rs
+++ b/zkvm/src/contract.rs
@@ -1,5 +1,5 @@
 use crate::encoding;
-use crate::encoding::Subslice;
+use crate::encoding::SliceReader;
 use crate::errors::VMError;
 use crate::predicate::Predicate;
 use crate::txlog::{TxID, UTXO};
@@ -70,12 +70,7 @@ impl Contract {
 
 impl Input {
     pub fn from_bytes(data: Vec<u8>) -> Result<Self, VMError> {
-        let mut reader = Subslice::new(&data);
-        let contract = Self::decode(&mut reader)?;
-        // Check that reader bytes have been fully consumed
-        if reader.len() != 0 {
-            return Err(VMError::FormatError);
-        }
+        let contract = SliceReader::parse(&data, |mut r| Self::decode(&mut r))?;
         Ok(contract)
     }
 
@@ -84,14 +79,12 @@ impl Input {
         self.contract.encode(buf);
     }
 
-    fn decode<'a>(reader: &mut Subslice<'a>) -> Result<Self, VMError> {
+    fn decode<'a>(reader: &mut SliceReader<'a>) -> Result<Self, VMError> {
         // Input  =  PreviousTxID || PreviousOutput
         // PreviousTxID  =  <32 bytes>
         let txid = TxID(reader.read_u8x32()?);
-        // Hash the contract into utxo before the subslice
-        // is advanced by the contract.decode method
-        let utxo = UTXO::from_output(&reader, &txid);
-        let contract = FrozenContract::decode(reader)?;
+        let (contract, contract_bytes) = reader.slice(|r| FrozenContract::decode(r))?;
+        let utxo = UTXO::from_output(contract_bytes, &txid);
         Ok(Input {
             contract,
             utxo,
@@ -124,7 +117,7 @@ impl FrozenContract {
         }
     }
 
-    fn decode<'a>(output: &mut Subslice<'a>) -> Result<Self, VMError> {
+    fn decode<'a>(output: &mut SliceReader<'a>) -> Result<Self, VMError> {
         //    Output  =  Predicate  ||  LE32(k)  ||  Item[0]  || ... ||  Item[k-1]
         // Predicate  =  <32 bytes>
         //      Item  =  enum { Data, Value }
--- a/zkvm/src/contract.rs
+++ b/zkvm/src/contract.rs
@@ -70,7 +70,7 @@ impl Contract {
 
 impl Input {
     pub fn from_bytes(data: Vec<u8>) -> Result<Self, VMError> {
-        let contract = SliceReader::parse(&data, |mut r| Self::decode(&mut r))?;
+        let contract = SliceReader::parse(&data, |r| Self::decode(r))?;
         Ok(contract)
     }
 
--- a/zkvm/src/encoding.rs
+++ b/zkvm/src/encoding.rs
@@ -5,20 +5,19 @@ use byteorder::{ByteOrder, LittleEndian};
 use core::ops::Range;
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
-use std::ops::Deref;
 
 use crate::errors::VMError;
 
 #[derive(Copy, Clone, Debug)]
-pub struct Subslice<'a> {
+pub struct SliceReader<'a> {
     whole: &'a [u8],
     start: usize,
     end: usize,
 }
 
-impl<'a> Subslice<'a> {
-    pub fn new(data: &'a [u8]) -> Self {
-        Subslice {
+impl<'a> SliceReader<'a> {
+    fn new(data: &'a [u8]) -> Self {
+        SliceReader {
             start: 0,
             end: data.len(),
             whole: data,
@@ -29,7 +28,7 @@ impl<'a> Subslice<'a> {
         if range.end > data.len() || range.start > data.len() {
             return Err(VMError::FormatError);
         }
-        Ok(Subslice {
+        Ok(SliceReader {
             start: range.start,
             end: range.end,
             whole: data,
@@ -44,26 +43,45 @@ impl<'a> Subslice<'a> {
         self.start..self.end
     }
 
-    /// Copies the contents of the subslice into a `Vec<u8>`.
-    pub fn to_vec(&self) -> Vec<u8> {
-        let mut buf = Vec::with_capacity(self.len());
-        buf.extend_from_slice(&self);
-        buf
+    pub fn slice<F, T>(&mut self, slice_fn: F) -> Result<(T, &[u8]), VMError>
+    where
+        F: FnOnce(&mut SliceReader) -> Result<T, VMError>,
+    {
+        let start = self.start;
+        let result = slice_fn(self)?;
+        let end = self.start;
+        Ok((result, &self.whole[start..end]))
     }
 
-    /// Returns a Subslice of the first `prefix_size` of bytes and advances
+    pub fn parse<F, T>(data: &[u8], parse_fn: F) -> Result<T, VMError>
+    where
+        F: FnOnce(SliceReader) -> Result<T, VMError>,
+    {
+        let reader = SliceReader::new(data);
+        let result = parse_fn(reader)?;
+        if reader.len() != 0 {
+            return Err(VMError::FormatError);
+        }
+        Ok(result)
+    }
+
+    pub fn skip_trailing_bytes(&mut self) -> usize {
+        unimplemented!()
+    }
+
+    /// Returns a slice of the first `prefix_size` of bytes and advances
     /// the internal offset.
-    pub fn read_bytes(&mut self, prefix_size: usize) -> Result<Subslice, VMError> {
+    pub fn read_bytes(&mut self, prefix_size: usize) -> Result<&[u8], VMError> {
         if prefix_size > self.len() {
             return Err(VMError::FormatError);
         }
-        let prefix = Subslice {
+        let prefix = SliceReader {
             start: self.start,
             end: self.start + prefix_size,
             whole: self.whole,
         };
         self.start = self.start + prefix_size;
-        Ok(prefix)
+        Ok(&prefix.whole[prefix.range()])
     }
 
     pub fn read_u8(&mut self) -> Result<u8, VMError> {
@@ -98,14 +116,6 @@ impl<'a> Subslice<'a> {
     pub fn read_scalar(&mut self) -> Result<Scalar, VMError> {
         let buf = self.read_u8x32()?;
         Scalar::from_canonical_bytes(buf).ok_or(VMError::FormatError)
-    }
-}
-
-impl<'a> Deref for Subslice<'a> {
-    type Target = [u8];
-
-    fn deref(&self) -> &[u8] {
-        &self.whole[self.start..self.end]
     }
 }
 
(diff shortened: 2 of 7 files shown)


zkvm/src/encoding.rs
```diff
@@ -5,20 +5,19 @@ use byteorder::{ByteOrder, LittleEndian};
 use core::ops::Range;
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
-use std::ops::Deref;
 
 use crate::errors::VMError;
 
 #[derive(Copy, Clone, Debug)]
```
oleganza: 1. Let's remove Copy entirely to make all clones explicit (this will break the `::parse` and hint that we probably should pass &mut to closure instead of moving the reader in it).
2. Let's _try_ to remove Clone and see if we can live without it. It would be safer if we never have to clone the reader and can use its specialized APIs to do all the things we need.

zkvm/src/encoding.rs
```diff
@@ -44,26 +43,45 @@ impl<'a> Subslice<'a> {
...
+        let start = self.start;
+        let result = slice_fn(self)?;
+        let end = self.start;
+        Ok((result, &self.whole[start..end]))
     }
 
-    /// Returns a Subslice of the first `prefix_size` of bytes and advances
+    pub fn parse<F, T>(data: &[u8], parse_fn: F) -> Result<T, VMError>
+    where
+        F: FnOnce(SliceReader) -> Result<T, VMError>,
```
oleganza: If we pass a &mut instead of move the reader in, probably we can also make this line cleaner:

before:
```rust
SliceReader::parse(&data, |mut r| Self::decode(&mut r))?
```
after:
```rust
SliceReader::parse(&data, |r| Self::decode(r))?
```

zkvm/src/encoding.rs
```diff
@@ -44,26 +43,45 @@ impl<'a> Subslice<'a> {
...
+        let reader = SliceReader::new(data);
+        let result = parse_fn(reader)?;
+        if reader.len() != 0 {
+            return Err(VMError::FormatError);
+        }
+        Ok(result)
+    }
+
+    pub fn skip_trailing_bytes(&mut self) -> usize {
+        unimplemented!()
```
oleganza: isn't it simply 

```rust
let trailing = self.end - self.start;
self.start = self.end;
trailing
```

zkvm/src/encoding.rs
```diff
@@ -44,26 +43,45 @@ impl<'a> Subslice<'a> {
...
     }
 
-    /// Copies the contents of the subslice into a `Vec<u8>`.
-    pub fn to_vec(&self) -> Vec<u8> {
-        let mut buf = Vec::with_capacity(self.len());
-        buf.extend_from_slice(&self);
-        buf
+    pub fn slice<F, T>(&mut self, slice_fn: F) -> Result<(T, &[u8]), VMError>
+    where
+        F: FnOnce(&mut SliceReader) -> Result<T, VMError>,
```
oleganza: it's more clear to use `Self` for the type of the receiver, so when the type is not the same, it's more evident.
```suggestion
        F: FnOnce(&mut Self) -> Result<T, VMError>,
```

zkvm/src/encoding.rs
oleganza:
```suggestion
        F: FnOnce(&mut Self) -> Result<T, VMError>,
```

zkvm/src/encoding.rs
```diff
@@ -44,26 +43,45 @@ impl<'a> Subslice<'a> {
...
+        Ok((result, &self.whole[start..end]))
     }
 
-    /// Returns a Subslice of the first `prefix_size` of bytes and advances
+    pub fn parse<F, T>(data: &[u8], parse_fn: F) -> Result<T, VMError>
+    where
+        F: FnOnce(SliceReader) -> Result<T, VMError>,
+    {
+        let reader = SliceReader::new(data);
+        let result = parse_fn(reader)?;
```
oleganza:
```suggestion
        let result = parse_fn(&mut reader)?;
```

zkvm/src/encoding.rs
```diff
@@ -44,26 +43,45 @@ impl<'a> Subslice<'a> {
...
+        let end = self.start;
+        Ok((result, &self.whole[start..end]))
     }
 
-    /// Returns a Subslice of the first `prefix_size` of bytes and advances
+    pub fn parse<F, T>(data: &[u8], parse_fn: F) -> Result<T, VMError>
+    where
+        F: FnOnce(SliceReader) -> Result<T, VMError>,
+    {
+        let reader = SliceReader::new(data);
```
oleganza:
```suggestion
        let mut reader = SliceReader::new(data);
```

zkvm/src/encoding.rs
```diff
@@ -44,26 +43,45 @@ impl<'a> Subslice<'a> {
...
+
+    /// Returns a slice of the first `prefix_size` of bytes and advances
     /// the internal offset.
-    pub fn read_bytes(&mut self, prefix_size: usize) -> Result<Subslice, VMError> {
+    pub fn read_bytes(&mut self, prefix_size: usize) -> Result<&[u8], VMError> {
         if prefix_size > self.len() {
             return Err(VMError::FormatError);
         }
-        let prefix = Subslice {
+        let prefix = SliceReader {
```
oleganza: We probably don't even need to construct this instance if we return a raw slice. And we can kill the `range()` method entirely too, i think.

How about:

```rust
pub fn read_bytes(&mut self, prefix_size: usize) -> Result<&[u8], VMError> {
    if prefix_size > self.len() {
        return Err(VMError::FormatError);
    }
    let prefix = self.whole[self.start..(self.start + prefix_size)];
    self.start += prefix_size;
    Ok(prefix)
}
```

zkvm/src/types.rs
```diff
@@ -230,7 +230,7 @@ impl Data {
     pub fn to_predicate(self) -> Result<Predicate, VMError> {
         match self {
             Data::Opaque(data) => {
-                let point = Subslice::new(&data).read_point()?;
+                let point = SliceReader::parse(&data, |mut r| r.read_point())?;
```
oleganza: per above comment w.r.t passing &mut reader, this could become

```suggestion
                let point = SliceReader::parse(&data, |r| r.read_point())?;
```

zkvm/src/types.rs
```diff
@@ -243,7 +243,7 @@ impl Data {
     pub fn to_commitment(self) -> Result<Commitment, VMError> {
         match self {
             Data::Opaque(data) => {
-                let point = Subslice::new(&data).read_point()?;
+                let point = SliceReader::parse(&data, |mut r| r.read_point())?;
```
oleganza: ditto

```suggestion
                let point = SliceReader::parse(&data, |r| r.read_point())?;
```

zkvm/src/encoding.rs
```diff
@@ -44,26 +43,45 @@ impl<'a> Subslice<'a> {
...
 
-    /// Returns a Subslice of the first `prefix_size` of bytes and advances
+    pub fn parse<F, T>(data: &[u8], parse_fn: F) -> Result<T, VMError>
+    where
+        F: FnOnce(SliceReader) -> Result<T, VMError>,
+    {
+        let reader = SliceReader::new(data);
+        let result = parse_fn(reader)?;
+        if reader.len() != 0 {
+            return Err(VMError::FormatError);
```
oleganza: I'd add a more specific error to indicate that there was some extra garbage in the input string to help with debugging:

```suggestion
            return Err(VMError::TrailingBytes);
```

zkvm/src/verifier.rs
```diff
@@ -55,7 +55,7 @@ impl<'a, 'b> Delegate<r1cs::Verifier<'a, 'b>> for Verifier<'a, 'b> {
         &mut self,
         run: &mut Self::RunType,
     ) -> Result<Option<Instruction>, VMError> {
-        let mut program = Subslice::new_with_range(&run.program, run.offset..run.program.len())?;
+        let mut program = SliceReader::new_with_range(&run.program, run.offset..run.program.len())?;
```
oleganza: Per suggestion (5) in https://github.com/interstellar/slingshot/issues/80 we can use `::parse` here like so:

```rust
let (instr, remainder) SliceReader::parse(&run.program[run.offset..], |r| {
    (Instruction::parse(r)?, r.skip_trailing_bytes())
})
if remainder.len() == 0 {
    return Ok(None);
}
run.offset = run.program.len() - remainder;
Ok(Some(instr))
```

zkvm/src/vm.rs
```diff
@@ -307,11 +307,7 @@ where
 
     fn r#const(&mut self) -> Result<(), VMError> {
         let data = self.pop_item()?.to_data()?.to_bytes();
-        let mut slice = Subslice::new(&data);
-        let scalar = slice.read_scalar()?;
-        if slice.len() != 0 {
-            return Err(VMError::FormatError);
-        }
+        let scalar = SliceReader::parse(&data, |mut r| r.read_scalar())?;
```
oleganza: ditto, if reader is passed by &mut

```suggestion
        let scalar = SliceReader::parse(&data, |r| r.read_scalar())?;
```

zkvm/src/contract.rs
```diff
@@ -70,12 +70,7 @@ impl Contract {
 
 impl Input {
     pub fn from_bytes(data: Vec<u8>) -> Result<Self, VMError> {
-        let mut reader = Subslice::new(&data);
-        let contract = Self::decode(&mut reader)?;
-        // Check that reader bytes have been fully consumed
-        if reader.len() != 0 {
-            return Err(VMError::FormatError);
-        }
+        let contract = SliceReader::parse(&data, |mut r| Self::decode(&mut r))?;
```
oleganza: would that work now?

```suggestion
        let contract = SliceReader::parse(&data, |r| Self::decode(r))?;
```

zkvm/src/errors.rs
```diff
@@ -15,6 +15,10 @@ pub enum VMError {
     #[fail(display = "Format in invalid")]
     FormatError,
 
+    /// This error occurs when there are trailing bytes left unread.
```
oleganza:
```suggestion
    /// This error occurs when there are trailing bytes left unread by the parser.
```

zkvm/src/verifier.rs
```diff
@@ -55,14 +55,13 @@ impl<'a, 'b> Delegate<r1cs::Verifier<'a, 'b>> for Verifier<'a, 'b> {
...
         run: &mut Self::RunType,
     ) -> Result<Option<Instruction>, VMError> {
-        let mut program = Subslice::new_with_range(&run.program, run.offset..run.program.len())?;
-
-        // Reached the end of the program - no more instructions to execute.
-        if program.len() == 0 {
+        let (instr, remainder) = SliceReader::parse(&run.program[run.offset..], |r| {
+            Ok((Instruction::parse(r)?, r.skip_trailing_bytes()))
+        })?;
+        if remainder == 0 {
```
oleganza: Sorry, my mistake. Previous code was checking for length _before_ parsing instruction. The correct way is to: 
1. check that program.offset is not at the end
2. then parse the instruction and update the offset

```rust
// Reached the end of the program - no more instructions to execute.
if run.offset == run.program.len() {
    return Ok(None);
}
let (instr, remainder) = SliceReader::parse(&run.program[run.offset..], |r| {
    Ok((
        Instruction::parse(r)?,
        r.skip_trailing_bytes()
    ))
})?;
run.offset = run.program.len() - remainder;
Ok(Some(instr))
```

PR #104, opened 2019-02-21
slidechain: delete processed exports

debnil:
Deletes rows that have been successfully processed (either success or nonretriable failure) from the `exports` table. We now only process rows that have not been pegged yet or have failed with a retriable error.

https://trello.com/c/5awzmfR4

### DIFF
--- a/slidechain/export.go
+++ b/slidechain/export.go
@@ -25,6 +25,15 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+type pegOutState int
+
+const (
+	pegOutNotYet pegOutState = iota
+	pegOutOK
+	pegOutRetry
+	pegOutFail
+)
+
 const baseFee = 100
 
 // Runs as a goroutine.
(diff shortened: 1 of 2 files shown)


slidechain/export.go
```diff
@@ -90,12 +99,27 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
 			}
 
 			log.Printf("pegging out export %x: %d of %s to %s", txid, amounts[i], asset.String(), exporters[i])
-			// TODO(vniu): flag txs that fail with unretriable errors in the db
+
+			var peggedOut pegOutState
```
vickiniu: nit: could be slightly more readable to initialize to `peggedOut = pegOutOK` and then re-assign `peggedOut` on failure cases

slidechain/export.go
```diff
@@ -25,6 +25,15 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
+type pegOutState int
+
+const (
+	pegOutNotYet pegOutState = iota
```
vickiniu: another option would be to delete the row from the peg-out table on success or failure — thoughts?

vickiniu: and then we would just have the same `unpegged` flag (and retriable errors would just not set that to be true)

slidechain/export.go
```diff
@@ -90,14 +98,27 @@ func (c *Custodian) pegOutFromExports(ctx context.Context) {
 			}
 
 			log.Printf("pegging out export %x: %d of %s to %s", txid, amounts[i], asset.String(), exporters[i])
-			// TODO(vniu): flag txs that fail with unretriable errors in the db
+
+			peggedOut := pegOutOK
```
vickiniu: I think we can get rid of the `pegOutState` enum and just have a bool for retrying now! since that is the only condition we check

Issue #91, opened 2019-02-15
zkvm: integer-preserving arithmetic on ScalarWitness


oleganza:
We want to preserve `ScalarWitness` through a number of arithmetic operations in case it ends up in a rangeproof that requires an in-range integer. Example: you compute a payment amount in a contract, and then use it for `qty` to be paid. The VM will insert a rangeproof, and (on the prover side) it won't like to receive a Scalar instead of a `u64`. So we need to preserve this and catch it early on the prover side when u64 overflows.

Suggestion:

* Implement `Neg`, `Add` and `Mul` traits for ScalarWitness.
* In `Add` and `Mul` traits check for overflows of signed integer and upgrade it to ScalarWitness::Scalar in such case.
* Review the existing code to detect cases where we do not preserve integers, but can.

oleganza:
Related: #98.

PR #106, opened 2019-02-21
spacesuit: checked arithmetic for SignedInteger

vickiniu:
Adds `checked_add` and `checked_mul` to check overflow on `SignedInteger`

Part of #91

### DIFF
--- a/spacesuit/src/value.rs
+++ b/spacesuit/src/value.rs
@@ -3,7 +3,7 @@ use core::ops::Neg;
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
 use rand::{CryptoRng, Rng};
-use std::ops::Add;
+use std::ops::{Add, Mul};
 use subtle::{Choice, ConditionallySelectable};
 
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
@@ -107,20 +107,6 @@ impl SignedInteger {
             Some(self.0 as u64)
         }
     }
-
-    pub fn checked_add(self, other: SignedInteger) -> Option<SignedInteger> {
-        match self.0.checked_add(other.0) {
-            Some(x) => Some(SignedInteger(x)),
-            None => None,
-        }
-    }
-
-    pub fn checked_mul(self, other: SignedInteger) -> Option<SignedInteger> {
-        match self.0.checked_mul(other.0) {
-            Some(x) => Some(SignedInteger(x)),
-            None => None,
-        }
-    }
 }
 
 impl From<u64> for SignedInteger {
@@ -140,10 +126,24 @@ impl Into<Scalar> for SignedInteger {
 }
 
 impl Add for SignedInteger {
-    type Output = SignedInteger;
+    type Output = Option<SignedInteger>;
 
-    fn add(self, rhs: SignedInteger) -> SignedInteger {
-        SignedInteger(self.0 + rhs.0)
+    fn add(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        match self.0.checked_add(rhs.0) {
+            Some(x) => Some(SignedInteger(x)),
+            None => None,
+        }
+    }
+}
+
+impl Mul for SignedInteger {
+    type Output = Option<SignedInteger>;
+
+    fn mul(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        match self.0.checked_mul(rhs.0) {
+            Some(x) => Some(SignedInteger(x)),
+            None => None,
+        }
     }
 }
 
--- a/spacesuit/src/value.rs
+++ b/spacesuit/src/value.rs
@@ -142,11 +142,15 @@ impl Mul for SignedInteger {
     type Output = Option<SignedInteger>;
 
     fn mul(self, rhs: SignedInteger) -> Option<SignedInteger> {
-        let p = self.0 * rhs.0;
-        if p <= (u64::max as i128) && p >= -(u64::max as i128) {
-            Some(SignedInteger(p))
-        } else {
-            None
+        match self.0.checked_mul(rhs.0) {
+            Some(p) => {
+                if p <= (u64::max as i128) && p >= -(u64::max as i128) {
+                    Some(SignedInteger(p))
+                } else {
+                    None
+                }
+            }
+            None => None,
         }
     }
 }
--- a/spacesuit/src/value.rs
+++ b/spacesuit/src/value.rs
@@ -131,9 +131,9 @@ impl Add for SignedInteger {
     fn add(self, rhs: SignedInteger) -> Option<SignedInteger> {
         let s = self.0 + rhs.0;
         if s <= (u64::max as i128) && s >= -(u64::max as i128) {
-            return Some(SignedInteger(s));
+            Some(SignedInteger(s))
         } else {
-            return None;
+            None
         }
     }
 }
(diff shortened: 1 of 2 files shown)


oleganza:
Please also see some checked arithmetic for ScalarWitness in #98 including my comments.

oleganza:
Can we avoid `checked_*` and simplн define `Add` with `Add::Output=Option<SignedInteger>`? So any attempt to add signed integers will be always checked.

spacesuit/src/value.rs
```diff
@@ -126,10 +126,24 @@ impl Into<Scalar> for SignedInteger {
 }
 
 impl Add for SignedInteger {
-    type Output = SignedInteger;
+    type Output = Option<SignedInteger>;
 
-    fn add(self, rhs: SignedInteger) -> SignedInteger {
-        SignedInteger(self.0 + rhs.0)
+    fn add(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        match self.0.checked_add(rhs.0) {
```
oleganza: `checked_add` here is "checked" w.r.t. i128 overflows, not our funky 65-bit integer. Since we know that our i128 are guaranteed to be in 64-bit range by absolute value, they will never overflow 128 bits, so we can:

1. simply add them together
2. check that the result is ≤ than `(u64::MAX as i128)` and ≥ than `-(u64::MAX as i128)`.

spacesuit/src/value.rs
```diff
@@ -126,10 +126,24 @@ impl Into<Scalar> for SignedInteger {
...
+            None => None,
+        }
+    }
+}
+
+impl Mul for SignedInteger {
+    type Output = Option<SignedInteger>;
+
+    fn mul(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        match self.0.checked_mul(rhs.0) {
```
oleganza: Same as above, `checked_mul` is protecting against i128 overflows. However, `(2^64-1)^2` is larger than `2^127-1`, so we must use `checked_mul` here. But in addition to that we need to check that the resulting `x` is in good range ±(2^64-1) as we would do with `Add`.

spacesuit/src/value.rs
```diff
@@ -126,10 +126,26 @@ impl Into<Scalar> for SignedInteger {
...
 
 impl Add for SignedInteger {
-    type Output = SignedInteger;
+    type Output = Option<SignedInteger>;
 
-    fn add(self, rhs: SignedInteger) -> SignedInteger {
-        SignedInteger(self.0 + rhs.0)
+    fn add(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        let s = self.0 + rhs.0;
+        if s <= (u64::max as i128) && s >= (u64::min as i128) {
```
oleganza: `u64::min` is 0 and `i64::min` is not what we need. We need exactly `-(u64::MAX as i128)` for minimum (first cast `2^64-1` to i128, then negate it to get `-(2^64-1)`).

spacesuit/src/value.rs
```diff
@@ -126,10 +126,26 @@ impl Into<Scalar> for SignedInteger {
...
+        None
+    }
+}
+
+impl Mul for SignedInteger {
+    type Output = Option<SignedInteger>;
+
+    fn mul(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        let p = self.0 * rhs.0;
+        if p <= (u64::max as i128) && p >= (u64::min as i128) {
```
oleganza: ditto

spacesuit/src/value.rs
```diff
@@ -126,10 +126,26 @@ impl Into<Scalar> for SignedInteger {
...
+            return Some(SignedInteger(p));
+        }
+        None
```
oleganza: symmetric if-then-else is easier to read here IMHO:

```rust
if ... {
    Some(...)
} else {
    None
}
```

spacesuit/src/value.rs
```diff
@@ -126,10 +126,28 @@ impl Into<Scalar> for SignedInteger {
...
 impl Add for SignedInteger {
-    type Output = SignedInteger;
+    type Output = Option<SignedInteger>;
 
-    fn add(self, rhs: SignedInteger) -> SignedInteger {
-        SignedInteger(self.0 + rhs.0)
+    fn add(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        let s = self.0 + rhs.0;
+        if s <= (u64::max as i128) && s >= -(u64::max as i128) {
+            return Some(SignedInteger(s));
```
oleganza: no need for `return` and a semicolon since the if/else statement is the last one in the function

```suggestion
            Some(SignedInteger(s))
```

Same for the `mul`.

spacesuit/src/value.rs
```diff
@@ -126,10 +126,28 @@ impl Into<Scalar> for SignedInteger {
...
+            None
+        }
+    }
+}
+
+impl Mul for SignedInteger {
+    type Output = Option<SignedInteger>;
+
+    fn mul(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        let p = self.0 * rhs.0;
```
oleganza: Sorry, i probably was not clear enough in https://github.com/interstellar/slingshot/pull/106#discussion_r259173002

> However, `(2^64-1)^2` is larger than `2^127-1`, so we must use `checked_mul` here.

So we not only need to check that `p` is within `±u64::max`, but also that the multiplication did not overflow i128.

```suggestion
        if let Some(p) = self.0.checked_mul(&rhs.0) {
```

oleganza:
```rust
if check i128 overflow {
    if check ±u64 overflow {
        Some(...)
    } else {
        None
    }
} else {
    None
}
```

spacesuit/src/value.rs
```diff
@@ -126,10 +126,32 @@ impl Into<Scalar> for SignedInteger {
...
 
 impl Add for SignedInteger {
-    type Output = SignedInteger;
+    type Output = Option<SignedInteger>;
 
-    fn add(self, rhs: SignedInteger) -> SignedInteger {
-        SignedInteger(self.0 + rhs.0)
+    fn add(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        let s = self.0 + rhs.0;
+        if s <= (u64::max as i128) && s >= -(u64::max as i128) {
```
oleganza: oops, this `as i128` conversion is unsafe. The `max` is actually a function and its pointer is being cast into i128 silently. You need `u64::max_value()` instead.

```
std::u64::MAX    = 18446744073709551615
u64::max_value() = 18446744073709551615
u64::max as i128 = 94475748142400
std::u64::MIN    = 0
u64::min_value() = 0
u64::min as i128 = 94475748142592
```

spacesuit/src/value.rs
```diff
@@ -126,10 +126,32 @@ impl Into<Scalar> for SignedInteger {
...
+        }
+    }
+}
+
+impl Mul for SignedInteger {
+    type Output = Option<SignedInteger>;
+
+    fn mul(self, rhs: SignedInteger) -> Option<SignedInteger> {
+        match self.0.checked_mul(rhs.0) {
+            Some(p) => {
```
oleganza: you can place if statement into the match arm to make it more succinct.
```rust
match self.0.checked_mul(rhs.0) {
    Some(p) if p <= (u64::max as i128) && p >= -(u64::max as i128) => Some(SignedInteger),
    _ => None
}
```

spacesuit/src/value.rs
vickiniu: oh, man! that would've been bad 😮

PR #107, opened 2019-02-22
slidechain: test different input and export amounts

debnil:
Tests different amounts for input and export in the end-to-end test. Simplifies check for export tx to use `wait` logic in submission, instead of a hand-rolled function.

### DIFF
--- a/slidechain/slidechain_test.go
+++ b/slidechain/slidechain_test.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"database/sql"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -412,7 +411,7 @@ func TestEndToEnd(t *testing.T) {
 			}
 			defer resp.Body.Close()
 			if resp.StatusCode/100 != 2 {
-				t.Fatalf("status code %d from POST /submit", resp.StatusCode)
+				t.Fatalf("status code %d from POST /submit?wait=1", resp.StatusCode)
 			}
 
 			// Check for successful retirement.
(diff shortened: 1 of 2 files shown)


slidechain/slidechain_test.go
```diff
@@ -265,209 +265,222 @@ func TestImport(t *testing.T) {
...
+			}
+			root, err := hclient.Root()
+			if err != nil {
+				t.Fatalf("error getting horizon client root: %s", err)
+			}
+			accountID, seed, err := custodianAccount(ctx, db, hclient)
+			if err != nil {
+				t.Fatalf("error creating custodian account: %s", err)
+			}
+			c := &Custodian{
```
vickiniu: Could we factor out the custodian launching + exporter setup so we only do that once, and use the same custodian / exporter for all the test cases?

slidechain/slidechain_test.go
```diff
@@ -294,178 +302,183 @@ func TestEndToEnd(t *testing.T) {
...
+			inputAmount := tt.inputAmount
+			exportAmount := tt.exportAmount
+			exporterPub, exporterPrv, err := ed25519.GenerateKey(nil)
+			if err != nil {
+				t.Fatalf("error generating txvm recipient keypair: %s", err)
+			}
+			var exporterSeed [32]byte
+			copy(exporterSeed[:], exporterPrv)
+			exporter, err := keypair.FromRawSeed(exporterSeed)
+			err = stellar.FundAccount(exporter.Address())
```
vickiniu: Could we also move this exporter account creation + funding outside of the test loop?

slidechain/slidechain_test.go
```diff
@@ -266,8 +266,16 @@ func TestImport(t *testing.T) {
 }
 
 func TestEndToEnd(t *testing.T) {
-	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
```
vickiniu: Just curious — is 1 minute no longer enough time to run this test?

slidechain/slidechain_test.go
```diff
@@ -312,161 +318,167 @@ func TestEndToEnd(t *testing.T) {
...
+			t.Log("building export tx...")
+			exportTx, err := BuildExportTx(ctx, native, int64(exportAmount), int64(inputAmount), temp, anchor, exporterPrv, seqnum)
+			if err != nil {
+				t.Fatalf("error building retirement tx %s", err)
+			}
+			txbits, err := proto.Marshal(&exportTx.RawTx)
+			if err != nil {
+				t.Fatal(err)
+			}
+			resp, err := http.Post(sv.URL+"/submit", "application/octet-stream", bytes.NewReader(txbits))
```
vickiniu: Do we want to replace this with the same `/submit?wait=1` call?

slidechain/slidechain_test.go
debnil: The test for a single input case takes ~40 seconds, so no. However, the 3 is a vestige from when the goroutines were being launched anew for each test case. Now, the end-to-end takes ~90s consistently, so I'm happy to just make this 2 minutes.

vickiniu: Ah, yes! Adding more test cases answers my question :) I think keeping it at 3 is probably still a totally reasonable limit

PR #108, opened 2019-02-22
slidechain: clean up export cmd

debnil:
Simplifies checking for the export tx on TxVM in `cmd/export`: rather than repeatedly checking new blocks for the tx, it simply calls the `wait` logic within the `submit` endpoint. Renames `mustDecode` to `mustDecodeHex`, in line with the rest of the codebase. De-exports the `isExportTx` check.

### DIFF
--- a/slidechain/cmd/export/export.go
+++ b/slidechain/cmd/export/export.go
@@ -85,9 +85,9 @@ func main() {
 	}
 	*slidechaind = strings.TrimRight(*slidechaind, "/")
 
-	// Build + submit pre-export tx
+	// Build and submit the pre-export transaction.
 
-	// Check that stellar account exists
+	// Check that stellar account exists.
 	var seed [32]byte
 	rawbytes := mustDecodeHex(*prv)
 	copy(seed[:], rawbytes)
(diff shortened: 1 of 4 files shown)


slidechain/cmd/export/export.go
```diff
@@ -126,73 +124,21 @@ func main() {
...
-	if err != nil {
-		log.Fatalf("error reading block from response body: %s", err)
-	}
-	err = block.FromBytes(bits)
-	if err != nil {
-		log.Fatalf("error unmarshaling block: %s", err)
-	}
-
-	resp, err = http.Post(*slidechaind+"/submit", "application/octet-stream", bytes.NewReader(txbits))
+	resp, err = http.Post(*slidechaind+"/submit?wait=1", "application/octet-stream", bytes.NewReader(txbits))
```
vickiniu: having the `/submit?wait=1` is awesome! maybe we could add a comment here to explain that this endpoint will submit the tx & block until it's included in the txvm chain (or return error)

vickiniu: to that effect, might also be good to build an `*http.Request` so that we can pass in a context object with some timeout to avoid blocking forever in an error case where the tx isn't included

PR #113, opened 2019-02-22
zkvm: add a builder API for assembling programs

### LINKED ISSUE #111: zkvm: program builder API
To make use of low-level API nicer, we can use a [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern):

```rust
// ops.rs
pub struct Program(Vec<Instruction>);
```

```rust
// some_test.rs
use ops::Program;
...
let instructions: Vec<Instruction> = Program::new()
.push(Input)
.input()
.signtx()
.push(predicate)
.issue()
.signtx()
...
.to_vec(); // converts to Vec<Instruction>
```

This allows us to avoid mouthful `Instruction::` prefixes and having to do `.into()` all the time, as we can define generic `push<T: Into<Data>>(T)` and call `.into()` inside the `push()`.

We can also pass `&mut program` to the helper functions just like we pass &mut transcript: the helper will simply add more instructions to the instance, w/o worrying who and when instantiated it.

We can also add "macros" via extension traits, e.g. a hypothetical `.spend_input(Input)` may do `.push(Input).input().signtx()`. We don't have to worry about this now, we'll see later which macros make sense and declutter the code, and which add unnecessary cognitive burden.

(based on the comment: https://github.com/interstellar/slingshot/pull/85/files#r259096201)

decentralisedkev:
Closes #111.

### DIFF
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -358,6 +358,7 @@ impl Program {
         self.0
     }
 
+    /// Adds a `push` instruction with an immediate data type that can be converted into `Data`.
     pub fn push<T: Into<Data>>(&mut self, data: T) -> &mut Program {
         self.0.push(Instruction::Push(data.into()));
         self
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -6,6 +6,8 @@ use crate::encoding::SliceReader;
 use crate::errors::VMError;
 use crate::types::Data;
 
+/// A builder type for assembling a sequence of `Instruction`s with chained method calls.
+/// E.g. `let prog = Program::new().push(...).input().push(...).output(1).to_vec()`.
 pub struct Program(Vec<Instruction>);
 
 #[derive(Clone, Debug)]
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -8,7 +8,7 @@ use crate::types::Data;
 
 /// A builder type for assembling a sequence of `Instruction`s with chained method calls.
 /// E.g. `let prog = Program::new().push(...).input().push(...).output(1).to_vec()`.
-#[derive(Clone,Debug)]
+#[derive(Clone, Debug)]
 pub struct Program(Vec<Instruction>);
 
 #[derive(Clone, Debug)]
@@ -283,7 +283,18 @@ impl Instruction {
     }
 }
 
+macro_rules! def_op {
+    ($func_name:ident, $op:ident) => (
+           pub fn $func_name(&mut self) -> &mut Program{
+             self.0.push(Instruction::$op);
+             self
+        }
+    );
+}
+
 impl Program {
+    def_op!(var, Var);
+
     pub fn new() -> Self {
         Program(vec![])
     }
@@ -314,11 +325,6 @@ impl Program {
 
     pub fn r#const(&mut self) -> &mut Program {
         self.0.push(Instruction::Const);
-        self
-    }
-
-    pub fn var(&mut self) -> &mut Program {
-        self.0.push(Instruction::Var);
         self
     }
 
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -293,4 +293,20 @@ impl Program {
         self.0.push(Instruction::Var);
         self
     }
+    fn nonce<'a>(&'a mut self) -> &'a mut Program {
+        self.0.push(Instruction::Nonce);
+        self
+    }
+    fn issue<'a>(&'a mut self) -> &'a mut Program {
+        self.0.push(Instruction::Issue);
+        self
+    }
+    fn sign_tx<'a>(&'a mut self) -> &'a mut Program {
+        self.0.push(Instruction::Signtx);
+        self
+    }
+    fn output<'a>(&'a mut self, size: usize) -> &'a mut Program {
+        self.0.push(Instruction::Output(size));
+        self
+    }
 }
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -289,23 +289,23 @@ impl Program {
         self.0.push(Instruction::Push(data.into()));
         self
     }
-    fn var<'a>(&'a mut self) -> &'a mut Program {
+    fn var(&mut self) -> &mut Program {
         self.0.push(Instruction::Var);
         self
     }
-    fn nonce<'a>(&'a mut self) -> &'a mut Program {
+    fn nonce(&mut self) -> &mut Program {
         self.0.push(Instruction::Nonce);
         self
     }
-    fn issue<'a>(&'a mut self) -> &'a mut Program {
+    fn issue(&mut self) -> &mut Program {
         self.0.push(Instruction::Issue);
         self
     }
-    fn sign_tx<'a>(&'a mut self) -> &'a mut Program {
+    fn sign_tx(&mut self) -> &mut Program {
         self.0.push(Instruction::Signtx);
         self
     }
-    fn output<'a>(&'a mut self, size: usize) -> &'a mut Program {
+    fn output(&mut self, size: usize) -> &mut Program {
         self.0.push(Instruction::Output(size));
         self
     }
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -342,6 +342,8 @@ impl Program {
         Program(vec![])
     }
 
+    /// Creates an empty `Program` and passes its &mut to the closure to let it add the instructions.
+    /// Returns the resulting program.
     pub fn build<F>(builder: F) -> Self
     where
         F: FnOnce(&mut Self) -> &mut Self,
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -293,24 +293,169 @@ impl Program {
         self.0.push(Instruction::Push(data.into()));
         self
     }
+
+    pub fn output(&mut self, size: usize) -> &mut Program {
+        self.0.push(Instruction::Output(size));
+        self
+    }
+
+    pub fn drop(&mut self) -> &mut Program {
+        self.0.push(Instruction::Drop);
+        self
+    }
+
+    pub fn r#const(&mut self) -> &mut Program {
+        self.0.push(Instruction::Const);
+        self
+    }
+
     pub fn var(&mut self) -> &mut Program {
         self.0.push(Instruction::Var);
         self
     }
-    pub fn nonce(&mut self) -> &mut Program {
-        self.0.push(Instruction::Nonce);
+
+    pub fn alloc(&mut self) -> &mut Program {
+        self.0.push(Instruction::Alloc);
         self
     }
+
+    pub fn mintime(&mut self) -> &mut Program {
+        self.0.push(Instruction::Mintime);
+        self
+    }
+
+    pub fn maxtime(&mut self) -> &mut Program {
+        self.0.push(Instruction::Maxtime);
+        self
+    }
+
+    pub fn expr(&mut self) -> &mut Program {
+        self.0.push(Instruction::Expr);
+        self
+    }
+
+    pub fn neg(&mut self) -> &mut Program {
+        self.0.push(Instruction::Neg);
+        self
+    }
+
+    pub fn add(&mut self) -> &mut Program {
+        self.0.push(Instruction::Add);
+        self
+    }
+
+    pub fn mul(&mut self) -> &mut Program {
+        self.0.push(Instruction::Mul);
+        self
+    }
+
+    pub fn eq(&mut self) -> &mut Program {
+        self.0.push(Instruction::Eq);
+        self
+    }
+
+    pub fn and(&mut self) -> &mut Program {
+        self.0.push(Instruction::And);
+        self
+    }
+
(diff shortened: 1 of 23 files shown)


zkvm/src/ops.rs
```diff
@@ -277,3 +279,18 @@ impl Instruction {
         }
     }
 }
+
+impl Program {
+    fn new() -> Self {
+        Program(vec![])
+    }
+
+    fn push<'a, T: Into<Data>>(&'a mut self, data: T) -> &'a mut Program {
```
oleganza: Are you sure `'a` is not deduced from `&self` automatically by the compiler?

```suggestion
    fn push<T: Into<Data>>(&mut self, data: T) -> &mut Program {
```

decentralisedkev:
Majority of methods were auto-generated, so no problem to rip them out if needed

zkvm/src/ops.rs
```diff
@@ -6,6 +6,8 @@ use crate::encoding::SliceReader;
 use crate::errors::VMError;
 use crate::types::Data;
 
+pub struct Program(Vec<Instruction>);
```
oleganza: in the spirit of upcoming https://github.com/interstellar/slingshot/issues/114 let's add docs to the things.

```suggestion
/// A builder type for assembling a sequence of `Instruction`s with chained method calls.
/// E.g. `let prog = Program::new().push(...).input().push(...).output(1).to_vec()`.
pub struct Program(Vec<Instruction>);
```

zkvm/src/ops.rs
```diff
@@ -277,3 +279,213 @@ impl Instruction {
         }
     }
 }
+
+impl Program {
+    pub fn new() -> Self {
+        Program(vec![])
+    }
+
+    pub fn to_vec(&self) -> Vec<Instruction> {
```
oleganza: let's consume `self` here and avoid cloning a vector. The `Program` type can get `#[derive(Clone,Debug)]` so that the user can clone it if needed.

zkvm/src/ops.rs
```diff
@@ -277,3 +279,213 @@ impl Instruction {
...
+        self.0.push(Instruction::Dup(size));
+        self
+    }
+
+    pub fn roll(&mut self, size: usize) -> &mut Program {
+        self.0.push(Instruction::Roll(size));
+        self
+    }
+
+    pub fn r#const(&mut self) -> &mut Program {
```
oleganza: Can we have a macro for simple opcodes that would create one-liners like 

```rust
def_op!(r#const, Const)
def_op!(var, Var)
def_op!(alloc, Alloc)
def_op!(mintime, Mintime)
def_op!(maxtime, Maxtime)
...
```

zkvm/src/ops.rs
```diff
@@ -277,3 +279,213 @@ impl Instruction {
...
+        self.0.push(Instruction::Right);
+        self
+    }
+
+    pub fn delegate(&mut self) -> &mut Program {
+        self.0.push(Instruction::Delegate);
+        self
+    }
+
+    pub fn ext(&mut self, n: u8) -> &mut Program {
```
oleganza: You are not supposed to use extension opcodes until they are actually defined. So i'd remove this from the API.

zkvm/tests/zkvm.rs
```diff
@@ -12,26 +12,24 @@ fn issue_contract(
...
-        Instruction::Issue,                       // stack: issue-contract
-        Instruction::Push(nonce_pred.into()),     // stack: issue-contract, pred
-        Instruction::Nonce,                       // stack: issue-contract, nonce-contract
-        Instruction::Signtx,                      // stack: issue-contract
-        Instruction::Signtx,                      // stack: issued-value
-        Instruction::Push(recipient_pred.into()), // stack: issued-value, pred
-        Instruction::Output(1),                   // stack: empty
-    ]
+    let mut instructions = Program::new();
+    instructions
```
oleganza: Why extra variable? Can just do a single chain ` Program::new().push().var()....to_vec()`.

zkvm/src/ops.rs
decentralisedkev: would we need to change the name to `into_vec` ?

decentralisedkev: We may need to change the methods for the instructions to take self also as they currently mutably borrow, which would give a compile error if we use to_vec because we are moving out of borrowed content.

If we keep it the same, I think it would work if we did:

let mut instr = Program::new().sign_tx()...;
instr.to_vec();

If the instruction methods take self, I think that we would not be able to do:

let mut instr = Program::new()
instr.push(..);
instr.sign_tx();
instr.var();

it would work if we:

let mut instr = Program::new()
instr = instr.push(..);
instr = instr.sign_tx();
instr = instr.var();

oleganza: Crazy idea: we can make these methods placed in a trait generic over BorrowMut<Program>, so that you could call them on both &mut and owned program. Then, if you started a chain with an owned program, it’d be able to consume self with to_vec().

oleganza: Let’s go with &mut in methods and move in to_vec, so you have to use a variable to move the vector out (your first snippet).

zkvm/src/ops.rs
```diff
@@ -277,3 +282,90 @@ impl Instruction {
...
+    ($func_name:ident, $op:ident) => (
+           pub fn $func_name(&mut self) -> &mut Program{
+             self.0.push(Instruction::$op);
+             self
+        }
+    );
+}
+
+impl Program {
+    def_op!(drop, Drop);
```
oleganza: Maybe for clarity we can order all ops in the canonical order, and have an extra macro `def_op_usize` for dup/roll/output/contract?

zkvm/src/ops.rs
```diff
@@ -277,3 +282,76 @@ impl Instruction {
...
 }
+
+macro_rules! def_op {
+    ($func_name:ident, $op:ident) => (
+           pub fn $func_name(&mut self) -> &mut Program{
+             self.0.push(Instruction::$op);
+             self
+        }
+    );
+    ($func_name:ident, $op:ident, $type:ty) => (
```
oleganza: clever!

zkvm/src/ops.rs
```diff
@@ -277,3 +282,76 @@ impl Instruction {
...
+           pub fn $func_name(&mut self, size :$type) -> &mut Program{
```
oleganza: it's not always the "size", i'd rename it to something more generic like "arg".

zkvm/src/ops.rs
```diff
@@ -277,3 +282,76 @@ impl Instruction {
...
+    def_op!(reblind, Reblind);
+    def_op!(retire, Retire);
+    def_op!(right, Right);
+    def_op!(roll, Roll, usize);
+    def_op!(sign_tx, Signtx);
+    def_op!(unblind, Unblind);
+    def_op!(var, Var);
+    def_op!(verify, Verify);
+
+    pub fn new() -> Self {
```
oleganza: how about in addition to `new()` we'd have `build()` that takes `FnOnce(&mut Self)` closure and allows composing things w/o extra variable: 

```rust
let instructions: Vec<Instruction> = Program::build(|p| {
    p
    .push()
    .input()
    .signtx()
}).to_vec()
```

WDYT?

oleganza: `build()` would pass &mut Self into closure, but would create and return `self`, so `to_vec` could consume it.

zkvm/tests/zkvm.rs
```diff
@@ -39,71 +39,56 @@ fn test_helper(program: Vec<Instruction>) -> Result<TxID, VMError> {
...
-        Instruction::Signtx,                     // stack: issued-value
-    ]
+) {
+    program
+        .push(Commitment::from(CommitmentWitness {
+            value: qty.into(),
+            blinding: Scalar::from(1u64),
+        })) // stack: qty
+        .var() // stack: qty-var
+        .push(Commitment::from(CommitmentWitness::unblinded(flv))) // stack: qty-var, flv
```
oleganza: suggestion: compute `flv` from `issuance_pred` using `Value::issue_flavor()`, so we don't have to pass it by argument at all.

zkvm/tests/zkvm.rs
```diff
@@ -39,71 +39,56 @@ fn test_helper(program: Vec<Instruction>) -> Result<TxID, VMError> {
...
-                )],
-                pred,
-                TxID([0; 32]),
-            )
-            .into(),
-        ), // stack: input-data
-        Instruction::Input,  // stack: input-contract
-        Instruction::Signtx, // stack: input-value
-    ]
+fn input_helper(program: &mut Program, qty: u64, flv: Scalar, pred: Predicate) {
```
oleganza: Suggestion: all `*_helper(&mut Program, ...)` methods could be more neatly organized using an _extension trait_:

1. Create a trait in the test module `ProgramHelper`.
2. Define methods such as `issue_helper(&mut self, ...)`, `input_helper(&mut self, ...)` etc that act like simpler instructions: returning `&mut Program` to allow chaining.
3. `impl ProgramHelper for Program {}` to attach these methods to Program.
4. Use these methods in a chained fashion in the test w/o extra variables — this will be a cleaner composition.

oleganza: Bonus points if there's a better name that `_helper`. Maybe `issue_value`, to not clash with built-in instructions, but still provide a meaningful name.

zkvm/tests/zkvm.rs
decentralisedkev: Ew not sure I understand, `Value::issue_flavor()` gives a Scalar?

oleganza: Yes. It's the definition of a flavor based on the issuance predicate.

Spec: https://github.com/interstellar/slingshot/blob/main/zkvm/spec/ZkVM.md#issue

image

zkvm/src/ops.rs
```diff
@@ -277,3 +282,85 @@ impl Instruction {
...
```
```suggestion
    /// Creates an empty `Program`.
    pub fn new() -> Self {
```

zkvm/src/ops.rs
```diff
@@ -277,3 +282,85 @@ impl Instruction {
...
+        Program(vec![])
+    }
+
+    pub fn build<F>(builder: F) -> Self
```
```suggestion
    /// Creates an empty `Program` and passes its &mut to the closure to let it add the instructions.
    /// Returns the resulting program.
    pub fn build<F>(builder: F) -> Self
```

zkvm/src/ops.rs
```diff
@@ -277,3 +282,85 @@ impl Instruction {
...
+    where
+        F: FnOnce(&mut Self) -> &mut Self,
+    {
+        let mut program = Self::new();
+        builder(&mut program);
+        program
+    }
+
+    pub fn to_vec(self) -> Vec<Instruction> {
```
```suggestion
    /// Converts the program to a plain vector of instructions.
    pub fn to_vec(self) -> Vec<Instruction> {
```

zkvm/src/ops.rs
```diff
@@ -277,3 +282,85 @@ impl Instruction {
...
+        self.0
+    }
+
+    pub fn push<T: Into<Data>>(&mut self, data: T) -> &mut Program {
```
```suggestion
    /// Adds a `push` instruction with an immediate data type that can be converted into `Data`.
    pub fn push<T: Into<Data>>(&mut self, data: T) -> &mut Program {
```

zkvm/tests/zkvm.rs
```diff
@@ -4,6 +4,79 @@ use hex;
...
+    ) -> &mut Self;
+
+    fn input_helper(&mut self, qty: u64, flv: Scalar, pred: Predicate) -> &mut Self;
+
+    fn cloak_helper(&mut self, input_count: usize, outputs: Vec<(u64, Scalar)>) -> &mut Self;
+
+    fn output_helper(&mut self, pred: Predicate) -> &mut Self;
+}
+
+impl ProgramHelper for Program {
```
oleganza: _Extension traits_ are called so because you can implement the methods _inside the trait_, so the user can attach the trait w/o implementing anything: `impl ExtTrait for MyThing {}`. Here it works either way, but it's cleaner to declare the methods inside the trait to show that the trait is not the abstract interface, but simply a collection of add-on functionality.

zkvm/tests/zkvm.rs
```diff
@@ -38,85 +111,17 @@ fn test_helper(program: Vec<Instruction>) -> Result<TxID, VMError> {
...
     flv: Scalar,
     issuance_pred: Predicate,
     nonce_pred: Predicate,
     output_pred: Predicate,
 ) -> Vec<Instruction> {
-    let mut instructions = vec![];
-    instructions.append(&mut issue_helper(qty, flv, issuance_pred, nonce_pred)); // stack: issued-val
-    instructions.append(&mut output_helper(output_pred)); // stack: empty
-    instructions
+    let mut program = Program::new();
```
oleganza: how about 

```rust
Program::build(|p| { p 
    .issue_helper()
    .output_helper()
}).to_vec()
```

decentralisedkev:
rustfmt seems to work with multiples of four.

Changing `|p|` to `prog` aligns the dots as expected

oleganza:
wow, true. I just noticed that with a chain of methods on `self.`

oleganza:
together with #129 we get this kind of code:

image

PR #115, opened 2019-02-22
spacesuit: deny missing docs

oleganza:
This enforces documentation for all exported items in spacesuit.

### DIFF
--- a/spacesuit/src/value.rs
+++ b/spacesuit/src/value.rs
@@ -6,14 +6,21 @@ use rand::{CryptoRng, Rng};
 use std::ops::{Add, Mul};
 use subtle::{Choice, ConditionallySelectable};
 
+/// A pair of a secret _quantity_ (64-bit integer)
+/// and a secret _flavor_ (scalar).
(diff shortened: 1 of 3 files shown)


spacesuit/src/value.rs
```diff
@@ -165,9 +177,13 @@ impl Neg for SignedInteger {
     }
 }
 
+/// Extension trait for committing Values to the Prover's constraint system.
+/// TBD: make this private by refactoring the benchmarks.
 pub trait ProverCommittable {
+    /// Type for a result of committing this type to a constraint system.
```
cathieyun: This is a confusing statement... is "this type" the `ProverCommittable` type? Maybe stating that explicitly would make this statement less weird

oleganza: how about this? 
https://github.com/interstellar/slingshot/pull/115/commits/6bb204b059dafc2f0cb4a2ab589c716f0711e69c

> Result of committing Self to a constraint system.

cathieyun: 👍

PR #118, opened 2019-02-25
spacesuit: add missing doc

vickiniu:
Adds missing doc to `pub fn to_scalar`

### DIFF
--- a/spacesuit/src/value.rs
+++ b/spacesuit/src/value.rs
@@ -120,6 +120,7 @@ impl SignedInteger {
         }
     }
 
+    /// Returns the Scalar representation of self.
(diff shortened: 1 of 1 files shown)


spacesuit/src/value.rs
```diff
@@ -120,6 +120,7 @@ impl SignedInteger {
         }
     }
 
+    /// Returns the Scalar representation of self.
```
oleganza: Pedant mode: Scalar is not strictly speaking a representation of integer. You can’t meaningfully convert back into u64, unless you knew that it never wrapped around after converting from u64.

How about?:

```suggestion
    /// Converts the integer to Scalar.
```

PR #119, opened 2019-02-25
zkvm: implement Mul for ScalarWitness

### LINKED ISSUE #91: zkvm: integer-preserving arithmetic on ScalarWitness

vickiniu:
Closes #91

### DIFF
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -13,7 +13,7 @@ use crate::ops::Instruction;
 use crate::predicate::Predicate;
 use crate::transcript::TranscriptProtocol;
 
-use std::ops::{Add, Neg};
+use std::ops::{Add, Mul, Neg};
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


vickiniu:
(travis failure is from missing doc, fixed in #118)

PR #120, opened 2019-02-25
zkvm/types: compute exact serialized length

### LINKED ISSUE #116: zkvm: precise length computation for Data types
1. Remove `min_serialized_length`
2. Add precise `serialized_len()->usize` to all data-like types in the hierarchy under `Data`.

vickiniu:
Computes exact serialized length for `Data` witness types, rather than providing a lower bound.

Fixes #116

### DIFF
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -358,7 +358,7 @@ impl DataWitness {
         match self {
             DataWitness::Program(instr) => instr.iter().map(|p| p.serialized_length()).sum(),
             DataWitness::Input(b) => 32 + b.contract.serialized_length(),
-            // All other types are encoded as 32-bit points or scalars
+            // All other types are encoded as 32-byte points or scalars
             _ => 32 as usize,
         }
     }
--- a/zkvm/src/types.rs
+++ b/zkvm/src/types.rs
@@ -100,6 +100,12 @@ pub enum ScalarWitness {
 }
 
 impl Commitment {
+    /// Returns the number of bytes needed to serialize the Commitment.
+    pub fn serialized_length(&self) -> usize {
+        32
+    }
+
+    /// Converts a Commitment to a compressed point.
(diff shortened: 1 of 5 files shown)


zkvm/src/types.rs
```diff
@@ -340,6 +339,15 @@ impl DataWitness {
             DataWitness::Input(b) => b.encode(buf),
         }
     }
+
+    fn serialized_len(&self) -> usize {
+        match self {
+            DataWitness::Program(instr) => instr.iter().map(|p| p.serialized_length()).sum(),
```
vickiniu: There's no way to avoid this recursion and still get the precise serialized length, but I figure it is alright since it shouldn't be possible to pass in an infinitely-recursing programs as input — not sure if there is a better guard against that!

oleganza: It’s totally fine here: prover creates the tree of types and it’s no big deal to traverse it here.

oleganza: If we use full “_length” elsewhere, why “_len” here? My mild personal preference is to use “_len” to match “len()” as used in collections.

zkvm/src/types.rs
```diff
@@ -354,6 +353,15 @@ impl DataWitness {
...
+            DataWitness::Input(b) => 32 + b.contract.serialized_length(),
+            // All other types are encoded as 32-bit points or scalars
```
```suggestion
            // All other types are encoded as 32-byte points or scalars
```

zkvm/src/types.rs
```diff
@@ -354,6 +353,15 @@ impl DataWitness {
...
+            _ => 32 as usize,
```
oleganza: I’d avoid “_” here and enumerate them all. So that when we add a new type we are not &@$&/:$SEGV.

oleganza: Ideally, i’d call serlen on each type and define them to return 32 within the scope of each type - so things are properly localized.

PR #130, opened 2019-02-25
zkvm: refactor Input/Contract serialization logic

### LINKED ISSUE #122: zkvm: de-emphasize witness APIs for more types
In https://github.com/interstellar/slingshot/pull/97 all of `Predicate` API is concentrated in the main `Predicate` type, while the `PredicateWitness` is left as a helper type. It must be public, but it does not have any public API and cannot be constructed other than via `Predicate`. This makes the code more straightforward as the user is exposed only to the type they actually care about while all the witness organization is left as an implementation detail.

We can do the same for other types, such as `Commitment`/`CommitmentWitness`. And maybe make `Frozen*` types `pub(crate)` and shrink down their API w.r.t. to usage within `VM`.

oleganza:
1. Renames `FrozenContract` to `Output`.
2. Makes other `Frozen*` types private.
3. Moves freezing/unfreezing logic into `contract.rs`.
4. Revisits private/pub(crate) methods.
5. Writing API is streamlined via `encoding::write_*` methods, so we can later remove assumption of `&mut Vec<u8>` and use generic `Writer` trait to make code cleaner and assuming less.

Closes #122.

### DIFF
--- a/zkvm/src/constraints.rs
+++ b/zkvm/src/constraints.rs
@@ -7,6 +7,7 @@ use curve25519_dalek::scalar::Scalar;
 use std::iter::FromIterator;
 use std::ops::{Add, Neg};
 
+use crate::encoding;
 use crate::scalar_witness::ScalarWitness;
 
 #[derive(Copy, Clone, Debug)]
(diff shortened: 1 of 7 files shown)


zkvm/src/constraints.rs
```diff
@@ -58,7 +59,7 @@ impl Commitment {
     }
 
     pub(crate) fn encode(&self, buf: &mut Vec<u8>) {
-        buf.extend_from_slice(&self.to_point().to_bytes());
+        encoding::write_point(&self.to_point(), buf);
```
oleganza: Note: later this and similar methods will become more like:

```rust
fn encode<W:Writer>(&self, writer: W) {
    writer.write_point(&self.to_point());
}
```

PR #134, opened 2019-02-26
zkvm: add an API guide

### LINKED ISSUE #96: zkvm: API guide

oleganza:
1. Added "ZkVM API" notes on Prover/Verifier API and witness types and rationale.
2. All docs moved to `docs/`
3. README emphasizes the links better.
4. Old spec location is preserved with a redirection link to avoid nasty 404s.

Rendered:

* [ZkVM spec](https://github.com/interstellar/slingshot/blob/oleg/zkvm-guide/zkvm/docs/zkvm-spec.md)
* [ZkVM API](https://github.com/interstellar/slingshot/blob/oleg/zkvm-guide/zkvm/docs/zkvm-api.md)
* [ZkVM spec old location](https://github.com/interstellar/slingshot/blob/oleg/zkvm-guide/zkvm/spec/ZkVM.md)

We can add more details to this document incrementally as we shape up ZkVM implementation.

Closes #96


zkvm/docs/zkvm-impl.md
```diff
@@ -0,0 +1,76 @@
...
+
+`Verifier` is an API for _verifying_ a transaction object `Tx`: it parses the bytecode, executes it, verifies the aggregated transaction signature and the R1CS proof, producing a `VerifiedTx` as a result. Verification logic is described by the [ZkVM specification](zkvm-spec.md).
+
+Note that `VM` execution is used twice: first, for _proving_, then for _verifying_. 
+
+How does the `Prover` know how to sign transaction and make a proof? The prover’s input is not an opaque sequence of instruction codes, but _witness-bearing instructions_. That is, a `push` instruction on the prover’s side does not hold an opaque string of bytes, but an accurate _witness type_ that may contain secret data and necessary structure for creating the proofs and signatures.
+
+## Witness types
+
+**Witness** is a type necessary to create a zero-knowledge proof or a signature.
```
vickiniu: maybe make explicit here that witness types contain secret information (that is blinded before the proof / sig / tx is composed for the verifier

zkvm/docs/zkvm-impl.md
```diff
@@ -0,0 +1,76 @@
...
+
+`Prover` is an API for _creating_ a transaction object `Tx` out of a stream of `Instruction`s:
+`VM` executes the instructions, producing a serialized program bytecode, aggregated transaction signature
+and a zero-knowledge R1CS proof for the [Cloak protocol](../../spacesuit/spec.md) and custom statements expressed by instructions.
+
+`Verifier` is an API for _verifying_ a transaction object `Tx`: it parses the bytecode, executes it, verifies the aggregated transaction signature and the R1CS proof, producing a `VerifiedTx` as a result. Verification logic is described by the [ZkVM specification](zkvm-spec.md).
+
+Note that `VM` execution is used twice: first, for _proving_, then for _verifying_. 
+
+How does the `Prover` know how to sign transaction and make a proof? The prover’s input is not an opaque sequence of instruction codes, but _witness-bearing instructions_. That is, a `push` instruction on the prover’s side does not hold an opaque string of bytes, but an accurate _witness type_ that may contain secret data and necessary structure for creating the proofs and signatures.
```
vickiniu: Might be useful to define our use of the terms `opaque` and `witness` (maybe a quick section on `Opaque types` before `Witness types`

zkvm/docs/zkvm-impl.md
```diff
@@ -0,0 +1,76 @@
...
+
+**Witness** is a type necessary to create a zero-knowledge proof or a signature.
+
+R1CS variables’ _witness_ is their secret _assignment_.
+
+VM instructions expect different types of witness data depending on the kind of object they operate on.
+
+### Commitments
+
+For the verifier, Pedersen commitments are opaque 32-byte [compressed points](zkvm-spec.md#point). For the verifier, these are represented by a `Data::Commitment` variant that holds a `Commitment` structure which can be `::Closed` or `::Open`.
```
vickiniu: This is a little confusing to me: can't the verifier only have `Commitment::Closed`?

zkvm/docs/zkvm-impl.md
```diff
@@ -0,0 +1,76 @@
...
+
+The `var` instruction in the prover’s VM expects an `::Open` commitment. An open commitment contains the secret value and its blinding factor explicitly, so they can be used to compute a R1CS proof.
+
+### Predicates
+
+`Predicate` type represents either an opaque predicate, or an entire tree where unused branches can be left opaque and visited branches are stored as-is. This structure can then produce correct input for the [`call`](zkvm-spec.md#call), [`left`](zkvm-spec.md#left) and [`right`](zkvm-spec.md#right) instructions.
```
vickiniu: link to spec here? for more details on the `Predicate` type

zkvm/docs/zkvm-impl.md
oleganza: Great point!

zkvm/docs/zkvm-impl.md
oleganza: ok!

zkvm/docs/zkvm-impl.md
oleganza: ypu!

zkvm/docs/zkvm-impl.md
oleganza: i've rewritten this, thanks

zkvm/README.md
```diff
@@ -2,6 +2,9 @@
 
 An evolution of TxVM with **cloaked assets** and **zero-knowledge smart contracts**.
 
+* [ZkVM specification](docs/zkvm-spec.md) — transaction validation rules.
+* [ZkVM implementation guide](docs/zkvm-impl.md) — transaction validation rules.
```
cathieyun: This should be something like "how to create transactions" instead of "transaction validation rules", I think

zkvm/docs/zkvm-impl.md
```diff
@@ -0,0 +1,75 @@
...
+### Expressions
+
+Expressions store their witness data as an optional assignment `Option<ScalarWitness>`.
+The verifier’s VM sees `None` and the prover’s VM sees `Some(...)`.
+
+See also [scalar witness](#scalar-witness).
+
+### Constraints
+
+Constraints do not have explicit witness data. It can be computed on the fly by evaluating the boolean function that the constraint represents, taking the underlying [expressions’](#expressions) assignments as input.
```
cathieyun: (nit): It -> They
(since "Constraints" is plural)

zkvm/docs/zkvm-impl.md
```diff
@@ -0,0 +1,75 @@
...
+The VM extracts the `Output` object from the `Input` and converts it to a [Contract type](zkvm-spec.md#contract-type) by allocating variables for each commitment within frozen values, turning them into actual [Value](zkvm-spec.md#value-type) types.
+
+### Scalar witness
+
+Scalar witness represents either:
+
+* a [scalar](zkvm-spec.md#scalar), or
+* a [signed integer](../../spacesuit/spec.md#signed-integer)
+
+Arithmetic operations on scalar witnesses _preserve integers until overflow_. If an addition/multiplication of two integers overflows the range of `±(2^64-1)`, the result is promoted to a scalar modulo Ristretto group order.
```
cathieyun: does this mean we have to do an overflow check after every arithmetic operation? 
Also this seems like strange behavior... we have to be able to operate over both integers and scalars, but only scalars after integers are overflowed... why not just have the scalar witnesses be represented as scalars all the time instead of doing this conversion? (Or, not allowing for/ throwing an error if an integer overflows? Though that seems like a functionality restriction, it would at least avoid this strange conversion step)

zkvm/README.md
oleganza: oops

zkvm/docs/zkvm-impl.md
oleganza: thx

zkvm/docs/zkvm-impl.md
oleganza: The rationale is:

1. Scalars are allowed to overflow because there's no definition of "overflow" for scalars :-)
2. If you integer arithmetic is well in bounds (normally it is), you are good.
3. If you accidentally overflow, rangeproof would fail because you've switched to the Scalar type. So you'll fail to even make a tx with ill-formed smart contract or bad input to it.
4. If you want to explicitly check whether you did overflow early, then you can check the result via `to_integer`, similar to how you check `Option<u64>` after `checked_add`.

cathieyun: 3. If you overflow, with the rangeproof _proving_ or _verifying_ will fail? If the _proving_ fails upon overflow, then that seems reasonable / safer than the other option.
4. I just question the rationale of allowing overflow (conversion to scalar upon overflow), since it seems like we'd be operating over unnecessarily large numbers at that point... which gives me hesitation for introducing extra complexity to support this edge case

oleganza: >3. If you overflow, will the rangeproof _proving_ or _verifying_ fail? If the _proving_ fails upon overflow, then that seems reasonable / safer than the other option.

Yes, proving.

oleganza: >4. I just question the rationale of allowing overflow (conversion to scalar upon overflow) at all, since it seems like we'd be operating over unnecessarily large numbers at that point... which gives me hesitation for introducing extra complexity to support this edge case

My thinking was that if we just happen to have a "small scalar" (0 or 1) that we want to use with scalars, we can just create ScalarWitness with this number, and it'll become automatically an integer. I felt that forcing it to be a Scalar from the start would make the interface heavier for the user. Also: if we force strict separation between scalars and integers, then we'd need to make Add/Neg/Mul failable (when subtypes cross each other), which may lead to a more ugly usage pattern. Hence, "upgrade int to scalar on overflow" seemed to be a good enough safety solution with none of the usability issues.

(Maybe i'm wrong, though)

cathieyun: Talked to @oleganza IRL, summary (for discussion point 4) here:
For operations over integers, we have to have some kind of failure mode for if the integers overflow. Either this can be throwing an error, in which case the functions have to be fallible, which is not ergonomics... or they can convert the result into a scalar, which is the proposal.

I think viewing the `integer -> overflow -> scalar` path as an _error handling_ solution, as opposed to a pathway that is _supposed to_ be used normally, makes a lot more sense to me. And I think that converting to a scalar is a much more user-friendly failure case than making the `add`, `mul` instructions themselves fallible. (though if we find that this is not true in the future, we can revisit this design decision)

oleganza: `add`/`mul` are still usefully "failable" in a sense that one can check what they got like they would check `Option<u64>` from `checked_add`. But their result type is usable in the subsequent operations, unlike Option or Result.

PR #136, opened 2019-02-26
slidechain: make block interval customizable

debnil:
Makes the expected interval between blocks upon submitting a transaction to TxVM customizable.

### DIFF
--- a/slidechain/cmd/slidechaind/slidechaind.go
+++ b/slidechain/cmd/slidechaind/slidechaind.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"time"
 
 	"github.com/chain/txvm/protocol"
 	"github.com/chain/txvm/protocol/bc"
@@ -23,9 +24,10 @@ func main() {
 	ctx := context.Background()
 
 	var (
-		addr   = flag.String("addr", "localhost:2423", "server listen address")
-		dbfile = flag.String("db", "slidechain.db", "path to db")
-		url    = flag.String("horizon", "https://horizon-testnet.stellar.org", "horizon server url")
+		addr          = flag.String("addr", "localhost:2423", "server listen address")
+		dbfile        = flag.String("db", "slidechain.db", "path to db")
+		url           = flag.String("horizon", "https://horizon-testnet.stellar.org", "horizon server url")
+		blockInterval = flag.Duration("interval", 5*time.Second, "expected interval between txvm blocks")
 	)
 
 	flag.Parse()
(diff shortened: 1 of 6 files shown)


slidechain/import.go
```diff
@@ -128,7 +128,7 @@ func (c *Custodian) doImport(ctx context.Context, nonceHash []byte, amount int64
 		return errors.Wrap(err, "computing transaction ID")
 	}
 	importTx.Runlimit = math.MaxInt64 - runlimit
-	_, err = c.S.submitTx(ctx, importTx)
+	_, err = c.S.submitTx(ctx, importTx, defaultBlockInterval)
```
vickiniu: Would we want to add this add a field on the `custodian` or the `submitter` so that we can configure it as a flag in `slidechaind`?

slidechain/submit.go
```diff
@@ -17,8 +17,8 @@ import (
 	"github.com/interstellar/slingshot/slidechain/net"
 )
 
-// TODO: make this configurable.
-var blockInterval = 5 * time.Second
+// DefaultBlockInterval is the default duration between expected blocks on TxVM.
+const DefaultBlockInterval = 5 * time.Second
```
vickiniu: Do we need to export this?

slidechain/cmd/slidechaind/slidechaind.go
```diff
@@ -39,6 +41,7 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
+	c.S.BlockInterval = *blockInterval
```
vickiniu: Slight preference for passing this block interval into `slidechain.GetCustodian` / `newCustodian` and then initializing it in the submitter there (rather than making the `BlockInterval` field public)

slidechain/submit.go
vickiniu: also, if we pass `blockInterval` as a parameter into `newCustodian` we could probably get rid of this constant (& just define a block interval in tests)

slidechain/submit.go
```diff
@@ -124,6 +126,16 @@ func (s *submitter) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 	wait := (waitStr != "")
 
+	blockIntervalStr := req.FormValue("blockInterval")
```
vickiniu: Do we want anyone who submits a transaction to be able to set the block interval? Seems potentially dangerous

slidechain/submit.go
debnil: Discussed offline, but for posterity -- the current approach helps ensure the same default block interval in `slidechaind` and `slidechain_test` and makes future changes slightly easier, so we'll keep it over hand-defining durations in different parts of the codebase.

Issue #95, opened 2019-02-19
token: abstraction for issuing and retiring tokens


oleganza:
This is related issue to **Accounts API** #94:

Features:

1. Create a token
2. Issue a token
3. Compute amount of outstanding tokens in circulation
4. Receive and retire tokens
5. Embed and verify metadata into the token's flavor.

See #94 for more considerations w.r.t. statefulness and FFI.

oleganza:
A few thoughts after reading through #139:

### API focus

Maybe the proper name for this API is not "Treasury", but "Token"?

1. You can define token with a given predicate.
2. You can issue token as a one-shot tx (issue-and-send-to), or as a partial program in a larger transaction.
3. Embed/verify metadata in a token.
4. Issue/retire protocol loop with abstracted storage.

### Flexibility

If we let the user provide the predicate, we can separate the concerns: they would be able to construct a multi-sig/multi-party configurations w/o the API making these decisions for them.

Then, we can have separate tooling for fancy predicates that could be used with both issuers and accounts.

### Program separation

Issuance is probably going to be one of the actions in a transaction. Even if we start with a "one shot" function that "issues to an address", it's still better to separate two-three layers, at least internally:

1. Issuing a value to a stack.
2. Transferring the value to a destination.
3. Composing the whole transaction (and adding `nonce` if there are no inputs).

### Signing separation

For all operations with tokens but signing an issuance, we should not require `PredicateWitness::Key`. So we need to allow creating a `Token` type with a more opaque predicate (maybe still a witness tree, but w/o signing keys, if we need to know the structure), but when composing a program for signing, be able to provide the keys separately. Maybe we'll need to redesign `PredicateWitness::Key` entirely to use an indirect reference to signing material — that may be needed to accommodate both in-HSM signing and multi-party signing.

### Circulation

Circulation seems to be out of scope, but the accurate way to compute it is by processing proved-to-be-published transactions, not just incrementing a counter in a table we one signed off. 

So we'll need a generic trait for the "state" part of the following "state machine":

1. When issuing: store the `qty_commitment -> (qty,blinding)` pair for later retrieval.
2. When retiring: store the `qty_commitment -> (qty,blinding)` and `flv_commitment -> (flv,blinding)` pairs for later retrieval.
[...]

PR #139, opened 2019-02-26
zkvm/treasury: add treasury to issue, retire, track tokens

vickiniu:
See #95


zkvm/src/lib.rs
```diff
@@ -12,6 +12,7 @@ mod prover;
 mod scalar_witness;
 mod signature;
 mod transcript;
+mod treasury;
```
oleganza: Maybe it makes sense to make treasury/accounts a separate _crate_ (`slingshot/treasury` instead of `slingshot/zkvm/src/treasury`)? This way `zkvm` crate will be solely focused on executing the VM and having a single layer API  not cluttered with any higher-level (opinionated) protocols.

vickiniu: ah, yes! sorry I meant to write that out but one of my questions was how best to structure this and a separate crate makes a lot of sense to me 👍 !

oleganza: That would also help us define what's the ZkVM should and should not be, that would allow us to simplify the internals too — if we know exactly what we need to expose to accounts, treasury, blockchain, then the rest can be freely reorganized and simplified.

zkvm/src/treasury.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    pub fn retire_token(flavor: Predicate, qty: u64) -> Result<(Tx, TxID, TxLog), VMError> {
+        unimplemented!()
+    }
+}
+
+impl Program {
+    pub fn issue_token(&mut self, qty: u64, flavor: Predicate, dest: Scalar) -> &mut Self {
+        let nonce = Predicate::from_signing_key(Scalar::random(&mut rand::thread_rng()));
+        self.push(Commitment::blinded_with_factor(qty, Scalar::random(&mut rand::thread_rng())))
```
oleganza: if you are going to use a random blinding factor, `Commitment::blinded()` does it already. `_with_factor` variant is provided for use with a pre-computed blinding factor.

oleganza:
@vickiniu a few thoughts on the token API: 
https://github.com/interstellar/slingshot/issues/95#issuecomment-467588251

vickiniu:
moved to token in #161

PR #140, opened 2019-02-26
slidechain: streamline peg-in RPC

debnil:
Streamlines peg-in RPC and surrounding logic. Previously, the `record` RPC only recorded the peg-in transaction in the `pegs` table. As a result, it required a prior server round-trip to submit and wait on the pre-peg-in TxVM transaction. Now, the RPC (renamed to `prepegin`) also builds and submits the pre-peg-in transaction. All pre-peg-in related logic is moved into a single file.


slidechain/prepeg.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		net.Errorf(w, http.StatusInternalServerError, "sending response: %s", err)
+		return
+	}
+	err = c.S.waitOnTx(ctx, tx.ID, r)
+	if err != nil {
+		net.Errorf(w, http.StatusInternalServerError, "sending response: %s", err)
+	}
+	// Record peg in database.
+	nonceHash := UniqueNonceHash(c.InitBlockHash.Bytes(), p.ExpMS)
```
bobg: If you return this value to the `/prepegin` RPC caller, then you don't have to export `UniqueNonceHash`.

slidechain/prepeg.go
```diff
@@ -0,0 +1,106 @@
...
+type PrePegIn struct {
+	BcID        []byte `json:"bc_id"`
+	Amount      int64  `json:"amount"`
+	AssetXDR    []byte `json:"asset_xdr"`
+	RecipPubkey []byte `json:"recip_pubkey"`
+	ExpMS       int64  `json:"exp_ms"`
+}
+
+// BuildPrePegInTx builds the pre-peg-in TxVM transaction to create a uniqueness token.
+func BuildPrePegInTx(bcid, assetXDR, recip []byte, amount, expMS int64) (*bc.Tx, error) {
```
bobg: This no longer needs to be exported.

slidechain/cmd/peg/peg.go
```diff
@@ -127,59 +128,32 @@ func main() {
...
-		return errors.Wrap(err, "recording to slidechaind")
+		return nil, errors.Wrap(err, "recording to slidechaind")
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode/100 != 2 {
-		return fmt.Errorf("status code %d from POST /record", resp.StatusCode)
+		return nil, fmt.Errorf("status code %d from POST /prepegin", resp.StatusCode)
+	}
+
+	nonceHashBytes, err := ioutil.ReadAll(resp.Body)
```
bobg: If instead of this you do:

```
var nonceHash [32]byte
_, err := io.ReadFull(resp.Body, nonceHash[:])
```

and if you then return `nonceHash`, you can simultaneously check that enough bytes were available and eliminate the `copy` in the caller.

slidechain/cmd/peg/peg.go
```diff
@@ -130,7 +128,8 @@ func main() {
 
 // doPrePegIn calls the pre-peg-in Slidechain RPC.
 // That RPC builds, submits, and waits for the pre-peg TxVM transaction and records the peg-in in the database.
-func doPrePegIn(bcid, assetXDR []byte, amount, expMS int64, pubkey ed25519.PublicKey, slidechaind string) ([]byte, error) {
+func doPrePegIn(bcid, assetXDR []byte, amount, expMS int64, pubkey ed25519.PublicKey, slidechaind string) ([32]byte, error) {
+	var zeroBytes [32]byte
```
bobg: This is awkward. A better solution is to declare `nonceHash` here and just return it wherever you're using `zeroBytes`.

PR #141, opened 2019-02-26
slidechain: fix amount discrepancy bug in peg cmd

debnil:
Fixes an amount discrepancy bug in `cmd/peg`. Lumen amounts were being passed as stroops to the pre-peg-in TxVM transaction, but as lumens to the import transaction. This discrepancy manifested itself as an `invalid prevout` error when submitting the import transaction, because the contract stack of the reconstructed pre-peg-in snapshot differed from that of the actual transaction.


slidechain/cmd/peg/peg.go
```diff
@@ -37,7 +38,7 @@ func main() {
 	)
 	flag.Parse()
 
-	if *amount == "" {
+	if *amount == -1 {
```
bobg: Suggest `if *amount <= 0` (but see note above).

slidechain/cmd/peg/peg.go
```diff
@@ -21,12 +21,13 @@ import (
...
 	"github.com/interstellar/slingshot/slidechain"
 	"github.com/interstellar/slingshot/slidechain/stellar"
+	"github.com/interstellar/starlight/worizon/xlm"
 )
 
 func main() {
 	var (
 		custodian   = flag.String("custodian", "", "Stellar account ID of custodian account")
-		amount      = flag.String("amount", "", "amount to peg, in lumens")
+		amount      = flag.Int64("amount", -1, "amount to peg, in lumens")
```
bobg: Can this not be fractional? And if it can't, should it be in stroops?

IMO the best option is that this remains a string, but one parseable by `xlm.Parse` (from `worizon`).

slidechain/cmd/peg/peg.go
```diff
@@ -99,16 +95,18 @@ func main() {
 		}
 	}
 
-	assetXDR, err := asset.MarshalBinary()
+	amountXLM, err := xlm.Parse(*amount)
```
bobg: I believe this only applies in the case of lumens, meaning this belongs in the `else` clause just above. For other asset types you need to do different parsing. I think it's sufficient in those cases to do `strconv.ParseInt`.

debnil: Discussed offline - for posterity, from https://www.stellar.org/developers/guides/concepts/assets.html#amount-precision-and-representation it appears that all assets are scaled down by a factor of ten million.

bobg: So 10000000 USD is one dollar. What could go wrong? 🙄

PR #142, opened 2019-02-26
zkvm/predicate: remove secret signing information from predicate

vickiniu:
Stores only verification key information in `PredicateWitness::Key`, with the prover providing the signature w/ secret signing keys in tx-building.

### DIFF
--- a/zkvm/src/predicate.rs
+++ b/zkvm/src/predicate.rs
@@ -4,7 +4,7 @@
 //! - disjunction: P = L + f(L,R)*B
 //! - program_commitment: P = h(prog)*B2
 use bulletproofs::PedersenGens;
-use curve25519_dalek::ristretto::{CompressedRistretto, RistrettoPoint};
+use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
 use merlin::Transcript;
 
@@ -30,13 +30,13 @@ pub enum Predicate {
     Program(Vec<Instruction>),
 
     /// Disjunction of two predicates.
-    Or(PredicateDisjunction),
+    Or(Box<PredicateDisjunction>),
 }
 
 #[derive(Clone, Debug)]
 pub struct PredicateDisjunction {
-    left: Box<Predicate>,
-    right: Box<Predicate>,
+    left: Predicate,
+    right: Predicate,
     // We precompute the disjunction predicate when composing it via `Predicate::or()`,
     // so that we can keep `to_point`/`encode` methods non-failable across all types.
     precomputed_point: CompressedRistretto,
@@ -53,7 +53,7 @@ impl Predicate {
         match self {
             Predicate::Opaque(p) => *p,
             Predicate::Key(k) => k.0,
-            Predicate::Or(d) => d.point,
+            Predicate::Or(d) => d.precomputed_point,
             Predicate::Program(prog) => {
                 let mut bytecode = Vec::new();
                 Instruction::encode_program(prog.iter(), &mut bytecode);
@@ -105,7 +105,16 @@ impl Predicate {
 
     /// Creates a disjunction of two predicates.
     pub fn or(self, right: Predicate) -> Result<Self, VMError> {
-        Ok(Predicate::Or(PredicateDisjunction::new(self, right)?))
+        let point = {
+            let l = self.to_point();
+            let f = Predicate::commit_or(l, right.to_point());
+            l.decompress().ok_or(VMError::InvalidPoint)? + f * PedersenGens::default().B
+        };
+        Ok(Predicate::Or(Box::new(PredicateDisjunction {
+            left: self,
+            right: right,
+            precomputed_point: point.compress(),
+        })))
     }
 
     /// Creates a program-based predicate.
@@ -130,22 +139,6 @@ impl Predicate {
 impl Into<CompressedRistretto> for Predicate {
     fn into(self) -> CompressedRistretto {
         self.to_point()
-    }
-}
-
-impl PredicateDisjunction {
-    pub fn new(left: Predicate, right: Predicate) -> Result<Self, VMError> {
-        let point = {
-            let l = left.to_point();
-            let f = Predicate::commit_or(l, right.to_point());
-            l.decompress().ok_or(VMError::InvalidPoint)? + f * PedersenGens::default().B
-        };
-
-        Ok(PredicateDisjunction {
-            left: Box::new(left),
-            right: Box::new(right),
-            point: point.compress(),
-        })
     }
 }
 
--- a/zkvm/src/predicate.rs
+++ b/zkvm/src/predicate.rs
@@ -108,7 +108,8 @@ impl Predicate {
         let point = {
             let l = self.to_point();
             let f = Predicate::commit_or(l, right.to_point());
-            l.decompress().ok_or(VMError::InvalidPoint)? + f * PedersenGens::default().B
+            let l = l.decompress().ok_or(VMError::InvalidPoint)?;
+            l + f * PedersenGens::default().B
         };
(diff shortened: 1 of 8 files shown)


zkvm/src/predicate.rs
```diff
@@ -32,8 +32,10 @@ pub enum PredicateWitness {
     /// (but known to be a valid Ristretto point).
     OpaqueBranch(RistrettoPoint),
 
-    /// Secret signing key for the predicate-as-a-verification-key.
-    Key(Scalar),
+    /// Signing key for the predicate-as-a-verification-key.
+    /// Prover will provide secret signing key separately when
+    /// constructing aggregated signature.
+    Key(RistrettoPoint),
```
oleganza: why not `Key(VerificationKey)`?

vickiniu: We could do `Key(VerificationKey)` with a bigger refactor of `Predicate`, getting rid of the `PredicateWitness` type (which I think might be a good idea!), but right now we need to keep the key as uncompressed type to keep `to_uncompressed_point` unfailable

vickiniu: ah, I spoke too soon! I think we might have to keep `Key(RistrettoPoint)` since we need the predicates to be uncompressed form for `to_point` to work correctly for `Predicate::Or` (pushed an updated commit)

zkvm/src/predicate.rs
```diff
@@ -52,7 +62,20 @@ impl Predicate {
...
         match self {
             Predicate::Opaque(p) => *p,
-            Predicate::Witness(w) => w.to_uncompressed_point().compress(),
+            Predicate::Key(k) => k.0,
+            Predicate::Or(l, r) => {
+                let l = l.to_point();
+                let r = r.to_point();
+                let f = Predicate::commit_or(l, r);
+                // l.decompress() + f * PedersenGens::default().B
+                unimplemented!()
```
oleganza: the need to decompress the point and desire to compute to_point w/o failing is what prompted creating `OpaqueBranch(RistrettoPoint)` originally.

oleganza: The failable conversion to point is bad because then `encode` must be failable and it poisons everything.

oleganza: Maybe try `Opaque(RistrettoPoint)` and fail on creation via helper method `Predicate::opaque(CompressedRistretto) -> Result<P, ...>`?

oleganza: But then it won't play nice with current design of PointOp that works with compressed points, so we'd need to redesign it too.

vickiniu: Took another pass at still maintaining `RistrettoPoint` internally (but compressing in `to_point`, etc) !

zkvm/src/predicate.rs
```diff
@@ -21,25 +21,23 @@ pub enum Predicate {
...
+    /// Prover will provide secret signing key separately when
+    /// constructing aggregated signature.
+    Key(VerificationKey),
 
     /// Representation of a predicate as commitment to a program.
     Program(Vec<Instruction>),
 
     /// Disjunction of two predicates.
-    Or(Box<PredicateWitness>, Box<PredicateWitness>),
+    Or(PredicateDisjunction),
```
oleganza: Maybe box things higher in the hierarchy, so that fewer types have indirect storage?

```rust
pub enum Predicate {
    ...
    Or(Box<PredicateDisjunction>)
}

pub struct PredicateDisjunction {
     left: Predicate,
     right: Predicate,
     point: CompressedRistretto,
}
```

WDYT?

zkvm/src/predicate.rs
```diff
@@ -21,25 +21,23 @@ pub enum Predicate {
...
+}
+
+#[derive(Clone, Debug)]
+pub struct PredicateDisjunction {
+    left: Box<Predicate>,
+    right: Box<Predicate>,
+    point: CompressedRistretto,
```
oleganza: clarify that this field is redundant, and comment why it is needed:
```suggestion
    // We precompute the disjunction predicate when composing it via `Predicate::or()`,
    // so that we can keep `to_point`/`encode` methods non-failable across all types.
    precomputed_point: CompressedRistretto,
```

zkvm/src/predicate.rs
```diff
@@ -144,36 +125,25 @@ impl Predicate {
...
     }
 }
 
-impl From<CompressedRistretto> for Predicate {
-    fn from(p: CompressedRistretto) -> Self {
-        Predicate::Opaque(p)
-    }
-}
+impl PredicateDisjunction {
+    pub fn new(left: Predicate, right: Predicate) -> Result<Self, VMError> {
```
oleganza: We probably don't need to use `PredicateDisjunction::new` in other contexts, so how about:

1. Removing `PredicateDisjunction::new` method.
2. Doing its job directly in the `Predicate::or` method.

zkvm/src/prover.rs
```diff
@@ -60,11 +60,15 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
 }
 
 impl<'a, 'b> Prover<'a, 'b> {
-    pub fn build_tx<'g>(
+    pub fn build_tx<'g, F>(
         program: Vec<Instruction>,
         header: TxHeader,
         bp_gens: &'g BulletproofGens,
-    ) -> Result<(Tx, TxID, TxLog), VMError> {
+        sign_tx_fn: F,
```
oleganza: Note for later: we'll most likely have a whole trait/interface/type called `Signer` or something, and it'll support  multiple keys, multiple parties and all of the above at various levels. Cf. https://github.com/interstellar/slingshot/issues/146

oleganza: With such interface it'd make sense to compose all these pubkeys into a N-of-N "witness" object (yes, on the verifier's side) which would be the same witness object that provers would use for their custom N-of-N multisig rules packed in a single key (not visible to the verifier).

zkvm/src/predicate.rs
vickiniu: ah, that's a great suggestion! thanks :)

zkvm/src/prover.rs
vickiniu: sounds good, will make a note!

zkvm/src/predicate.rs
```diff
@@ -88,46 +95,31 @@ impl Predicate {
...
     pub fn or(self, right: Predicate) -> Result<Self, VMError> {
-        let l = self.to_witness()?;
-        let r = right.to_witness()?;
-        Ok(Predicate::Witness(
-            PredicateWitness::Or(Box::new(l), Box::new(r)).into(),
-        ))
+        let point = {
+            let l = self.to_point();
+            let f = Predicate::commit_or(l, right.to_point());
+            l.decompress().ok_or(VMError::InvalidPoint)? + f * PedersenGens::default().B
```
oleganza: maybe make this line a bit clearer?

```suggestion
            let l = l.decompress().ok_or(VMError::InvalidPoint)?;
            l + f * PedersenGens::default().B
```

zkvm/tests/zkvm.rs
```diff
@@ -239,7 +271,7 @@ fn spend_1_2() {
         predicates[2].clone(), // output 2 predicate
     );
 
-    match test_helper(correct_program) {
+    match test_helper(correct_program, &scalars) {
```
oleganza: I don't feel good about passing all these scalars around — there should be a cleaner way. But maybe we'll have a better way closer to the "signer API" a bit later, so let's just keep it explicit for now.

vickiniu: for sure — it's not a very clean solution to providing secret signing information indirectly, but refactoring this with the signer API makes sense to me!

PR #144, opened 2019-02-27
slidechain: streamline parsing in export and peg cmds

debnil:
Streamlines amount and asset parsing in `cmd/export` and `cmd/peg`.

### DIFF
--- a/slidechain/cmd/export/export.go
+++ b/slidechain/cmd/export/export.go
@@ -62,16 +62,14 @@ func main() {
 
 	// XDR scales down an amount unit of every asset by a factor of 10^7.
 	// Thus, xlm.Parse works for both native and non-native assets.
-	exportXlm, err := xlm.Parse(*amount)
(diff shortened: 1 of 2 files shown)


slidechain/cmd/export/export.go
```diff
@@ -51,38 +50,29 @@ func main() {
...
-		inputXlm, err := xlm.Parse(*input)
-		if err != nil {
-			log.Fatalf("error parsing input amount %s: %s", *input, err)
-		}
-		inputAmount = int64(inputXlm)
 	}
+
+	// XDR scales down an amount unit of every asset by a factor of 10^7.
+	// Thus, xlm.Parse works for both native and non-native assets.
+	exportXlm, err := xlm.Parse(*amount)
```
bobg: Suggest calling this `exportAmount` (since it's not XLM-specific) and, instead of creating a new variable to hold the `int64`-typed version of this value, just type-convert inline at the point you use it.

Same thing for `inputXlm` below.

PR #147, opened 2019-02-27
zkvm: implement `and`, `or`, `verify` instructions

cathieyun:
- change `Constraint::{And, Or}` to be a tuple of two boxed constraints, instead of a vector (since they'll only ever be created from two constraints anyways, and we can do the flattening in the `verify` function
- implement `and`, `or` using the new Constraint type
- implement `verify`: enter 2nd phase of 2-phase challenge API, in order to get `z` challenge. Then recursively "flatten" the constraint, until the base case `Expression` type, and combine the `Expression`s according to the rules until you have one `Expression` that represents the whole constraint. Then push that constraint onto the constraint system.

Question: I implemented the `From<R1CSError> for VMError` trait, but in `verify` it does not automatically convert `R1CSError` to `VMError`, so I have to do it explicitly. I thought this would be implicit because of the `From` trait?
Update: What I'd want is `TryFrom`, which is nightly-only. So I'll just do explicit conversion for now!

### DIFF
--- a/zkvm/src/constraints.rs
+++ b/zkvm/src/constraints.rs
@@ -1,7 +1,7 @@
 //! Constraint system-related types and operations:
 //! Commitments, Variables, Expressions and Constraints.
 
-use bulletproofs::{r1cs, r1cs::RandomizedConstraintSystem, PedersenGens};
+use bulletproofs::{r1cs, r1cs::ConstraintSystem, r1cs::RandomizedConstraintSystem, PedersenGens};
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
 use std::iter::FromIterator;
@@ -9,7 +9,6 @@ use std::ops::{Add, Neg, Sub};
 
 use crate::encoding;
 use crate::scalar_witness::ScalarWitness;
-use crate::transcript::TranscriptProtocol;
 
 #[derive(Copy, Clone, Debug)]
 pub struct Variable {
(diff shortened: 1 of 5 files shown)


zkvm/src/constraints.rs
```diff
@@ -44,6 +45,42 @@ pub struct CommitmentWitness {
...
+
+            // Add the resulting expression to the constraint system
+            cs.constrain(expr.to_r1cs_lc());
+
+            Ok(())
+        })
+        .map_err(|e| VMError::R1CSError(e))
+    }
+
+    fn flatten<CS: r1cs::RandomizedConstraintSystem>(self, cs: &mut CS, z: Scalar) -> Expression {
```
oleganza: To avoid confusion, i'd flatten into the r1cs linearcombination type and used `zkvm::Expression` only for the operations that are defined in the spec.

zkvm/src/constraints.rs
```diff
@@ -209,6 +256,14 @@ impl Add for Expression {
     }
 }
 
+impl Sub for Expression {
```
oleganza: still don't need Sub :-P

zkvm/src/constraints.rs
```diff
@@ -44,6 +45,42 @@ pub struct CommitmentWitness {
     blinding: Scalar,
 }
 
+impl Constraint {
+    pub fn verify<CS: r1cs::ConstraintSystem>(self, cs: &mut CS) -> Result<(), VMError> {
+        cs.specify_randomized_constraints(move |cs| {
+            let z = cs.challenge_scalar(b"verify challenge");
+
+            // Flatten the constraint into one expression
+            // Note: cloning because we can't move out of captured variable in an `Fn` closure
```
oleganza: And Box<FnOnce> is not supported yet. Cf. https://github.com/dalek-cryptography/bulletproofs/issues/244

(maybe add this reference to this comment)

zkvm/src/constraints.rs
```diff
@@ -44,6 +45,42 @@ pub struct CommitmentWitness {
...
+            Ok(())
+        })
+        .map_err(|e| VMError::R1CSError(e))
+    }
+
+    fn flatten<CS: r1cs::RandomizedConstraintSystem>(self, cs: &mut CS, z: Scalar) -> Expression {
+        match self {
+            Constraint::Eq(expr1, expr2) => expr1 - expr2,
+            Constraint::And(c1, c2) => {
+                let a = c1.flatten(cs, z);
```
oleganza: i'm not sure you can use the same challenge across all levels of a boolean function. IMHO it's safer to generate it per each `and(a,b)` pair.

zkvm/src/constraints.rs
cathieyun: 👍 Added comment!

zkvm/src/constraints.rs
cathieyun: ah that's true, fixed!

zkvm/src/constraints.rs
```diff
@@ -44,6 +45,43 @@ pub struct CommitmentWitness {
...
+
+    fn flatten<CS: r1cs::RandomizedConstraintSystem>(self, cs: &mut CS) -> Expression {
+        match self {
+            Constraint::Eq(expr1, expr2) => expr1 - expr2,
+            Constraint::And(c1, c2) => {
+                let a = c1.flatten(cs);
+                let b = c2.flatten(cs);
+                // output expression: a + z * b
+                let z = Expression::constant(cs.challenge_scalar(b"verify challenge"));
+                a + b.multiply(z, cs)
```
oleganza: If these are r1cs::LinearCombinations, we can simply do 

```rust
let z = cs.challenge_scalar(b"ZkVM.verify.and-challenge");
a + b.multiply(z, cs)
```

BTW, we probably need to update the spec with the label `"ZkVM.verify.and-challenge"` because it affects the resulting proof.

zkvm/src/constraints.rs
```diff
@@ -44,6 +45,43 @@ pub struct CommitmentWitness {
...
+            }
+            Constraint::Or(c1, c2) => {
+                let a = c1.flatten(cs);
+                let b = c2.flatten(cs);
+                // output expression: a * b
+                a.multiply(b, cs)
```
oleganza: If these are r1cs::LinearCombinations, we can instead do

```rust
cs.multiply(a,b)
```

zkvm/src/constraints.rs
cathieyun: 👍

PR #148, opened 2019-02-27
slingshot: check all the markdown references

oleganza:
This adds a script that checks that most references in the specs are valid:

1. Checks all the local anchors within each `.md` file.
2. Checks all cross-file references and anchors between `.md` files.
3. Checks if remote URLs are loadable (pings `http(s)://` links).
4. Travis checks these on each PR.
5. `/vendor/` paths are ignored.

### Limitations

1. The title-to-anchor convertor is ad-hoc and may not produce the same result as Github/Markdown, so this may turn into false negatives or false positives.
2. Links to anchors to remote pages are not fully checked: only the URL is pinged, but the anchor within it is not verified (so it may become stale).
3. Remote URLs may die causing your PRs be rejected by Travis. In such case it's a good opportunity to fix the URLs or revisit the strictness of the script.

### DIFF
--- a/check-markdown-links.rb
+++ b/check-markdown-links.rb
@@ -2,7 +2,7 @@
 require 'open-uri'
 require 'pathname'
 require 'pp'
-CHECK_HTTP_LINKS = !!ENV["CHECK_HTTP_LINKS"]
+DO_NOT_CHECK_HTTP_LINKS = !!ENV["DO_NOT_CHECK_HTTP_LINKS"]
 
 # 1. Get all links
 # 2. For each link, check if it's good:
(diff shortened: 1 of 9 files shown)


oleganza:
Example:

image

cathieyun:
This seems good! Do you know how much it time it adds to Travis's runtime? Just curious...

check-markdown-links.rb
```diff
@@ -0,0 +1,144 @@
...
+          $check_links_failed = true
+        end
+      end
+    end
+  end
+end
+
+def check_url(url, cache = {})
+  return true if cache[url]
+  if url == "https://dx.doi.org/10.6028/NIST.FIPS.202"
```
cathieyun: what's the reason to special-case this link?

oleganza: sometimes it does not work if fetched from a script instead of a browser

PR #161, opened 2019-02-28
token: add basic Token API

### LINKED ISSUE #95: token: abstraction for issuing and retiring tokens

vickiniu:
Adds simple API for issuing new tokens

Fixes #95


token/src/token.rs
```diff
@@ -0,0 +1,89 @@
+use bulletproofs::{BulletproofGens, PedersenGens};
+use curve25519_dalek::scalar::Scalar;
+use zkvm::*;
```
oleganza: let's be more explicit about what we import? The rationale is that we are still not sure what zkvm API looks like, so exact usage in `./token`, `./blockchain` etc would inform us. The `use zkvm::*` does not tell us which types are actually used.

token/src/lib.rs
```diff
@@ -0,0 +1 @@
+mod token;
```
oleganza: Suggestion: add `#![deny(missing_docs)]` up-front.

token/src/token.rs
```diff
@@ -0,0 +1,89 @@
...
+    flavor: Predicate,
+    metadata: Vec<u8>,
+    nonce: Scalar,
+}
+
+impl Token {
+    pub fn new(pred: Predicate, metadata: &[u8]) -> Self {
+        Token {
+            flavor: pred,
+            metadata: metadata.to_vec(),
```
oleganza: If you need a `Vec<u8>`, you may simply ask for it right here and move it in w/o extra allocations.

https://rust-lang-nursery.github.io/api-guidelines/flexibility.html

image

token/src/token.rs
```diff
@@ -0,0 +1,89 @@
+use bulletproofs::{BulletproofGens, PedersenGens};
+use curve25519_dalek::scalar::Scalar;
+use zkvm::*;
+
+pub struct Token {
+    flavor: Predicate,
+    metadata: Vec<u8>,
+    nonce: Scalar,
```
oleganza: `nonce` is a per-transaction thing, and is even optional. Ideally the Token API does not even have to worry about it unless the token's issuance predicate tries to build a singleton or issue-once logic. It's the "transaction builder"'s business to add `nonce` if necessary to a final tx — if there are some `input`s added, `nonce` is not necessary.

token/src/token.rs
```diff
@@ -0,0 +1,89 @@
...
+            .var() // stack: qty-var
+            .push(Commitment::unblinded(Value::issue_flavor(
+                &self.flavor,
+                Data::Opaque(self.metadata.to_vec()),
+            ))) // stack: qty-var, flv
+            .var() // stack: qty-var, flv-var
+            .push(Data::Opaque(self.metadata.to_vec())) // stack: qty-var, flv-var, data
+            .push(self.flavor.clone()) // stack: qty-var, flv-var, data, flv-pred
+            .issue() // stack: issue-contract
+            .push(nonce) // stack: issue-contract, nonce-pred
```
oleganza: Per comment above, let's push the nonce-handling to the caller. If we have some `txbuilder` instance later, it will keep track of inputs and nonces, and add a `nonce` if necessary.

token/src/token.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    pub fn issue<'a>(&self, program: &'a mut Program, qty: u64) -> &'a mut Program {
+        let nonce = Predicate::Key(VerificationKey::from_secret(&self.nonce));
+        program
+            .push(Commitment::blinded(qty)) // stack: qty
+            .var() // stack: qty-var
+            .push(Commitment::unblinded(Value::issue_flavor(
+                &self.flavor,
+                Data::Opaque(self.metadata.to_vec()),
```
oleganza: oh, i overlooked. I don't think we need to ask for `Data` in `issue_flavor`. We only need a `Vec<u8>`.

token/src/token.rs
```diff
@@ -0,0 +1,89 @@
+use bulletproofs::{BulletproofGens, PedersenGens};
+use curve25519_dalek::scalar::Scalar;
+use zkvm::*;
+
+pub struct Token {
+    flavor: Predicate,
```
oleganza:
```suggestion
    issuance_predicate: Predicate,
```

token/src/token.rs
```diff
@@ -0,0 +1,89 @@
...
+            .push(Data::Opaque(self.metadata.to_vec())) // stack: qty-var, flv-var, data
+            .push(self.flavor.clone()) // stack: qty-var, flv-var, data, flv-pred
+            .issue() // stack: issue-contract
+            .push(nonce) // stack: issue-contract, nonce-pred
+            .nonce() // stack: issue-contract, nonce-contract
+            .sign_tx() // stack: issue-contract
+            .sign_tx() // stack: issued-value
+    }
+
+    pub fn issue_and_spend<'a>(
```
oleganza: bikeshedding: maybe call this `issue_to`?

token/src/token.rs
```diff
@@ -0,0 +1,89 @@
...
+        &self,
+        program: &'a mut Program,
+        qty: u64,
+        dest: Predicate,
+    ) -> &'a mut Program {
+        self.issue(program, qty).push(dest).output(1)
+    }
+
+    pub fn build(
```
oleganza: maybe keep it as a helper method in the tests - it's not likely that the `Token` API would take care of an entire transaction - that may be handled in a "tx builder" API or simply in a concrete application that assembles transactions from various pieces.

token/src/token.rs
vickiniu: made the imports more explicit! I think it would be really great to break `zkvm` into some submodules to make the imports more wieldy (but maybe we can try to define those based on usage :) )

token/src/token.rs
```diff
@@ -0,0 +1,190 @@
...
+            metadata: metadata,
+        }
+    }
+
+    /// Returns program that issues specified quantity of Token.
+    pub fn issue<'a>(program: &'a mut Program, token: &Token, qty: u64) -> &'a mut Program {
+        program
+            .push(Commitment::blinded(qty)) // stack: qty
+            .var() // stack: qty-var
+            .push(Commitment::unblinded(Value::issue_flavor(
```
oleganza: How about factoring flavor computation in a `pub fn flavor(&self) -> Scalar` so that the user of `Token` can always ask for its flavor?

token/src/token.rs
```diff
@@ -0,0 +1,190 @@
...
+            );
+            let dest = Predicate::Key(VerificationKey::from_secret(&dest_key));
+            let program = Program::build(|p| {
+                Token::issue_to(p, &usd, 10u64, &dest)
+                    .push(Predicate::Key(VerificationKey::from_secret(&nonce_key)))
+                    .nonce()
+                    .sign_tx()
+            })
+            .to_vec();
+            match build(program, &usd, vec![issue_key, nonce_key]) {
```
oleganza: you can just `build().unwrap()` - it will return a good transaction or panic in place just like `assert!`. So no need for `match` statement.

token/src/token.rs
```diff
@@ -0,0 +1,190 @@
...
+                Ok(x) => x,
+                Err(err) => return assert!(false, err.to_string()),
+            }
+        };
+
+        // Verify tx
+        let bp_gens = BulletproofGens::new(256, 1);
+        match Verifier::verify_tx(tx, &bp_gens) {
```
oleganza: same comment as above - can simply do:

```rust
let vtx = Verifier::verify_tx(tx, &bp_gens).unwrap();
assert_eq!(vtx.log, txlog);
```

token/src/token.rs
```diff
@@ -0,0 +1,190 @@
...
+        Token::issue(program, token, qty)
+            .push(dest.clone())
+            .output(1)
+    }
+
+    /// Returns program that retires a Token.
+    /// TBD: better symmetry here to accept a Token object
+    pub fn retire<'a>(
+        program: &'a mut Program,
+        qty: CompressedRistretto,
```
oleganza: maybe instead of asking for raw components of an `Input` and building it inside `retire`, ask for the `Input` itself? Or even better: ask for an iterator of `Input` (`<I: IntoIterator<Item=Input>>`) and add all the inputs and their retirement? And call the method `retire`.

vickiniu: the thought behind getting the raw components of `Input` is that it's easier to get from the txlog, whereas otherwise the user has to construct the `Input` themselves — but that might be better?

token/src/token.rs
```diff
@@ -0,0 +1,173 @@
...
+            .sign_tx() // stack: issued-value
+    }
+
+    /// Returns program that issues specified quantity of Token,
+    /// outputting it to the destination Predicate.
+    pub fn issue_to<'a>(
+        program: &'a mut Program,
+        token: &Token,
+        qty: u64,
+        dest: &Predicate,
```
oleganza: let's move in dest so the caller decided when to clone and when to not.

token/src/token.rs
oleganza: my thinking was that user would already have a store of utxos, so they'd be constructing or reading Inputs from there - not from the txlog. Txlog would be used only to confirm and record those utxos at an earlier time - when the incoming payment happens and way before the outgoing tx is formed.

token/src/token.rs
```diff
@@ -0,0 +1,173 @@
...
+            .output(1)
+    }
+
+    /// Returns program that retires a Token.
+    /// TBD: better symmetry here to accept a Token object
+    pub fn retire<'a>(
+        program: &'a mut Program,
+        qty: CompressedRistretto,
+        flv: CompressedRistretto,
+        pred: &Predicate,
```
oleganza: move in `pred` instead of cloning a borrowed one. https://rust-lang-nursery.github.io/api-guidelines/flexibility.html#caller-decides-where-to-copy-and-place-data-c-caller-control

token/src/token.rs
```diff
@@ -0,0 +1,173 @@
...
+        txid: TxID,
+    ) -> &'a mut Program {
+        let output = (Commitment::Closed(qty), Commitment::Closed(flv));
+        program
+            .push(Data::Input(Box::new(Input::new(
```
oleganza: there's `impl From<Input> for Data` so you can simply `.push(Input::new(...))`

token/src/token.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Returns the Token's flavor.
+    pub fn flavor(&self) -> Scalar {
+        Value::issue_flavor(
+            &self.issuance_predicate,
+            Data::Opaque(self.metadata.clone()),
+        )
+    }
+
+    /// Returns program that issues specified quantity of Token.
+    pub fn issue<'a>(program: &'a mut Program, token: Token, qty: u64) -> &'a mut Program {
```
oleganza:
```suggestion
    pub fn issue<'a>(&self, program: &'a mut Program, qty: u64) -> &'a mut Program {
```

token/src/token.rs
```diff
@@ -0,0 +1,169 @@
...
+            .var() // stack: qty-var, flv-var
+            .push(Data::Opaque(token.metadata.clone())) // stack: qty-var, flv-var, data
+            .push(token.issuance_predicate.clone()) // stack: qty-var, flv-var, data, flv-pred
+            .issue() // stack: issue-contract
+            .sign_tx() // stack: issued-value
+    }
+
+    /// Returns program that issues specified quantity of Token,
+    /// outputting it to the destination Predicate.
+    pub fn issue_to<'a>(
```
oleganza: `&self` instead of `token: Token`?

token/src/token.rs
oleganza: totally!

token/src/token.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    /// Returns the Token's flavor.
+    pub fn flavor(&self) -> Scalar {
+        Value::issue_flavor(
+            &self.issuance_predicate,
+            Data::Opaque(self.metadata.clone()),
+        )
+    }
+
+    /// Returns program that issues specified quantity of Token.
```
oleganza: This sounds a bit misleading: sounds like it has no side-effects. It does return a program reference only for chaining. But it _modifies_ the program with issuance logic.

Maybe smth like:

```suggestion
    /// Adds instructions to a program that issue a given quantity of this token.
```

token/src/token.rs
```diff
@@ -0,0 +1,164 @@
...
+            .push(Commitment::unblinded(self.flavor())) // stack: qty-var, flv
+            .var() // stack: qty-var, flv-var
+            .push(Data::Opaque(self.metadata.clone())) // stack: qty-var, flv-var, data
+            .push(self.issuance_predicate.clone()) // stack: qty-var, flv-var, data, flv-pred
+            .issue() // stack: issue-contract
+            .sign_tx() // stack: issued-value
+    }
+
+    /// Returns program that issues specified quantity of Token,
+    /// outputting it to the destination Predicate.
```
oleganza: Same as above: highlight mutating nature of this method.

```
/// Adds instructions to a program that issue a given quantity of this token to a given destination predicate.
```

token/src/token.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub fn issue_to<'a>(
+        &self,
+        program: &'a mut Program,
+        qty: u64,
+        dest: Predicate,
+    ) -> &'a mut Program {
+        self.issue(program, qty).push(dest).output(1)
+    }
+
+    /// Returns program that retires a Token.
```
oleganza: same as above:

`/// Adds instructions to a program to retire a given utxo`

token/src/token.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// TBD: better symmetry here to accept a Token object
+    pub fn retire<'a>(
```
oleganza: maybe we'll end up having `token.retire(qty)` which uses `&mut self` to access the storage, collect utxos automatically and return an error if there's not enough utxos. To hide utxos behind the abstraction per #170.

oleganza: but for now simpler `retire` is fine.

zkvm/src/contract.rs
```diff
@@ -52,27 +53,17 @@ enum FrozenItem {
...
-struct FrozenValue {
+pub struct FrozenValue {
     qty: Commitment,
     flv: Commitment,
 }
 
 impl Input {
-    /// Creates a "frozen contract" from payload (which is a vector of (qty, flv)) and pred.
-    /// Serializes the contract, and uses the serialized contract and txid to generate a utxo.
+    /// Serializes the contract, and uses the serialized contract and txid
```
oleganza: This is an implementation detail. Maybe just say "/// Creates an `Input` with a given `Output` and transaction id"?

token/src/token.rs
vickiniu: I had the same thought! added a TBD comment

zkvm/src/contract.rs
```diff
@@ -127,6 +118,11 @@ impl Output {
         buf
     }
 
+    /// Constructs a new Output from payload and predicate
+    pub fn new(payload: Vec<FrozenItem>, predicate: Predicate) -> Self {
```
oleganza: We can allow user construct the struct (mark all Output fields `pub`) themselves because there's no invariant to protect between the predicate and items. So we don't need to have helper methods that imply otherwise and do not make it more convenient: actually, less, since method argument names are not visible at the callsite, but struct field names are.

zkvm/src/contract.rs
```diff
@@ -52,27 +53,17 @@ enum FrozenItem {
...
-        I: IntoIterator<Item = (Commitment, Commitment)>,
-    {
-        let payload: Vec<FrozenItem> = payload
-            .into_iter()
-            .map(|(qty, flv)| FrozenItem::Value(FrozenValue { qty, flv }))
-            .collect();
-
-        let prev_output = Output { payload, predicate };
+    pub fn new(prev_output: Output, txid: TxID) -> Self {
         let utxo = UTXO::from_output(&prev_output.clone().to_bytes(), &txid);
```
oleganza: This output.clone.to_bytes() is unfortunate. There's no win in consuming `self` here because we need to allocate a Vec anyway. Let's change it to take `&self` and replace `.clone().to_bytes()` with `.to_bytes()` here and in the other place below.

zkvm/src/txlog.rs
```diff
@@ -101,7 +102,7 @@ impl Entry {
                 t.commit_u64(b"nonce.t", *maxtime);
             }
             Entry::Output(outstruct) => {
-                t.commit_bytes(b"output", outstruct);
+                t.commit_bytes(b"output", &outstruct.clone().to_bytes());
```
oleganza: Per comment above, if `Output::to_bytes()` takes &self, we don't need to clone the output here.

zkvm/src/contract.rs
```diff
@@ -245,3 +236,10 @@ impl Contract {
         Output { payload, predicate }
     }
 }
+
+impl FrozenValue {
+    /// Construct a new FrozenValue from quantity and flavor
+    pub fn new(qty: Commitment, flv: Commitment) -> Self {
```
oleganza: for the same reason we don't need Output::new, we don't need FrozenValue::new

zkvm/tests/zkvm.rs
```diff
@@ -128,13 +131,11 @@ fn test_helper(program: Vec<Instruction>, keys: &Vec<Scalar>) -> Result<TxID, VM
 
     // Verify tx
     let bp_gens = BulletproofGens::new(256, 1);
+
     match Verifier::verify_tx(tx, &bp_gens) {
```
oleganza:
```rust
    let vtx = Verifier::verify_tx(tx, &bp_gens)?;
    Ok(vtx.id)
}
```
?

PR #162, opened 2019-02-28
zkvm: deny(missing_docs)

### LINKED ISSUE #114: zkvm: deny missing docs
1. Add `#![deny(missing_docs)]` to `lib.rs`.
2. Add documentation lines to all public types and methods until compiler is happy.

Example: https://github.com/interstellar/slingshot/pull/115

oleganza:
This enforces presence of documentation for all items (modules, types, traits, methods, constants) throughout ZkVM.

It allows opt-out via per-item `#[allow(missing_docs)]`. I use it in two places for lists of opcodes and instructions where explicit documentation is not more helpful than the name of an instruction itself: the reader is supposed to check the specification instead.

Closes #114.

### DIFF
--- a/zkvm/src/constraints.rs
+++ b/zkvm/src/constraints.rs
@@ -47,7 +47,6 @@ pub enum Constraint {
     /// Disjunction of two constraints: at least one must evaluate to true.
     /// Created by `or` instruction.
     Or(Box<Constraint>, Box<Constraint>),
-
     // TBD: add `Not(Box<Constraint>)`.
 
     // no witness needed as it's normally true/false and we derive it on the fly during processing.
--- a/zkvm/src/constraints.rs
+++ b/zkvm/src/constraints.rs
@@ -5,36 +5,62 @@ use bulletproofs::{r1cs, r1cs::ConstraintSystem, PedersenGens};
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
 use std::iter::FromIterator;
-use std::ops::{Add, Neg, Sub};
+use std::ops::{Add, Neg};
 
 use crate::encoding;
 use crate::errors::VMError;
 use crate::scalar_witness::ScalarWitness;
 
+/// Variable represents a high-level R1CS variable specified by its
+/// Pedersen commitment. In ZkVM variables are actually indices to a list
+/// of stored commitments to permit commitment reblinding (see `reblind` instruction).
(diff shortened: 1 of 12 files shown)


zkvm/src/constraints.rs
```diff
@@ -5,36 +5,62 @@ use bulletproofs::{r1cs, r1cs::ConstraintSystem, PedersenGens};
...
+
+    /// Conjunction of two constraints: each must evaluate to true.
+    /// Created by `and` instruction.
     And(Box<Constraint>, Box<Constraint>),
+
+    /// Disjunction of two constraints: at least one must evaluate to true.
+    /// Created by `or` instruction.
     Or(Box<Constraint>, Box<Constraint>),
+
+    // TBD: add `Not(Box<Constraint>)`.
```
vickiniu: I think this gave a `cargo fmt` error? from travis

zkvm/src/ops.rs
```diff
@@ -298,12 +309,14 @@ impl Instruction {
...
 macro_rules! def_op {
     ($func_name:ident, $op:ident) => (
+           /// Adds a `$func_name` instruction.
            pub fn $func_name(&mut self) -> &mut Program{
              self.0.push(Instruction::$op);
              self
         }
     );
     ($func_name:ident, $op:ident, $type:ty) => (
+           /// Adds a `$func_name` instruction.
```
vickiniu: nit: maybe specify "Adds a `$func_name` instruction with argument of `$type`"

zkvm/src/types.rs
```diff
@@ -171,7 +203,7 @@ impl Data {
         }
     }
 
-    /// Encodes blinded Data values.
+    /// Encodes the data item to a byte buffer.
```
vickiniu: specify that we encode the opaque version of data?

zkvm/src/ops.rs
oleganza: $type is too generic here, so it won't sound nice

zkvm/src/types.rs
oleganza: is it better this way?

> /// Encodes the data item to an opaque bytestring.

zkvm/src/constraints.rs
oleganza: fixed, thx

zkvm/src/types.rs
vickiniu: that's great! thanks :)

PR #164, opened 2019-02-28
zkvm: implement `delegate` instruction

### LINKED ISSUE #155: zkvm: implement `delegate` instruction
[`delegate`](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#delegate) checks a signature with a predicate-as-a-pubkey over a given program string and executes it.

See also related issue about `call`: #154.

UPD: @decentralisedkev seems to be working on this issue.

decentralisedkev:
Closes #155

### DIFF
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -1030,7 +1030,7 @@ Then, the VM executes the current program till completion:
 1. Each instruction is read at the current program offset, including its immediate data (if any).
 2. Program offset is advanced immediately after reading the instruction to the next instruction.
 3. The instruction is executed per [specification below](#instructions). If the instruction fails, VM exits early with an error result.
-4. If VM encounters [`call`](#call) or [`delegate`](#delegate) instruction, the current program and the offset are saved in the program stack, and the new program with offset zero is set as the current program. 
+4. If VM encounters [`call`](#call) or [`delegate`](#delegate) instruction, the new program with offset zero is set as the current program. The next iteration of the vm will start from the beginning of the new program.
 5. If the offset is less than the current program’s length, a new instruction is read (go back to step 1).
 6. Otherwise (reached the end of the current program):
    1. If the program stack is not empty, pop top item from the program stack and set it to the current program. Go to step 5.
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -1727,27 +1727,30 @@ or if the third from the top item is not a [contract](#contract-type).
 
 _contract prog sig_ **delegate** → _results..._
 
-1. Pops [data](#data-type) `sig`, [data](#data-type) `prog` and the [contract](#contract-type) from the stack.
-2. Instantiates the [transcript](#transcript):
+1. Pop [data](#data-type) `sig`, [data](#data-type) `prog` and the [contract](#contract-type) from the stack.
+
+2. Push the [Contract payload](#contract-payload) onto the stack.
+
+3. Instantiate the [transcript](#transcript):
     ```
     T = Transcript("ZkVM.delegate")
     ```
-3. Commits the program `prog` to the transcript:
+4. Commit the program `prog` to the transcript:
     ```
     T.commit("prog", prog)
     ```
-4. Extracts nonce commitment `R` and scalar `s` from a 64-byte data `sig`:
+5. Extract nonce commitment `R` and scalar `s` from a 64-byte data `sig`:
     ```
     R = sig[ 0..32]
     s = sig[32..64]
     ```
-5. Performs the [signature protocol](#aggregated-signature) using the transcript `T`, public key `contract.predicate` and the values `R` and `s`:
+6. Perform the [signature protocol](#aggregated-signature) using the transcript `T`, public key `contract.predicate` and the values `R` and `s`:
     ```
     (s = dlog(R) + e·dlog(P))
     s·B  ==  R + e·P
     ```
-6. Adds the statement to the list of [deferred point operations](#deferred-point-operations).
-7. Saves the current program in the program stack, sets the `prog` as current and [runs it](#vm-execution).
+7. Add the statement to the list of [deferred point operations](#deferred-point-operations).
+8. Set the `prog` as current.
 
 Fails if:
 1. the `sig` is not a 64-byte long [data](#data-type),
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -1729,7 +1729,7 @@ _contract prog sig_ **delegate** → _results..._
 
 1. Pop [data](#data-type) `sig`, [data](#data-type) `prog` and the [contract](#contract-type) from the stack.
 
-2. Push the [Contract payload](#contract-payload) onto the stack.
+2. Place the [contract payload](#contract-payload) on the stack (last item on top), discarding the contract.
 
 3. Instantiate the [transcript](#transcript):
     ```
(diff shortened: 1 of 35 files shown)


zkvm/src/vm.rs
```diff
@@ -567,6 +570,29 @@ where
         })
     }
 
+    fn delegate(&mut self) -> Result<(), VMError> {
+        let sig = self.pop_item()?.to_data()?.to_bytes();
+        let prog = self.pop_item()?.to_data()?.to_bytes();
+        let contract = self.pop_item()?.to_contract()?;
+
+        let R = SliceReader::parse(&sig[0..32], |r| r.read_point())?;
```
oleganza: this is undoing the safety feature of `SliceReader::parse`: you are not longer checking that `sig` is fully read.

Instead use one `parse` call to guarantee that all bytes are read:

```rust
let (R,s) = SliceReader::parse(&sig, |r| {
    Ok((r.read_point()?, r.read_scalar()?))
})?;
```

zkvm/src/vm.rs
```diff
@@ -567,6 +573,33 @@ where
...
 
+    fn delegate(&mut self) -> Result<(), VMError> {
+        // Pop signature and put into signature struct
+        let sig = self.pop_item()?.to_data()?.to_bytes();
+        if sig.len() != 64 {
+            return Err(VMError::FormatError);
+        }
+        let mut buf = [0u8; 64];
+        buf[..].copy_from_slice(&sig);
+        let signature = Signature::from_bytes(buf)?;
```
oleganza: This code does not check that `sig` is fully read.

Let's add `read_u8x64` method to `encoding::SliceReader` and do `Signature::from_bytes(SliceReader::parse(&sig){|r| r.read_u8x64() }?)`.

zkvm/src/vm.rs
```diff
@@ -115,6 +119,8 @@ pub trait Delegate<CS: r1cs::ConstraintSystem> {
     /// Returns Ok(None) if there is no more instructions to execute.
     fn next_instruction(&mut self, run: &mut Self::RunType)
         -> Result<Option<Instruction>, VMError>;
+
+    fn next_run(&mut self, prog: &[u8]) -> Self::RunType;
```
oleganza: `next_run` is a bit misleading - it implies that the creating the run has sideeffects, while it should not - VM manages the runstack, not the delegate. How about:

```suggestion
    fn new_run(&self, prog: &[u8]) -> Self::RunType;
```

(note the self is immutably borrowed)

zkvm/src/vm.rs
decentralisedkev: sorry didn't see this

decentralisedkev:
For the prover.new_run, I'm unsure of how to convert from the Vec<u8> to Vec<Instructions> we need without failing

decentralisedkev:
I changed new_run to return a Result, because Prover's new_run could fail. Let me know if this needs to be changed

zkvm/src/prover.rs
```diff
@@ -57,6 +58,21 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
         Ok(run.program.pop_front())
     }
 
+    fn new_run(&self, program: &[u8]) -> Result<Self::RunType, VMError> {
```
oleganza: I think the program must be Data item, so for the prover it is converted to Data::Program (there is likely already a method for it), because that’s what allows having witnesses during vm execution.

zkvm/src/prover.rs
```diff
@@ -57,6 +57,16 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
         Ok(run.program.pop_front())
     }
 
+    fn new_run(&self, program: Data) -> Result<Self::RunType, VMError> {
+        let instructions = match program {
```
oleganza: maybe factor this out to `Data::to_program` converter like we do with `to_predicate`, `to_commitment` etc?

zkvm/src/vm.rs
```diff
@@ -579,6 +583,35 @@ where
         })
     }
 
+    fn delegate(&mut self) -> Result<(), VMError> {
+        // Signature
+        let sig = self.pop_item()?.to_data()?.to_bytes();
+        if sig.len() != 64 {
```
oleganza: this check is not needed - `::parse` does it already

zkvm/src/vm.rs
```diff
@@ -579,6 +583,35 @@ where
...
+            return Err(VMError::FormatError);
+        }
+        let signature = Signature::from_bytes(SliceReader::parse(&sig, |r| r.read_u8x64())?)?;
+
+        // Program
+        let prog = self.pop_item()?.to_data()?;
+        let prog_bytes = prog.clone().to_bytes();
+
+        // Verification key
+        let contract = self.pop_item()?.to_contract()?;
```
oleganza: OOPS, my bad - there's a bug in the spec - we need to place all payload of the contract onto the stack before we continue execution of the new program. Could you please add this step to the spec and change the code accordingly?

```
        for item in contract.payload.into_iter() {
            self.push_item(item);
        }
```

decentralisedkev: Yep sure!

zkvm/docs/zkvm-spec.md
```diff
@@ -1727,27 +1727,30 @@ or if the third from the top item is not a [contract](#contract-type).
 
 _contract prog sig_ **delegate** → _results..._
 
-1. Pops [data](#data-type) `sig`, [data](#data-type) `prog` and the [contract](#contract-type) from the stack.
-2. Instantiates the [transcript](#transcript):
+1. Pop [data](#data-type) `sig`, [data](#data-type) `prog` and the [contract](#contract-type) from the stack.
+
+2. Push the [Contract payload](#contract-payload) onto the stack.
```
oleganza: maybe use the same wording as in `signtx`?
```suggestion
2. Place the [contract payload](#contract-payload) on the stack (last item on top), discarding the contract.
```

zkvm/src/vm.rs
```diff
@@ -579,6 +583,37 @@ where
...
+        for item in contract.payload.into_iter() {
+            self.push_item(item);
+        }
+
+        // Verification key from predicate
+        let verification_key = contract.predicate.to_key()?;
+
+        // Verify signature using Verification key, over the message `program`
+        let mut t = Transcript::new(b"ZkVM.delegate");
+        t.commit_bytes(b"prog", &prog_bytes);
```
oleganza: let's avoid extra variables (`prog_bytes`) to make the logic cleaner:
```suggestion
        t.commit_bytes(b"prog", &prog.clone().to_bytes());
```

zkvm/src/types.rs
```diff
@@ -182,6 +182,14 @@ impl Data {
         }
     }
 
+    /// Downcast the data item to a `Program` type.
+    pub fn to_program(self) -> Result<Vec<Instruction>, VMError> {
+        match self {
+            Data::Program(program) => Ok(program),
+            _ => Err(VMError::InconsistentWitness),
```
oleganza: Other downcast methods use `TypeNot<Type>`, let's do the same here?:

```suggestion
            _ => Err(VMError::TypeNotProgram),
```

zkvm/src/verifier.rs
```diff
@@ -67,6 +67,10 @@ impl<'a, 'b> Delegate<r1cs::Verifier<'a, 'b>> for Verifier<'a, 'b> {
         Ok(Some(instr))
     }
 
+    fn new_run(&self, prog: Data) -> Result<Self::RunType, VMError> {
+        Ok(VerifierRun::new(prog.to_bytes().to_vec()))
```
oleganza: `data.to_bytes()` already gives a `Vec<u8>`, no?

```suggestion
        Ok(VerifierRun::new(prog.to_bytes()))
```

PR #166, opened 2019-02-28
zkvm: Add rangeproof instruction

### LINKED ISSUE #151: zkvm: implement `range` instruction
This should be fairly simple as we already have the `add_range_proof` helper in the VM.

decentralisedkev:
Closes #151

### DIFF
--- a/zkvm/src/vm.rs
+++ b/zkvm/src/vm.rs
@@ -51,17 +51,17 @@ pub struct Tx {
 
 /// Represents a verified transaction: a txid and a list of state updates.
 pub struct VerifiedTx {
-    /// Header metadata
+    /// Transaction header
     pub header: TxHeader,
 
     /// Transaction ID
     pub id: TxID,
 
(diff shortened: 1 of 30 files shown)


zkvm/src/vm.rs
```diff
@@ -317,6 +317,13 @@ where
         Ok(())
     }
 
+    fn range(&mut self, i: u8) -> Result<(), VMError> {
+        let expr = self.pop_item()?.to_expression()?;
+        self.add_range_proof(i as usize, expr.clone());
```
oleganza: Let's make `add_range_proof` failable and check that `i` is between `1` and `64` (0 and >64 are not allowed).

oleganza:
```
error: missing documentation for a variant
   --> src/errors.rs:120:5
    |
120 |     InvalidBitrange,
    |     ^^^^^^^^^^^^^^^
    |
note: lint level defined here
   --> src/lib.rs:1:9
    |
1   | #![deny(missing_docs)]
    |         ^^^^^^^^^^^^
```

PR #169, opened 2019-03-01
slidechain: check for export tx more robustly

debnil:
Checks for the export transaction more robustly. A log statement is added to the called export contract, and the goroutine watching for the export transaction also checks for the existence of the accompanying tuple in the transaction log.

### DIFF
--- a/slidechain/watch.go
+++ b/slidechain/watch.go
@@ -145,9 +145,9 @@ func (c *Custodian) watchExports(ctx context.Context) {
 				continue
 			}
 
-			exportDataLogItem := tx.Log[1]
+			exportDataInfoItem := tx.Log[1]
 			var info pegOut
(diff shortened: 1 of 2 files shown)


slidechain/watch.go
```diff
@@ -136,9 +137,17 @@ func (c *Custodian) watchExports(ctx context.Context) {
 				continue
 			}
 
-			logItem := tx.Log[1]
+			exportSeedLogItem := tx.Log[len(tx.Log)-3]
+			if exportSeedLogItem[0].(txvm.Bytes)[0] != txvm.LogCode {
+				continue
+			}
+			if bytes.Compare(exportSeedLogItem[1].(txvm.Bytes), exportContract1Seed[:]) != 0 {
```
bobg: `if !bytes.Equal(...)`

PR #173, opened 2019-03-01
keytree: specify nightly

tessr:
keytree requires bulletproofs.
bulletproofs requires non-lexical lifetimes. 
therefore bulletproofs and keytree require this.


tessr:
jk!

PR #174, opened 2019-03-01
 keytree: generate Xpub from Xprv

tessr:
This implements https://github.com/interstellar/slingshot/blob/main/keytree/keytree.md#convert-xprv-to-xpub

### DIFF
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -1,6 +1,8 @@
 #![deny(missing_docs)]
 //! Implementation of the key tree protocol, a key blinding scheme for deriving hierarchies of public keys.
 
+use bulletproofs::PedersenGens;
+use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
 use rand::{CryptoRng, RngCore};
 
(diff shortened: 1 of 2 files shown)


keytree/src/lib.rs
```diff
@@ -20,6 +22,23 @@ impl Xprv {
...
 
+/// Xpub represents an extended public key.
+pub struct Xpub {
+    point: CompressedRistretto,
+    dk: [u8; 32],
+}
+
+impl Xpub {
+    /// Returns a new Xpub, generated from the provided Xprv.
+    pub fn from_xprv(xprv: Xprv) -> Self {
```
oleganza: suggestion: instead of `Xpub::from_xprv(xprv)` let's turn this around, so it's `xprv.to_xpub()` instead:

```rust
impl Xprv {
    pub fn to_xpub(&self) -> Self {...}
}
```

Also: we don't need to consume xprv to create an xpub, so using a reference (&Xprv in your code, &self in my suggestion) is better.

keytree/src/lib.rs
```diff
@@ -18,6 +26,15 @@ impl Xprv {
         rng.fill_bytes(&mut dk);
         Xprv { scalar, dk }
     }
+
+    /// Returns a new Xpub, generated from the provided Xprv.
+    pub fn to_xpub(xprv: &Self) -> Xpub {
```
oleganza: use the keyword `self` as a designated first argument which turns the _function_ into a _method_:
```suggestion
    pub fn to_xpub(&self) -> Xpub {
```

oleganza: Note: the method can still be called as `Xprv::to_xpub(&xprv)`, but with use of `self` it can _also_ be called as a usual OOP-y method: `xprv.to_xpub()`.

tessr: Thank you! I got it, now

PR #180, opened 2019-03-01
slidechain: simplify zeroval creation for post-peg-out finalize

debnil:
Simplifies the creation of a zero value for the `finalize` instruction in post-peg-out. Previously, we created a new one using the `nonce` instruction. Now, we split off the zero-value from the value contained in the existing contract.


slidechain/export.go
```diff
@@ -58,34 +58,33 @@ const (
...
+	jump:$checksig       #                                                                                            
+	                     #                                                                                            
+	$doretire            #                                                                                            
+	                     #  zeroval, {exporter}, value, json                                                          
+	put put drop         #  zeroval                                     json, value                                   
+	x'%x' contract call  #  zeroval                                                                                   
+	                     #                                                                                                                                                                    
+	$checksig            #                                                                                            
+	put                  #                                              zeroval
```
debnil: I realize that returning these arguments in this order is not ideal, because it forces the caller to swap their order (to call `finalize` before `sigchecker`). Unfortunately, if the zero value is on the contract stack, it seems to be getting included on the contract stack of the contract resulting from `yield`, so there is nothing to `put` on the arg stack. This doesn't seem to be expected behavior from [yield](https://github.com/chain/txvm/blob/main/specifications/txvm.md#yield) -- is there another way to do this and avoid the swapping?

slidechain/export.go
```diff
@@ -58,34 +58,33 @@ const (
...
-	put put drop         #                                     json, value                                   
-	x'%x' contract call  #                                                                                   
-	                     #                                                                                                        
-	$checksig            #                                                                                   
-	[%s] yield           #                                     sigchecker                                    
-                                                                                   
+	                     #  con stack                                   arg stack                 log
+	                     #  ---------                                   ---------                 ---
+	                     #  {exporter}, json, val                       selector                                      
+	0 split 3 bury swap  #  zeroval, {exporter}, value, json            selector
```
bobg: TxVM style nit[*] - prefer the assembler macro `splitzero` over `0 split`.

[*] - ATTINWDTS[**]
[**] - Acknowledging that there is no well-defined TxVM style

slidechain/export.go
bobg: Yes! You could do this instead:

```
...
$checksig
[%s] contract put # to put a sigcheck contract on the arg stack
put               # to put a zeroval on the arg stack on top of the sigcheck contract
```

Doing `X contract put` is equivalent to `X yield` except that the first one creates a new contract, with a new seed, while the second one just updates the current contract's program, but retains its seed. For the sigchecker, we don't care about the seed, so it's OK to replace one form with the other.

In both cases, the caller is obliged to invoke the contract on the arg stack in order to clear it from the VM.

slidechain/export.go
bobg: Those empty brackets were supposed to contain a double star...

PR #181, opened 2019-03-01
zkvm: implement `call` instruction

### LINKED ISSUE #154: zkvm: implement `call` instruction
[`call`](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#call) opens up a Predicate commitment to a program and executes it.

See also related issue about `delegate`: #155.

decentralisedkev:
Closes #154

### DIFF
--- a/zkvm/src/vm.rs
+++ b/zkvm/src/vm.rs
@@ -553,6 +553,28 @@ where
         Ok(())
     }
 
+    fn call(&mut self) -> Result<(), VMError> {
+        // Pop program contract and predicate
+        let prog = self.pop_item()?.to_data()?;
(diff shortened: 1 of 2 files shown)


zkvm/src/vm.rs
```diff
@@ -736,6 +754,13 @@ where
             .map(|(content, _)| content)
     }
 
+    fn set_program(&mut self, prog: Data) -> Result<(), VMError> {
```
oleganza: Nice! How about a more accurate name, something like:

```suggestion
    fn continue_with_program(&mut self, prog: Data) -> Result<(), VMError> {
```

PR #184, opened 2019-03-01
spacesuit: implement type-safe `BitRange` for range proofs

### LINKED ISSUE #167: spacesuit: make sure rangeproofs are not used with `n > 64`
We can silently overflow u64 in release builds if `n>64` in the `range_proof` gadgets.

decentralisedkev:
Closes #167

### DIFF
--- a/spacesuit/src/bit_range.rs
+++ b/spacesuit/src/bit_range.rs
@@ -12,8 +12,7 @@ impl BitRange {
         }
     }
 
-    /// max returns a BitRange
-    /// representing a 64-bit usize
+    /// Returns 64-bit range
     pub fn max() -> Self {
         BitRange(64)
     }
--- a/spacesuit/src/bit_range.rs
+++ b/spacesuit/src/bit_range.rs
@@ -2,7 +2,7 @@
 pub struct BitRange(usize);
 
 impl BitRange {
-    /// Returns BitRange if n is <= 64>
+    /// Returns Some(BitRange) if `n` is ≤ 64.
     /// Otherwise returns None.
     pub fn new(n: usize) -> Option<Self> {
         if n > 64 {
--- a/spacesuit/src/bit_range.rs
+++ b/spacesuit/src/bit_range.rs
@@ -1,2 +1,18 @@
 /// Represents a usize with value in the range [0,64]
 pub struct BitRange(usize);
+
+impl BitRange {
+    fn new(n: usize) -> Option<Self> {
+        if n > 64 {
+            None
+        } else {
+            Some(BitRange(n))
+        }
+    }
+}
+
+impl Into<usize> for BitRange {
+    fn into(self) -> usize {
+        self.0
+    }
+}
--- a/spacesuit/src/bit_range.rs
+++ b/spacesuit/src/bit_range.rs
@@ -1,4 +1,5 @@
 /// Represents a usize with value in the range [0,64]
+#[derive(Copy, Clone, Debug)]
 pub struct BitRange(usize);
 
 impl BitRange {
@@ -15,6 +16,11 @@ impl BitRange {
     /// Returns 64-bit range
     pub fn max() -> Self {
         BitRange(64)
+    }
+
+    /// Converts BitRange to a `u8`
+    pub fn to_u8(self) -> u8 {
+        self.0 as u8
     }
 }
 
--- a/spacesuit/src/bit_range.rs
+++ b/spacesuit/src/bit_range.rs
@@ -1,0 +1,2 @@
+/// Represents a usize with value in the range [0,64]
+pub struct BitRange(usize);
--- a/spacesuit/src/bit_range.rs
+++ b/spacesuit/src/bit_range.rs
@@ -1,12 +1,13 @@
+use bulletproofs::r1cs::R1CSError;
 /// Represents a usize with value in the range [0,64]
 pub struct BitRange(usize);
 
 impl BitRange {
-    fn new(n: usize) -> Option<Self> {
+    pub fn new(n: usize) -> Result<Self,Err> {
         if n > 64 {
-            None
+            Err("Invalid Bitrange value. Value must be less than 64")
         } else {
-            Some(BitRange(n))
+            Ok(BitRange(n))
         }
     }
 }
--- a/spacesuit/src/bit_range.rs
+++ b/spacesuit/src/bit_range.rs
@@ -22,6 +22,12 @@ impl BitRange {
     pub fn to_u8(self) -> u8 {
         self.0 as u8
     }
+
+    /// Checks whether the Bitrange lies
+    /// between 0 and 64
+    pub fn is_valid(&self) -> bool {
+        return self.0 > 64;
+    }
 }
 
 impl Into<usize> for BitRange {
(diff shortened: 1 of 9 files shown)


decentralisedkev:
Changed from Option to Result; my understanding was that if the protocol should stop immediately if the value is not returned, then we should use Result. The None value, in this case signifying the absence of a bitrange would not be valid if I understand correctly.

spacesuit/src/cloak.rs
```diff
@@ -34,7 +35,12 @@ pub fn cloak<CS: ConstraintSystem>(
     // Range Proof
     // Check that each of the quantities in `outputs` lies in [0, 2^64).
     for output in outputs {
-        range_proof(cs, output.q.into(), output.assignment.map(|v| v.q), 64)?;
+        range_proof(
+            cs,
+            output.q.into(),
+            output.assignment.map(|v| v.q),
+            BitRange::new(64)?,
```
oleganza: suggestion: add method `::max()` that does not fail and makes a 64-bit bitrange

spacesuit/src/bit_range.rs
```diff
@@ -0,0 +1,23 @@
+use bulletproofs::r1cs::R1CSError;
+/// Represents a usize with value in the range [0,64]
+pub struct BitRange(usize);
+
+impl BitRange {
+    /// Returns BitRange if n is <= 64>
+    /// Otherwise Errors
+    pub fn new(n: usize) -> Result<Self, R1CSError> {
```
oleganza: R1CSError is from another framework, not from spacesuit. For such simple cases when there is only one kind of error (see also [`Scalar::from_canonical_bytes`](https://github.com/dalek-cryptography/curve25519-dalek/blob/master/src/scalar.rs#L227-L237)) it is customary to return `Option` instead of `Result`.

One can always convert a None to an error in a higher-level code via `.ok_or(SomeError)?`

spacesuit/src/bit_range.rs
```diff
@@ -0,0 +1,26 @@
+/// Represents a usize with value in the range [0,64]
+pub struct BitRange(usize);
+
+impl BitRange {
+    /// Returns BitRange if n is <= 64>
```
oleganza:
```suggestion
    /// Returns Some(BitRange) if `n` is ≤ 64
```

spacesuit/src/bit_range.rs
```diff
@@ -0,0 +1,26 @@
...
+    /// Otherwise Errors
```
oleganza:
```suggestion
    /// Otherwise returns None.
```

spacesuit/src/bit_range.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub fn new(n: usize) -> Option<Self> {
+        if n > 64 {
+            None
+        } else {
+            Some(BitRange(n))
+        }
+    }
+
+    /// max returns a BitRange
+    /// representing a 64-bit usize
```
oleganza: That's weird wording. Maybe state simply "/// Returns 64-bit range"?

decentralisedkev: Yeah, this sounds better

spacesuit/src/bit_range.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /// Converts BitRange to a `u8`
+    pub fn to_u8(self) -> u8 {
+        self.0 as u8
+    }
+
+    /// Checks whether the Bitrange lies
+    /// between 0 and 64
+    pub fn is_valid(&self) -> bool {
```
oleganza: Opinion: this method is unnecessary on philosophical grounds. Validity of the BitRange must be encoded _in the type_ (see also `Tx` vs `ValidTx`). We already make sure it's impossible to construct BitRange with invalid integer, so any BitRange instance is automatically a valid one.

That's the whole point of having a separate type instead of `usize`. So we can avoid repetitive checks all over the place that can be sometimes forgotten. "Encode protocol requirements in the type system" and all that :-)

zkvm/src/ops.rs
```diff
@@ -184,8 +184,8 @@ impl Instruction {
             Opcode::Mul => Ok(Instruction::Mul),
             Opcode::Eq => Ok(Instruction::Eq),
             Opcode::Range => {
-                let bit_width = program.read_u8()?;
-                Ok(Instruction::Range(bit_width))
+                let bit_width = BitRange::new(program.read_u8()? as usize);
```
oleganza: generally it's a better style to add `.ok_or` before the variable is created, so all variables in the program are maximally "valid".

zkvm/src/vm.rs
```diff
@@ -321,12 +322,12 @@ where
         Ok(())
     }
 
-    fn range(&mut self, i: u8) -> Result<(), VMError> {
-        if i < 1 || i > 64 {
+    fn range(&mut self, i: BitRange) -> Result<(), VMError> {
+        if !i.is_valid() {
```
oleganza: No need for this check - it's already encoded in the BitRange type. See the comment above.

spacesuit/src/bit_range.rs
decentralisedkev: got it, I thought maybe someone could do BitRange(65) somewhere up the line

decentralisedkev: oh right:

```
            Opcode::Range => {
                let bit_width =
                    BitRange::new(program.read_u8()? as usize).ok_or(VMError::FormatError)?;
                Ok(Instruction::Range(bit_width))
            }
```

spacesuit/src/bit_range.rs
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+
+    /// Returns 64-bit range
+    pub fn max() -> Self {
+        BitRange(64)
+    }
+
+    /// Converts BitRange to a `u8`
+    pub fn to_u8(self) -> u8 {
```
oleganza: why not `Into<u8>` like with `Into<usize>`?

decentralisedkev: changed. It could not determine the type with .into when there was more than one into impl for some reason. I used an extra line to help the compiler : https://github.com/interstellar/slingshot/pull/184/commits/145ebff13dde8d2aad74a626bb87a7352f59aae7#diff-9cf510fc0d33929cbc49a13f477fe8f7R259

Let me know if there is a better solution

oleganza:
Looks great! Just needs a few fixes to make travis happy.

oleganza:
Oh, could you please update the `zkvm-spec.md` to allow 0 as well 1–64?

spacesuit/src/bit_range.rs
oleganza: I'm curious what was the error? It seems to me the `.push` is clearly expecting `u8`, and the Range op has `BitRange` type, so there should be no ambiguity about `Into<u8>`. 🤔

decentralisedkev: It was "Could not infer type `T`" for ops.rs and "Into int32 was not implemented" for range_proof.rs . I thought the same thing too

oleganza:
last nitpick: i'd remove `InvalidBitrange` error as it's unused now. And rebase over `main`.

I've already done these two things here, but i cannot push to your branch, so you may peek at this one if you'd like: https://github.com/interstellar/slingshot/tree/pull/184/head

PR #188, opened 2019-03-04
zkvm/txlog: construct and verify Merkle proofs of inclusion

vickiniu:
Allows building and verifying proofs that a given `Entry` is in the Merkle tree whose root is `TxID`.


zkvm/src/txlog.rs
```diff
@@ -76,6 +84,66 @@ impl TxID {
         entry.commit_to_transcript(&mut t);
         t.challenge_bytes(b"merkle.leaf", result);
     }
+
+    /// Computes the Merkle proof of inclusion required to validate an
+    /// entry at index i, not including the original item i.
+    /// Fails when requested index is out of bounds
+    pub fn proof(
```
oleganza: Would be good to separate public API from internal recursive API (especially if it's going to be rewritten into a non-recursive one later). E.g. have `pub fn proof(list, index) -> Vec<MerkleHash>` and keep transcript creation and recursion behind it with private `fn subproof(t, list, index &mut vec)` (which may also become infailable, as the public API already has checked for the possible errors).

oleganza: See how `from_log` is public and `node` is recursive and private because it assumes good inputs that are provided internally-only.

zkvm/src/txlog.rs
```diff
@@ -31,6 +32,13 @@ pub struct TxID(pub [u8; 32]);
 #[derive(Copy, Clone, PartialEq, Debug)]
 pub struct UTXO(pub [u8; 32]);
 
+/// MerkleHash represents a step in a Merkle proof of inclusion.
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub enum MerkleHash {
```
oleganza: Since we'll probably need proofs for multiple items, would it make sense to compute the tree with all intermediate hashes (so we don't repeat them again and again), and then query necessary paths from it at no cost?

oleganza: Bonus points if we can lazily compute/cache/reuse internal hashes, so that if we use the tree for one-two items, we don't pay the cost of hashing all the nodes.

zkvm/src/txlog.rs
```diff
@@ -76,6 +84,73 @@ impl TxID {
...
+    }
+
+    fn subproof(
+        t: Transcript,
+        list: &[Entry],
+        index: usize,
+        result: &mut Vec<MerkleHash>,
+    ) -> Result<(), VMError> {
+        match list.len() {
+            0 => Err(VMError::InvalidMerkleProof),
```
oleganza: seems like this case is already checked in the `proof`, allowing us to make `subproof` non-failable.

zkvm/src/txlog.rs
```diff
@@ -76,6 +84,73 @@ impl TxID {
...
+            1 => Ok(()),
+            n => {
+                let k = n.next_power_of_two() / 2;
+                if index >= k {
+                    let mut lefthash = [0u8; 32];
```
oleganza: maybe i missed something, but seems like `lefthash`/`righthash` remain zeroed and not actually computed?

oleganza:
If that's of any help, here's a TxVM merkle tree implementation in Rust (in the enclave project): https://github.com/interstellar/i10r/blob/dbab14815e44c001a2ef0d6e9a23af8a2338161b/enclave/enclave/src/crypto/protocol/src/merkle.rs

zkvm/src/txlog.rs
vickiniu: they get set when we pass into `Self::node(...)` !

oleganza: Seems like we are hashing everything quadratic number of times? `::node` recurses by itself, and then we descend one level down, and call `::node` again, that walks over the same hashes.

WDYT of simply building the whole tree and sampling it? Cf. https://github.com/interstellar/slingshot/pull/188/files#r262289600

vickiniu: that makes a lot of sense to me! will start working on that :)

zkvm/src/txlog.rs
```diff
@@ -31,6 +32,27 @@ pub struct TxID(pub [u8; 32]);
...
+    root: MerkleNode,
+    size: usize,
+}
+
+/// MerkleNode represents a precomputed node in the Merkle tree
+/// for easily computing proofs of inclusion.
+struct MerkleNode {
+    left: Option<Box<MerkleNode>>,
+    right: Option<Box<MerkleNode>>,
+    node: [u8; 32],
```
oleganza: Maybe "MerkleNode.hash" sounds better?

```suggestion
    hash: [u8; 32],
```

zkvm/src/txlog.rs
```diff
@@ -31,6 +32,27 @@ pub struct TxID(pub [u8; 32]);
...
 
+/// MerkleHash represents a step in a Merkle proof of inclusion.
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub enum MerkleHash {
+    Left([u8; 32]),
+    Right([u8; 32]),
+}
+
+/// MerkleRoot defines the root of a Merkle tree.
+pub struct MerkleRoot {
```
oleganza: Maybe like this? So we can say "MerkleTree.root" instead of "MerkleRoot.root".
```suggestion
pub struct MerkleTree {
```

zkvm/src/txlog.rs
```diff
@@ -76,6 +98,84 @@ impl TxID {
         entry.commit_to_transcript(&mut t);
         t.challenge_bytes(b"merkle.leaf", result);
     }
+
+    /// Builds a precomputed Merkle tree for the entries.
+    pub fn build_tree(list: &[Entry]) -> (Self, MerkleRoot) {
+        let t = Transcript::new(b"ZkVM.txid");
+        let root = Self::build_node(t, list);
+        (
+            TxID(root.node),
```
oleganza: i don't think it's necessary to return txid here — we can simply return MerkleTree object and then query its `.root` field or method. Also let's not tie it to txid too much — we are likely to factor this out for other use in the blockchain headers.

oleganza: That said, it's fine to use `&[Entry]`, but later it'll be something like `&[I] where I: MerkleItem`

zkvm/src/txlog.rs
```diff
@@ -76,6 +98,84 @@ impl TxID {
...
+            }
+        }
+    }
+
+    /// Verifies that an entry satisfies the Merkle proof of inclusion
+    /// for a given TxID
+    pub fn verify_proof(&self, entry: Entry, proof: Vec<MerkleHash>) -> Result<(), VMError> {
+        let transcript = Transcript::new(b"ZkVM.txid");
+        let mut result = [0u8; 32];
+        Self::node(transcript.clone(), &[entry], &mut result);
```
oleganza: Should it be a leaf instead of an intermediary node?
```suggestion
        Self::leaf(transcript.clone(), &[entry], &mut result);
```

zkvm/src/txlog.rs
```diff
@@ -118,3 +161,147 @@ impl Entry {
...
+    /// Constructs a new MerkleTree based on the input list of entries.
+    pub fn new(list: &[Entry]) -> Option<MerkleTree> {
+        if list.len() == 0 {
+            return None;
+        }
+        let t = Transcript::new(b"ZkVM.txid");
+        Some(Self::build_node(t, list))
+    }
+
+    fn build_node(mut t: Transcript, list: &[Entry]) -> MerkleTree {
```
oleganza: ?
```suggestion
    fn build_tree(mut t: Transcript, list: &[Entry]) -> MerkleTree {
```

zkvm/src/txlog.rs
```diff
@@ -76,6 +91,34 @@ impl TxID {
...
+                    t.challenge_bytes(b"merkle.node", &mut result);
+                }
+                MerkleNeighbor::Right(r) => {
+                    t.commit_bytes(b"L", &result);
+                    t.commit_bytes(b"R", r);
+                    t.challenge_bytes(b"merkle.node", &mut result);
+                }
+            }
+        }
+        if self.0 == result {
```
oleganza: use `subtle` and it's `.ct_eq` method for const-time comparison.

zkvm/src/txlog.rs
```diff
@@ -118,3 +167,142 @@ impl Entry {
...
+                t.commit_bytes(b"R", right.hash());
+                t.challenge_bytes(b"merkle.node", &mut node);
+                return MerkleNode::Node(node, Box::new(left), Box::new(right));
+            }
+        }
+    }
+
+    fn subproof(&self, t: Transcript, index: usize, size: usize, result: &mut Vec<MerkleNeighbor>) {
+        let k = size.next_power_of_two() / 2;
+        if index >= k {
```
oleganza: maybe it'll be clearer if you transposed this if/else and matches? Meaning, match first, and if Leaf - return. Then, inside the ::Node case check which neighbour you need and drill in accordingly.

vickiniu: ah, yes! doing it that way is much nicer :)

zkvm/src/txlog.rs
```diff
@@ -118,3 +167,137 @@ impl Entry {
...
+    fn build_tree(mut t: Transcript, list: &[Entry]) -> Self {
+        match list.len() {
+            0 => {
+                let mut leaf = [0u8; 32];
+                TxID::empty(t, &mut leaf);
+                return MerkleNode::Leaf(leaf);
+            }
+            1 => {
+                let mut leaf = [0u8; 32];
+                TxID::leaf(t, &list[0], &mut leaf);
```
oleganza: let's move `empty/leaf/node` functions from under TxID to MerkleTree.

Also since `build_tree` is not even taking `&self`, we can move it under MerkleTree too.

zkvm/src/txlog.rs
```diff
@@ -31,6 +33,25 @@ pub struct TxID(pub [u8; 32]);
...
 
+/// MerkleNeighbor represents a step in a Merkle proof of inclusion.
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub enum MerkleNeighbor {
+    Left([u8; 32]),
+    Right([u8; 32]),
+}
+
+/// MerkleTree represents a node in the Merkle tree, containing either
+/// an uncomputed Entry or the precomputed hash.
```
oleganza: stale comment. MerkleTree now represents the actual merkle tree :-)

Issue #92, opened 2019-02-15
musig: implementation of aggregated N-of-N signatures with Ristretto and Merlin


oleganza:
For multi-signature accounts and multi-party contracts we need to implement a n-of-n MuSig protocol [1, 2].

The verification construction is very similar to the [aggregated transaction signature](https://github.com/interstellar/slingshot/blob/main/zkvm/spec/ZkVM.md#aggregated-signature) except the key aggregation happens offline — the verifier is exposed only to the single pubkey.

The proving construction is a 3-round MuSig procedure (see the paper [2]).

We'll also need to extend the ZkVM witness types and Prover API to handle the multi-party signing: we'd need to produce verifiable "signing instructions" that every co-signer can check before computing their part of the signature. It's not clear yet how that should be structured - this remains to be designed.

## FAQ

**What's the role of Merlin transcripts?**

MuSig can be specified more cleanly with transcripts rather than with nested hash constructions: e.g. instead of making a delinearization factor as `H(my key, H(all keys))` you'd just commit all pubkeys into a transcript, then squeeze all factors one by one from it. Also, no need to specify message hashing and padding - just have the user prepare the signature transcript however they please and then do the schnorr proof with it w/o worrying about the message.

**How do we do 2-of-3 and other thresholds?**

Via [predicate tree](https://github.com/interstellar/slingshot/blob/main/zkvm/spec/ZkVM.md#predicate-tree): happy path would be the most-expected 2-of-3, while other combinations tucked into deeper branches. This does not scale well to large number of participants, but works fine for small sizes.

**Why not a threshold scheme?**

Threshold signatures (m-of-n, m ≤ n) use a very similar cryptographic construction outlined in MuSig paper, yet have two important practical differences:

1. The secret keys have to be interactively created. N-of-N aggregation allows every party to generate or even reuse their existing keys (e.g. derived from a Keytree #87) w/o having to make a multi-round protocol to compute some synthetic keys and have them stored.
2. Threshold signatures are more compact than a merkle tree of combinations, but not _accountable_: you can never learn who were exact parties that signed off the particular message. For some protocols it's important to know whether "cold storage key" or "escrow agent" was involved or not. Combining n-of-n keys with the predicate tree allows seeing who actually signed, while threshold signatures erase this information. Note: accountability is available only to the parties in-the-know, for the rest of the network all the keys look random.

We'd eventually need threshold scheme too, especially for large thresholds or for cases where accountability is not necessary and we only need redundancy.

[...]

PR #190, opened 2019-03-05
zkvm/signature: implement MuSig protocol

cathieyun:
Implement musig, as described here: https://github.com/interstellar/slingshot/issues/92

First step: implement musig as laid out in the paper, without the mitigation of russell's attack. This will be a good stepping stone to figuring out the API and how the signer module should interface with the rest of ZkVM. 
Questions: 
- should the signer module be its own crate, or remain as a module under ZkVM? (Is it reusable / a standalone implementation? Or does it have a lot of ZkVM-specific integration?)
- do we have a better name for it than `signer`? (yay name bikeshedding!)
- for message communication: should we have a central dealer (a la multiparty rangeproof aggregation), or have parties communicate peer-to-peer? (oleg says starting off with a central dealer is easier)

Next steps detailed here: https://github.com/interstellar/slingshot/issues/92


zkvm/src/signer/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    agg_pubkey: PubKey,
+    // L = H(X_1 || X_2 || ... || X_n)
+    agg_pubkey_hash: PubKeyHash,
+    // TODO: what should the message representation format be?
+    message: Vec<u8>,
+}
+
+impl MultikeyWitness {
+    pub fn aggregate(&self, transcript: &mut Transcript) -> (PubKey, PubKeyHash) {
+        // L = H(X_1 || X_2 || ... || X_n)
```
oleganza: The nested hash construction is not necessary if we use STROBE. Think of "L" being the internal state of the sponge. What we get with transcript API is a way cleaner "commit all keys, squeeze all randomization factors", without extra intermediate stuff:

image

Cf. https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#aggregated-signature

zkvm/src/signer/prover.rs
```diff
@@ -0,0 +1,150 @@
...
+
+use crate::signer::*;
+use crate::transcript::TranscriptProtocol;
+use curve25519_dalek::ristretto::RistrettoPoint;
+use curve25519_dalek::scalar::Scalar;
+use rand;
+
+#[derive(Copy, Clone)]
+pub struct Nonce(Scalar);
+pub struct NoncePrecommitment(Scalar);
```
oleganza:
```suggestion
pub struct NoncePrecommitment([u8;32]);
```

we don't use it as a scalar, we only hash it to 32 pseudo-random bytes and compare them with what we get in the next step.

zkvm/src/signer/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+use rand;
+
+// Modules for signing protocol
+
+pub mod prover;
+
+pub struct PrivKey(Scalar);
+pub struct PubKey(RistrettoPoint);
+pub struct MultikeyWitness(Vec<PubKey>); // TODO: also include Option<Scalar> for signing key?
+pub struct PubKeyHash(Scalar);
```
oleganza: do we need this type throughout the protocol?

zkvm/src/signer/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+            let a_i = a_i_transcript.challenge_scalar(b"a_i");
+            X = X + a_i * X_i.0;
+        }
+
+        (PubKey(X), PubKeyHash(L))
+    }
+}
+
+impl Signature {
+    pub fn verify(&self, shared: Shared) -> bool {
```
oleganza: We can safely hardcode ristretto basepoint and pass &mut Transcript here. Then, the API would hardly be different from this: https://github.com/interstellar/slingshot/blob/main/zkvm/src/signature.rs#L28

That is, we also pass in compressed pubkey and R, so we don't need to re-compress them. And decompression is deferred to PointOp.

Note: the current `verify_aggregated` API in zkvm is the poor attempt at multi-message API - please ignore the part about multiple keys.

oleganza: >We can safely hardcode ristretto basepoint and pass &mut Transcript here.
I mean, no need for `Shared`.

zkvm/src/signer/mod.rs
```diff
@@ -0,0 +1,229 @@
...
+    // X_agg = sum_i ( a_i * X_i )
+    X_agg: PubKey,
+    // L = H(X_1 || X_2 || ... || X_n)
+    L: PubKeyHash,
+    // message being signed
+    m: Vec<u8>,
+}
+
+impl MultiKey {
+    pub fn aggregate(&self, transcript: &mut Transcript) -> (PubKey, PubKeyHash) {
```
oleganza: Aggregation is the "entry-point" w.r.t. the transcript (unlike signing which is composable with whatever prior ZKP operations) — so instead of passing it in, we need to just hardcode the label like `Transcript::new("ZkVM.aggregated-key")`.

cathieyun: should we return the trancript from this function then?

oleganza: nope. It's an "internal hash" instance only.

cathieyun: Hmm. the hardcoded label used in the `aggregate` function HAS to match the label in the signer states (since otherwise, they would get different `a_i`, `c_i` challenges). So, I'm guessing that means we should also hardcode a transcript state into `PartyAwaitingPrecommitments::new()` as well

zkvm/src/signer/mod.rs
cathieyun: I mentioned this in the other comment, but the transcript for the aggregation and signer and verifier all have to be the exact same, or else it won't verify properly.

So, if we are hardcoding the transcript into the aggregation, then we should hardcode the transcript into the verify function (instead of passing in &mut Transcript). Which should it be... all hardcoded, or all passed in?

zkvm/src/signer/prover.rs
cathieyun: I think it's good to change the hash output from being a scalar to a byte array, but the transcript `commit_bytes` function outputs `[u8;64]` so I will use that length of byte array instead.

zkvm/src/signer/mod.rs
cathieyun: I got rid of the `PubKeyHash` type, now that we just have this as a function call over `Multikey`.

zkvm/src/signer/mod.rs
cathieyun: This approach will work for the multikey aggregation; however, when each individual party wants to compute its own `a_i`, then they will have to either have to have all of the individual pubkeys (in order to pass in the pubkeys that are "before" them in ordering to the transcript), or they will have to receive their `a_i` from the coordinator (which seems awkward for coordination). Do you think it's worth the loss in the ergonomic-ness? If so, WDYT is the better approach?

zkvm/src/signer/prover.rs
cathieyun: actually, implemented as `[u8;32]` after all.

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub fn new(pubkeys: Vec<VerificationKey>) -> Result<Self, VMError> {
+        // Create transcript for Multikey
+        let mut transcript = Transcript::new(b"ZkVM.aggregated-key");
+
+        // Hash in pubkeys
+        // L = H(X_1 || X_2 || ... || X_n)
+        for X_i in &pubkeys {
+            transcript.commit_point(b"X_i.L", &X_i.0);
+        }
+        let L = transcript.challenge_scalar(b"L");
```
oleganza: Pre-hashed transcript is equivalent to "L", so we don't need this variable at all.

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        Ok(Multikey {
+            transcript,
+            X_agg: VerificationKey(X_agg.compress()),
+            L,
+        })
+    }
+
+    // Make a_i
+    pub fn a_i(&self, X_i: &VerificationKey) -> Scalar {
```
oleganza: is there a better name for this? Like, a `factor_for_key`?

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,64 @@
...
+        let L = transcript.challenge_scalar(b"L");
+
+        // Make aggregated pubkey
+        // X = sum_i ( a_i * X_i )
+        // a_i = H(L, X_i). Compenents of L have already been fed to transcript.
+        let mut X_agg = RistrettoPoint::default();
+        for X_i in &pubkeys {
+            let mut a_i_transcript = transcript.clone();
+            a_i_transcript.commit_point(b"X_i", &X_i.0);
+            let a_i = a_i_transcript.challenge_scalar(b"a_i");
```
oleganza: to avoid duplication we can use `self.a_i()` method (aka `self.factor_for_key`).

zkvm/src/signature/multikey.rs
cathieyun: ok, I removed the generation of "L" and also removed the "L" field.

zkvm/src/signature/multikey.rs
cathieyun: We don't have `self` yet, since this is happening within the constructor function.
So, should we make a "temporary, half-formed" `multikey` and then call `self.factor_for_key()` on that? It seems messy, but I'll try it out... let me know what you think?

zkvm/src/signature/multikey.rs
cathieyun: yes!

zkvm/src/signer/mod.rs
cathieyun: Resolved - we are hardcoding the transcript for pubkey aggregation, and passing in the transcript for nonce commitment generation, factor_for_key generation etc (basically, for prover and verifier use)

zkvm/src/signer/mod.rs
cathieyun: Resolved - we are hardcoding the transcript for pubkey aggregation, and passing in the transcript for nonce commitment generation, factor_for_key generation etc (basically, for prover and verifier use)

zkvm/src/signer/mod.rs
cathieyun: For now, I am just implementing this as detailed in the paper (with the exception of not explicitly generating and then committing to "L"... that is done implicitly by feeding the "X_i"s to the transcript"). So, each party feeds its own "X_i" into a clone of the transcript, and gets its own factor_for_key. I think that's the cleanest design, since then parties can get their own factor_for_key without having to know all the other pubkeys.

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,58 @@
...
+use merlin::Transcript;
+
+#[derive(Clone)]
+pub struct Multikey {
+    transcript: Transcript,
+    X_agg: VerificationKey,
+}
+
+impl Multikey {
+    pub fn new(pubkeys: Vec<VerificationKey>) -> Result<Self, VMError> {
```
oleganza: maybe use "lightweight" Option instead of Result here? We do not have a lot of errors to throw via `?` shortcut.

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,58 @@
...
+        // Create transcript for Multikey
+        let mut transcript = Transcript::new(b"ZkVM.aggregated-key");
+
+        // Hash in pubkeys
+        // L = H(X_1 || X_2 || ... || X_n)
+        for X_i in &pubkeys {
+            transcript.commit_point(b"X_i.L", &X_i.0);
+        }
+
+        let mut multikey = Multikey {
```
oleganza: would it be simpler to create the Multikey closer to the end, so you don't need to have dummy `X_agg`.

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,58 @@
...
+use crate::signature::VerificationKey;
+use crate::transcript::TranscriptProtocol;
+use curve25519_dalek::ristretto::RistrettoPoint;
+use curve25519_dalek::scalar::Scalar;
+use merlin::Transcript;
+
+#[derive(Clone)]
+pub struct Multikey {
+    transcript: Transcript,
+    X_agg: VerificationKey,
```
oleganza: how about?
```suggestion
    aggregated_key: VerificationKey,
```

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl Multikey {
+    pub fn new(pubkeys: Vec<VerificationKey>) -> Result<Self, VMError> {
+        // Create transcript for Multikey
+        let mut transcript = Transcript::new(b"ZkVM.aggregated-key");
+
+        // Hash in pubkeys
+        // L = H(X_1 || X_2 || ... || X_n)
```
oleganza: for disambiguation we need a length prefix before committing any variable-length lists: `commit_u64("n", pubkeys.len())`

zkvm/src/signature/multikey.rs
cathieyun: See discussion here:
https://github.com/interstellar/slingshot/pull/190#discussion_r264976476
I am trying not to deduplicate the `factor_for_key` code, which is why I am creating this half-formed multikey (so I can call the `factor_for_key` function on it

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,58 @@
...
+        for X_i in &pubkeys {
+            transcript.commit_point(b"X_i.L", &X_i.0);
```
oleganza: is there a better label than jargon from paper? E.g. `"P"` for "pubkey"? Even `"X_i"` is better than `"X_i.L"` since L is the transcript state itself.

zkvm/src/signature/musig.rs
```diff
@@ -0,0 +1,175 @@
...
+        transcript: &Transcript,
+        X_agg: VerificationKey,
+    ) -> Result<(), VMError> {
+        let G = RISTRETTO_BASEPOINT_POINT;
+        let mut transcript = transcript.clone();
+
+        // Make c = H(X_agg, R, m)
+        // The message `m` should already have been fed into the transcript
+        let c = {
+            transcript.commit_point(b"X_agg", &X_agg.0);
```
oleganza: the verifier should not be exposed to the fact that the key is aggregated or not. So for the verifier, it's just "a pubkey"

```suggestion
            transcript.commit_point(b"P", &P.0);
```

zkvm/src/signature/musig.rs
```diff
@@ -0,0 +1,175 @@
...
+
+#[derive(Clone)]
+pub struct Message(pub Vec<u8>);
+
+impl Signature {
+    pub fn verify(
+        &self,
+        // The message `m` should already have been fed into the transcript
+        transcript: &Transcript,
+        X_agg: VerificationKey,
```
oleganza:
```suggestion
        P: VerificationKey,
```

see comment below

zkvm/src/signature/musig.rs
```diff
@@ -0,0 +1,175 @@
...
+
+        // Make c = H(X_agg, R, m)
+        // The message `m` should already have been fed into the transcript
+        let c = {
+            transcript.commit_point(b"X_agg", &X_agg.0);
+            transcript.commit_point(b"R", &self.R.compress());
+            transcript.challenge_scalar(b"c")
+        };
+
+        let X_agg = match X_agg.0.decompress() {
```
oleganza: i wonder if it makes sense to revamp `VerificationKey` to be decompressed to make usage non-failable. We can keep the compressed copy inside too, so we don't need to recompress when hashing into transcript?.. we can get to this in a separate PR, tho.

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl PartyAwaitingPrecommitments {
+    pub fn new(
+        // The message `m` should already have been fed into the transcript
+        transcript: &Transcript,
+        x_i: PrivKey,
+        multikey: Multikey,
+    ) -> (Self, NoncePrecommitment) {
+        let transcript = transcript.clone();
+        let mut rng = transcript.build_rng().finalize(&mut rand::thread_rng());
```
oleganza: we should also commit `x_i` to the transcriptRNG in order to rely not only on the system RNG's entropy.

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,191 @@
...
+        ))
+    }
+}
+
+impl PartyAwaitingSiglets {
+    pub fn receive_siglets(self, siglets: Vec<Siglet>) -> Signature {
+        // s = sum(siglets)
+        let s: Scalar = siglets.iter().map(|siglet| siglet.0).sum();
+        // R = sum(R_i). nonce_commitments = R_i
+        let R: RistrettoPoint = self.nonce_commitments.iter().map(|R_i| R_i.0).sum();
```
oleganza: we already compute it when doing F-S challenge, so we can pass it to the next state w/o recomputing it again.

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,191 @@
...
+                nonce_commitments,
+                c,
+            },
+            Siglet(s_i),
+        ))
+    }
+}
+
+impl PartyAwaitingSiglets {
+    pub fn receive_siglets(self, siglets: Vec<Siglet>) -> Signature {
```
oleganza: maybe `receive_trusted_siglets` name for method that does not verify siglets, and `receive_siglets` for method that does (cf. range proof API: https://github.com/dalek-cryptography/bulletproofs/blob/main/src/range_proof/dealer.rs#L328)?

Also: would it be more clear or less if we rename "siglets" into "shares" like with rangeproof API?

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,191 @@
...
+            let X_i = pubkeys[i].0;
+            let R_i = self.nonce_commitments[i].0;
+
+            // Make a_i = H(L, X_i)
+            let a_i = self
+                .multikey
+                .factor_for_key(&VerificationKey(X_i.compress()));
+
+            // Check that S_i = R_i + c * a_i * X_i
+            assert_eq!(S_i, R_i + self.c * a_i * X_i);
```
oleganza: let's fail with Result<> instead of panic?

oleganza: the error should also return the index of the party that failed

zkvm/src/signature/multikey.rs
cathieyun: 👍

zkvm/src/signature/multikey.rs
cathieyun: 👍

zkvm/src/signature/multikey.rs
cathieyun: ok. I originally called it `X_i.L` to differentiate it from the `X_i` that goes into the transcript for `factor_for_key`. I can change it to `P`!

zkvm/src/signature/musig.rs
cathieyun: sgtm

zkvm/src/signature/prover.rs
cathieyun: that makes sense!

zkvm/src/signature/musig.rs
```diff
@@ -0,0 +1,175 @@
...
+#[derive(Clone)]
+pub struct Message(pub Vec<u8>);
+
+impl Signature {
+    pub fn verify(
+        &self,
+        // The message `m` should already have been fed into the transcript
+        transcript: &Transcript,
+        X_agg: VerificationKey,
+    ) -> Result<(), VMError> {
```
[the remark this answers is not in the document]
cathieyun: yeah, in the next PR I'll integrate it into the ZkVM API which includes returning this as PointOp.

zkvm/src/signature/musig.rs
cathieyun: If `VerificationKey` is being passed around between parties, then it makes sense to be compressed... but for within-ZkVM use, it would be easier for it to be decompressed. I don't have a good sense for the communication requirements for `VerificationKey`s. Maybe a separate PR?

zkvm/src/signature/prover.rs
cathieyun: I don't understand... pass what to the next state? The `PartyAwaitingSiglets` is the last state, so it has to create a `Signature` (unless we want to get rid of the `PartyAwaitingSiglets` state and just have the aggregator add up the `s_i` and `R_i` components to make `Signature` themselves?

zkvm/src/signature/prover.rs
cathieyun: siglets technically aren't "signature shares", because they don't contain the `R_i` components... so it's not quite the same as the rangeproof API. I don't have a strong opinion on calling them "siglets" or "shares".

I renamed the siglet receiving functions, that makes sense!

zkvm/src/signature/prover.rs
cathieyun: sounds good!

zkvm/src/signer/mod.rs
oleganza: Yeah, seems like cloning transcript for each factor is the way to go to avoid creating all the factors and picking only yours.

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        // Hash in pubkeys
+        // L = H(X_1 || X_2 || ... || X_n)
+        for X_i in &pubkeys {
+            transcript.commit_point(b"P", &X_i.0);
+        }
+
+        let mut multikey = Multikey {
+            transcript,
+            aggregated_key: VerificationKey(RistrettoPoint::default().compress()),
```
oleganza: there's `curve25519_dalek::constants::RISTRETTO_BASEPOINT_COMPRESSED` to avoid compression cost.

zkvm/src/signature/multikey.rs
oleganza: `factor_for_key` can internally use factored-out private type-level function `Multikey::compute_factor(&mut Transcript, &VerificationKey)` which we can reuse here w/o a dummy instance.

zkvm/src/signature/musig.rs
```diff
@@ -0,0 +1,176 @@
...
+pub struct PrivKey(pub Scalar);
+
+#[derive(Debug, Clone)]
+pub struct Signature {
+    pub s: Scalar,
+    pub R: RistrettoPoint,
+}
+
+#[derive(Clone)]
+pub struct Message(pub Vec<u8>);
```
oleganza: do we need this type?

zkvm/src/signature/musig.rs
```diff
@@ -0,0 +1,176 @@
...
+use crate::errors::VMError;
+use crate::signature::VerificationKey;
+use crate::transcript::TranscriptProtocol;
+use curve25519_dalek::constants::RISTRETTO_BASEPOINT_POINT;
+use curve25519_dalek::ristretto::RistrettoPoint;
+use curve25519_dalek::scalar::Scalar;
+use merlin::Transcript;
+
+#[derive(Clone)]
+pub struct PubKey(pub RistrettoPoint);
```
oleganza: maybe use one pubkey type? `VerificationKey` can work, and if we need to make it decompressed, we can make it a struct with private fields for `Option<compressed>` and decompressed point. Compressed is to avoid re-compression cost if instantiating from a compressed point.

zkvm/src/signature/musig.rs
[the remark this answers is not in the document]
oleganza: sounds good

zkvm/src/signature/musig.rs
```diff
@@ -0,0 +1,176 @@
...
+
+impl Signature {
+    pub fn verify(
+        &self,
+        // The message `m` should already have been fed into the transcript
+        transcript: &Transcript,
+        P: VerificationKey,
+    ) -> Result<(), VMError> {
+        let G = RISTRETTO_BASEPOINT_POINT;
+        let mut transcript = transcript.clone();
```
oleganza: i guess this clone is to match the clone inside the prover? Yeah, we definitely need to revisit it in a follow-up PR.

zkvm/src/signature/musig.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        // Make c = H(X, R, m)
+        // The message `m` should already have been fed into the transcript
+        let c = {
+            transcript.commit_point(b"P", &P.0);
+            transcript.commit_point(b"R", &self.R.compress());
+            transcript.challenge_scalar(b"c")
+        };
+
+        let P = match P.0.decompress() {
```
oleganza: suggestion: `let P = decompress().ok_or(VMError::InvalidPoint)?;`

zkvm/src/signature/musig.rs
```diff
@@ -0,0 +1,176 @@
...
+            Some(P) => P,
+            None => return Err(VMError::InvalidPoint),
+        };
+        // Check sG = R + c * aggregated_key
+        match self.s * G == self.R + c * P {
```
oleganza: why not

```rust
if self.s * G == self.R + c * P {
    Ok(())
} else {
    Err(VMError::...)
}
```

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,195 @@
...
+use crate::transcript::TranscriptProtocol;
+use curve25519_dalek::constants::RISTRETTO_BASEPOINT_POINT;
+use curve25519_dalek::ristretto::RistrettoPoint;
+use curve25519_dalek::scalar::Scalar;
+use merlin::Transcript;
+use rand;
+use subtle::ConstantTimeEq;
+
+#[derive(Copy, Clone, Debug)]
+pub struct Nonce(Scalar);
```
oleganza: i'm not sure we need this type - the nonce is always internal to the signer, so we can simply use Scalar. Same for Privkey type.

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,195 @@
...
+
+impl PartyAwaitingPrecommitments {
+    pub fn new(
+        // The message `m` should already have been fed into the transcript
+        transcript: &Transcript,
+        x_i: PrivKey,
+        multikey: Multikey,
+    ) -> (Self, NoncePrecommitment) {
+        let mut rng_transcript = transcript.clone();
+        rng_transcript.commit_scalar(b"x_i", &x_i.0);
```
oleganza: `Transcript::commit_scalar` is for public data.

For mixing in entropy there's `TranscriptRngBuilder::commit_witness_bytes`:

```rust
let mut rng = transcript
    .build_rng()
    .commit_witness_bytes(b"witness1", witness_data)
    .commit_witness_bytes(b"witness2", more_witness_data)
    .finalize(&mut rand::thread_rng());
```

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,195 @@
...
+        nonce_precommitments: Vec<NoncePrecommitment>,
+    ) -> (PartyAwaitingCommitments, NonceCommitment) {
+        // Store received nonce precommitments in next state
+        (
+            PartyAwaitingCommitments {
+                transcript: self.transcript,
+                multikey: self.multikey,
+                x_i: self.x_i,
+                r_i: self.r_i,
+                nonce_precommitments,
```
oleganza: if you'd like you can do `PartyAwaitingCommitments { nonce_precommitments, .. self }` to fill in the rest of the fields from `self`.

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,195 @@
...
+                r_i,
+                R_i,
+            },
+            precommitment,
+        )
+    }
+
+    pub fn receive_precommitments(
+        self,
+        nonce_precommitments: Vec<NoncePrecommitment>,
```
oleganza: do we need to receive our own precommitment at a correct index here?

zkvm/src/signature/prover.rs
oleganza: sorry, i meant R is summed up first inside `PartyAwaitingCommitments::receive_commitments`, so we can pass it into `PartyAwaitingSiglets` and avoid recomputing it here.

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,195 @@
...
+            // Make a_i = H(L, X_i)
+            let a_i = self
+                .multikey
+                .factor_for_key(&VerificationKey(X_i.compress()));
+
+            // Check that S_i = R_i + c * a_i * X_i
+            if S_i != R_i + self.c * a_i * X_i {
+                return Err(VMError::SignatureShareError { index: i });
+            }
+            assert_eq!(S_i, R_i + self.c * a_i * X_i);
```
oleganza: leftover assert?

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,195 @@
...
+            },
+            self.R_i,
+        )
+    }
+}
+
+impl PartyAwaitingCommitments {
+    pub fn receive_commitments(
+        mut self,
+        nonce_commitments: Vec<NonceCommitment>,
```
oleganza: Not for this PR, but i wonder if there's a more clean way to pass around and collect every party's (pre)commitment? Maybe we need some thin Dealer implementation to orchestrate this?

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,195 @@
...
+pub struct Nonce(Scalar);
+
+#[derive(Copy, Clone)]
+pub struct NoncePrecommitment([u8; 32]);
+
+#[derive(Copy, Clone, Debug)]
+pub struct NonceCommitment(RistrettoPoint);
+
+#[derive(Copy, Clone, Debug)]
+pub struct Siglet(Scalar);
```
oleganza: maybe we things will still look nice w/o this extra type wrapper? Just trying to minimize amount of entities...

zkvm/src/signature/musig.rs
cathieyun: yup exactly... if the verifier didn't clone the transcript (if it received a mutable borrow of the transcript), then the transcript states would differ for the prover & verifier at the end of the protocol, which would be bad for future parts.

zkvm/src/signature/musig.rs
cathieyun: removed - it looks fine. I wasn't sure how much abstraction / wrapping we wanted so I originally leaned towards more wrappers.

zkvm/src/signature/prover.rs
cathieyun: we do not necessarily have to receive our own precommitment a a "correct" index, but we do have to receive it at the same index that we receive our commitment at, in the next step. 
(we could also assume that the dealer sends all of the precommitments and commitments _except_ for those already owned by the party, but it just requires more sorting on the dealer's part so I started with the most simple case first.)

zkvm/src/signature/prover.rs
cathieyun: I considered having a dealer implementation, but each state change would just be a one-line map function (with the current implementation, such as in the tests), since the dealer doesn't actually have to hold on to any state... so I thought the logic was too simple to warrant making a dealer for. But if it makes the API easier to use, then it could be good!

zkvm/src/signature/prover.rs
cathieyun: Sure! I erred toward more type wrappers for clarity's sake, but can remove the unnecessary ones now!

zkvm/src/signature/prover.rs
cathieyun: removed!

zkvm/src/signature/prover.rs
cathieyun: whoops, removed!

zkvm/src/signature/musig.rs
cathieyun: done, that does look better!

zkvm/src/signature/musig.rs
cathieyun: nice!

zkvm/src/signature/multikey.rs
cathieyun: nice!

zkvm/src/signature/prover.rs
cathieyun: ah that is much nicer, then I don't have to make a mutable clone of the transcript! thanks

zkvm/src/signature/prover.rs
cathieyun: I think this only works if `self` is of the same type as the struct you're filling in? 

I tried this, and I got the error "mismatched types... expected struct `PartyAwaitingCommitments`, found type `PartyAwaitingPrecommitments`.

I tried to look up documentation on filling the rest of the fields using `.. self` but couldn't find any, I'm curious where to find more information on using this trick?

zkvm/src/signature/prover.rs
cathieyun: Maybe I'll get a better idea of what the API should look like in the next MuSig task, which is integrating this signing scheme into ZkVM!

zkvm/src/signature/prover.rs
oleganza: Yeah, maybe it only works with the same type, oops.
https://doc.rust-lang.org/1.6.0/book/structs.html#update-syntax

image

zkvm/src/signature/prover.rs
oleganza: We have siglet validation in `receive_siglets ` - there siglets and nonces all must have consistent indices.

zkvm/src/signature/musig.rs
cathieyun: Done! I will use `VerificationKey` instead of `PubKey` for all instances. For now, it seems fine to pass it around as a `CompressedRistretto` wrapper, can update the struct if it gets to clumsy to wrap and rewrap.

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        multikey.aggregated_key = VerificationKey(aggregated_key.compress());
+
+        Some(multikey)
+    }
+
+    // Make a_i
+    pub fn factor_for_key(&self, X_i: &VerificationKey) -> Scalar {
+        // a_i = H(L, X_i). Compenents of L have already been fed to transcript.
```
oleganza:
```suggestion
        // a_i = H(L, X_i). Components of L have already been fed to transcript.
```

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,189 @@
...
+pub struct PartyAwaitingSiglets {
+    multikey: Multikey,
+    c: Scalar,
+    nonce_commitments: Vec<NonceCommitment>,
+}
+
+impl NonceCommitment {
+    fn precommit(&self) -> NoncePrecommitment {
+        let mut h = Transcript::new(b"MuSig.nonce-precommit");
+        h.commit_point(b"R_i", &self.0.compress());
```
oleganza:
```suggestion
        h.commit_point(b"R", &self.0.compress());
```

I generally want to simplify the labels and naming so we don't cargocult quite arbitrary notation from the paper.

oleganza: Especially since the paper can be rewritten, but labels must stay forever for compatibility.

zkvm/src/signature/multikey.rs
cathieyun: thanks!

zkvm/src/signature/prover.rs
cathieyun: In my mind, the thing we are committing is the "per-party" `R` (and hence, called `R_i`)... whereas the "aggregated" `R` is the one that should really be called `R`.

Anyway, not a big deal & I changed it.

zkvm/src/signature/prover.rs
cathieyun: Ahh, that totally makes sense. Updated!

zkvm/src/signature/prover.rs
cathieyun: Oh, now that I have gotten rid of the `Siglet` struct, there is less context on what "siglet" means. So, calling the functions `...shares` instead of `...siglets` makes a lot more sense now. Updated the naming!

zkvm/src/signature/multikey.rs
cathieyun: oh whoops, I didn't think of that. Thanks for the suggestion!

(Also, I think we can pass in `&Transcript` instead of `&mut Transcript` to `compute_factor`, since it is only cloning the transcript & modifying the clone, not modifying the original.

zkvm/src/signature/multikey.rs
cathieyun: Oleg suggested a solution here,
https://github.com/interstellar/slingshot/pull/190#discussion_r265362498
and I implemented it! resolved!

zkvm/src/signature/prover.rs
cathieyun: oh interesting!
yeah that makes sense that it would only work with the same type, because then the compiler could know what fields are supposed to correspond with each other, I think? Anyway, I learned something new!

zkvm/src/signature/prover.rs
cathieyun: Yup - basically, the precommitments, commitments, and siglets all need to be received in the same per-party order, or else the protocol will fail at some step. I don't think it matters what order it is, as long as it is consistent for all messages.

zkvm/src/signature/multikey.rs
```diff
@@ -0,0 +1,56 @@
...
+
+impl Multikey {
+    pub fn new(pubkeys: Vec<VerificationKey>) -> Option<Self> {
+        // Create transcript for Multikey
+        let mut transcript = Transcript::new(b"ZkVM.aggregated-key");
+        transcript.commit_u64(b"n", pubkeys.len() as u64);
+
+        // Hash in pubkeys
+        // L = H(X_1 || X_2 || ... || X_n)
+        for X_i in &pubkeys {
```
oleganza: Super nitpicky, but in this code it does not add any value to use a subscript (there are no other subscripts or non-subscript variables anyway). I'd just say `X` and `a` instead of `X_i` and `a_i`. The less non-informative pieces we have, the easiest it is to grasp the logic.

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,187 @@
...
+        let mut h = Transcript::new(b"MuSig.nonce-precommit");
+        h.commit_point(b"R", &self.0.compress());
+        let mut precommitment = [0u8; 32];
+        h.challenge_bytes(b"precommitment", &mut precommitment);
+        NoncePrecommitment(precommitment)
+    }
+}
+
+impl PartyAwaitingPrecommitments {
+    pub fn new(
```
oleganza: maybe make the entry point nicer via `struct Party{}`: `Party::new() -> (PartyAwaitingPrecommitments, NoncePrecommitment)`?

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,187 @@
...
+            .iter()
+            .zip(nonce_commitments.iter())
+        {
+            // Make H(comm) = H(R_i)
+            let actual_precomm = comm.precommit();
+
+            // Compare H(comm) with pre_comm, they should be equal
+            let equal = pre_comm.0.ct_eq(&actual_precomm.0);
+            if equal.unwrap_u8() == 0 {
+                return Err(VMError::InconsistentWitness);
```
oleganza: would be good to signal which party is at fault, so we can decrease their karma

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,187 @@
...
+                nonce_commitments,
+                c,
+                R,
+            },
+            s_i,
+        ))
+    }
+}
+
+impl PartyAwaitingSiglets {
```
oleganza: if we call them "shares", lets remove the rest of "siglets" from type and method names?

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,187 @@
...
+    pub fn receive_trusted_shares(self, shares: Vec<Scalar>) -> Signature {
+        // s = sum(s_i), s_i = shares[i]
+        let s: Scalar = shares.iter().map(|share| share).sum();
```
oleganza: maybe moving things in (since we don't reuse them later) will be more efficient?
```suggestion
        let s: Scalar = shares.into_iter().map(|share| share).sum();
```

zkvm/src/signature/multikey.rs
cathieyun: ok, that makes sense. I wanted to make it match the paper as closely as possible, but I can see how the code is simpler by getting rid of these details.

zkvm/src/signature/prover.rs
```diff
@@ -0,0 +1,187 @@
...
+        Signature { s, R: self.R }
+    }
+
+    pub fn receive_shares(
+        self,
+        shares: Vec<Scalar>,
+        pubkeys: Vec<VerificationKey>,
+    ) -> Result<Signature, VMError> {
+        // Check that all shares are valid
+        for (i, s_i) in shares.iter().enumerate() {
```
oleganza: i wonder if we need to represent shares as a state machine themselves:

```
SharePrecommitted -> H(R_i)
ShareCommitted -> R_i (if verified to have correct R_i)
ShareSigned -> (R_i,s_i) (if verified to have correct s_i)
```

oleganza: then we can store identifiers inside shares and match them up correctly, or at least, factor verification logic out into more localized pieces, so we don't have big loops over multiple arrays.

zkvm/src/signature/prover.rs
cathieyun: done!

zkvm/src/signature/prover.rs
cathieyun: Why would this be more efficient... because the function would have ownership of the shares?

I made the change!

zkvm/src/signature/prover.rs
oleganza: the error must be musig-specific, like `MuSigError::InvalidNonceCommitment(usize)`, similarly for share validation.

zkvm/src/signature/prover.rs
oleganza: we can do this in another PR as a follow-up, since this one seems to be correct and working

zkvm/src/signature/prover.rs
cathieyun: good call!

zkvm/src/signature/prover.rs
cathieyun: ok!

zkvm/src/signature/prover.rs
cathieyun: So, curently we have:
```
NoncePrecommitment -> H(R_i)
NonceCommitment -> R_i (if verified to have correct R_i)
```
and we also used to have `Siglet -> R_i` (but you suggested removing it).

These structs seems to be the same thing as what you suggested above (slightly different names). So what your'e basically suggesting is that we store indexes in each of these structs as well, so we can match up each party with their index with their messages? (Then, would we also want to assign an index to each party?)

oleganza: Except the idea is to focus on the _state of the counter-party_ and organize necessary checks around it, not on individual components - nonces, precommitments, siglets etc. Those are just pieces of the "counterparty's state".

I think my suggested `Share*` naming is off and it needs to be more like `CounterpartyPrecommitted`, `CounterpartyCommitted`, `CounterpartySigned`. We can later see what's more convenient/efficient - treating oneself as counterparty too and repeat all messages and checks, or only track truly others' state.

oleganza: Maybe we need to start with another state: `Counterparty`, which holds compressed pubkey `X_i` which we can use as an identifier, and no longer worry about indices.

cathieyun: I think I understand what you mean, but just to gain clarity on details, can we hop on a call to talk about this design today (after stand-up)?

oleganza: sure! also: if you prefer, we can merge this as is and iterate in smaller PRs on top.

oleganza: the reason we have state machine (`Counterparty*`) within a state machine (`Party*`) is because each party is simultaneously a "dealer". So in RP MPC we had a separate dealer, so each each party was worrying only about its own state, not the others'.

zkvm/src/signature/signature.md
```diff
@@ -0,0 +1,44 @@
...
+	clone / fork transcript to make per-key factors
+2. signing & verifying
+	each party has state, wants to go to next state by verifying & performing computations. so you can "encode validity of something in the type".
+	party transitions need to be non-replayable (is this true?)
+
+dealer & parties in rangeproof:
+- dealer accumulates all factors, holds onto the transcript, gives challenges
+- parties don't have to know about other parties, who's failing, transcript, etc.
+
+oleg says: shouldn't have dealer
```
oleganza: Yeah, because each party has to validate what they sign, and each other's contributions before they proceed. In RP MPC it is not that important because each party proves their own statement and dealer only stitches those together, - parties individually are not critically affected by misbehavior of others. E.g. cannot be fooled to leak the key or accidentally prove something else ("replay a signature"). Also in RP MPC we explicitly do not have a goal of verifying transcript, so parties are simply proving the RP "contextless", that is shifted to the outer protocol, which possibly would have to replay dealer's state like we would replay counterparties' states.

zkvm/src/signature/signature.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+create 3 counterparty structs,
```
oleganza: how about this format:
```
Counterparty{pubkey}
  ↓
.precommit_nonce(H) // simply adds precommitment
  ↓
CounterpartyPrecommitted{H, pubkey}
  ↓
.commit_nonce(R) // verifies hash(R) == H
  ↓
CounterpartyCommitted{R, pubkey}
/* R_total := sum{R_i}, commit to transcript, compute and send out siglet */
  ↓
.sign(s) // verifies s_i*G == R_i + c * factor * pubkey_i
  ↓
  s

/* s_total = sum{s_i} */
```

zkvm/src/signature/signature.md
```diff
@@ -0,0 +1,253 @@
...
+### Key Aggregation
+
+Key aggregation happens in the `Multikey::new(...)` function. 
+
+Input:
+- pubkeys: `Vec<VerificationKey>`. This is a list of compressed public keys that will be aggregated, as long as they can be decompressed successfully.
+
+Operation:
+- Create a new transcript using the tag "ZkVM.aggregated-key". (TODO: remove the "ZkVM." if we make the signature crate separate from the ZkVM crate.)
+- Commit all the pubkeys to the transcript. This reflects the creation of the `<L>` encoding of the list of pubkeys detailed in the MuSig paper: `<L> = H(X_1 || X_2 || ... || X_n)`. The difference is that we do not need to squeeze `<L>` out of the transcript, since we can use the prepared transcript directly for future randomness.
```
oleganza: If we read the `L = H(X_1, ..., X_n)` as "commitment to all the keys", then transcript state implements exactly this, so there is no conceptual difference — the only difference is superficial — in the syntax. Because when you permute the sponge, it's _exactly just like calling a hash function_, just with a bigger output than people are used to.

```suggestion
- Commit all the pubkeys to the transcript. The transcript state corresponds to the commitment `<L>` in the MuSig paper: `<L> = H(X_1 || X_2 || ... || X_n)`.
```

zkvm/src/signature/signature.md
```diff
@@ -0,0 +1,253 @@
...
+- a new multikey, with the transcript and aggregated key detailed above.
+
+### Signing
+
+There are several paths to signing:
+1. Make a Schnorr signature with one public key (derived from one private key).
+    Function: `Signature::sign_single(...)`
+
+    Input: 
+    - transcript: `&Transcript` - the message to be signed should have been committed to the transcript beforehand. (TODO: pass in a mutable borrow of a transcript instead of a borrow.)
```
oleganza: let's write the doc as we need this to work, and figure out borrowck issues in the implementation?
```suggestion
    - transcript: `&mut Transcript` - the message to be signed should have been committed to the transcript beforehand.
```

zkvm/src/signature/signature.md
```diff
@@ -0,0 +1,253 @@
...
+    - privkey: `Scalar`
+
+    Operation:
+    - Use the transcript to generate a random factor (the nonce, `r`), by committing to the privkey and passing in a `thread_rng`
```
oleganza: we need to clone transcript before mixing it with privkey and RNG.

```suggestion
    - Clone the transcript state, mix it with the privkey and system-provided RNG to generate the nonce `r`. This makes the nonce uniquely bound to a message and private key, and also makes it non-deterministic to prevent "rowhammer" attacks.
```

zkvm/src/signature/signature.md
```diff
@@ -0,0 +1,253 @@
...
+    - Use the nonce to create a nonce commitment `R = r * G`
+    - Mutably clone the transcript
```
oleganza: We should update the user-provided transcript here.
```suggestion
```

zkvm/src/signature/signature.md
```diff
@@ -0,0 +1,253 @@
...
+2. Make a Schnorr signature with one aggregated key (`Multikey`), derived from multiple public keys.
+    TODO (tie to protocol for party & counterparty state transitions)
+
+### Verifying
+
+Signature verificaiton happens in the `Signature::verify(...)` function.
+
+Input: 
+- `&self`
+- transcript: `&Transcript` - the message to be signed should have been committed to the transcript beforehand. (TODO: pass in a mutable borrow of a transcript instead of a borrow.)
```
oleganza:
```suggestion
- transcript: `&mut Transcript` - the message to be signed should have been committed to the transcript beforehand.
```

zkvm/src/signature/signature.md
```diff
@@ -0,0 +1,253 @@
...
+- P: `VerificationKey`
+
+Operation:
+- Mutably clone the trancript
```
oleganza:
```suggestion
```

zkvm/src/signature/signature.md
```diff
@@ -0,0 +1,253 @@
...
+Also, for now we will assume that all of the messages passed into each party state arrive in the same order (each party's message is in the same index). This allows us to skip the step of ordering them / assigning indexes. (Future work: allow for unordered inputs, have the parties sort them.)
+
+### Party
+
+Fields: none
+
+Function: `new(...)`
+
+Input: 
+- transcript: `&Transcript` - the message to be signed should have been committed to the transcript beforehand. (TODO: pass in a mutable borrow of a transcript instead of a borrow.)
```
oleganza:
```suggestion
- transcript: `&mut Transcript` - the message to be signed should have been committed to the transcript beforehand.
```

zkvm/src/signature/signature.md
cathieyun: Yes that does sound way better, thanks!

zkvm/src/signature/signature.md
cathieyun: incorporated into the writeup, thanks!

zkvm/src/signature/prover.rs
cathieyun: Incorporated into the writeup!

oleganza:
Follow up things:

1. Separate error type, or more explicit naming for `ShareError` -> `MuSigShareError`.
2. `&mut Transcript`.
3. Counterparty state machines within each Party.

PR #191, opened 2019-03-05
keytree: add extension trait to Merlin

tessr:
Copy-pasted from zkvm.


tessr:
ok, I'll remove it for now!

PR #193, opened 2019-03-05
zkvm: introduce anchors

### LINKED ISSUE #172: zkvm: make contracts unique
PR: #193 

Currently `delegate` instruction signs only a program and a predicate, but not the contract's contents. So if there are two contracts with the same predicate, delegated program signed for one can be reused for another.

Half-way solution to this is to serialize the contract too. Due to (normally) unique blinding factors making the value commitments unique, this prevents reuse for most cases. However, there are still possible corner cases when values are repeated and/or non-blinded.

Full-way solution is to preserve some unique "anchor" (see also [TxVM anchors](https://github.com/chain/txvm/blob/main/specifications/txvm.md#values)) to a contract. However, unlike TxVM, we need this anchor attached to a _contract_, not the _values_.

How do we do that?

1. Each contract that is instantiated from a UTXO already has an anchor - the UTXO ID.
2. Each contract created _during tx execution_ needs to sample a unique value. For that we can have an extra running Transcript instance that's instantiated on the first `input`/`nonce` use and seeded with the corresponding utxo id or nonce. Subsequent inputs/nonces similarly seed the transcript. And transiently created contracts squeeze the unique anchors from it.

Why don't we squeeze anchors for claimed utxos too? Because then the signing is more dependent on all the preceding steps, making it impossible to sign an individual contract for use with `delegate` _prior to forming the transaction_ — that's an important feature for payment channels, for instance.

We can replace `.unique` flag with this `Option<Transcript>` instance.

oleganza:
This introduces **anchors** and **contract IDs**, bringing back good ideas from TxVM, but doing it slightly differently. 

1. Each contract has a globally unique identifier, **contract ID**, which allows `delegate` to sign the unique instance in presence of reused predicate keys (see #172).
2. Uniqueness of contracts is provided by unique **anchors** contained within the contracts.
3. Nonce-based contracts have their anchor computed out of the nonce tuple `(predicate, blockid, exptime)`. Blockchain state machine tracks these nonce anchors to ensure their uniqueness.
4. UTXO ID is simply a contract ID.
5. Newly created contracts and outputs **consume VM's last anchor** and **set their ID as a new last anchor**.
6. Instead of a "uniqueness" flag, VM simply keeps and updates `Option<Anchor>`.
7. There is no longer "input structure", only "output structure", since the output is fully self-contained and unique.
8. `Input` type is renamed to `Output`: its purpose is to simply store contract ID alongside the contract w/o computing it by serializing the contract.

Different from TxVM:

1. Anchors are stored _per contract_, not _per value_. Values remain stateless and consisting purely of commitments to quantity and flavor.
2. All contracts are unique. In TxVM if the value was not changed, UTXO could remain unchanged, or even flip-flop between reused identifiers.
3. All-zero blockid is not allowed: in TxVM it was a special case to allow hypothetical cross-chain swaps with replayable transactions, which is not safe (also) because derived anchors cannot be guaranteed to be unique.

Closes #172.

### DIFF
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -1479,7 +1479,7 @@ _pred blockid_ **nonce** → _contract_
     T.commit("blockid", blockid)
     T.commit("predicate", pred)
     T.commit("maxtime", LE64(tx.maxtime))
-    nonce_anchor = T.challenge_bytes()
+    nonce_anchor = T.challenge_bytes("anchor")
     ```
 4. Pushes a new [contract](#contract-type) with:
     * an empty [payload](#contract-payload)
@@ -1504,7 +1504,7 @@ Blockchain state machine performs the following checks:
 1. The `blockid` is checked against the IDs of “recent” blocks when the transaction is applied to the blockchain. It must match a recent block, or the initial block of the blockchain.
 2. The `nonce_anchor` is checked for uniqueness against “recent” nonce anchors.
 
-To perform these checks, validators must keep a set of recent nonces and a set of recent block headers available. For scalability and to reduce resource demands on the network, these sets must be limited in size. So block signers can and should impose reasonable limits on the value of exp (which is the time by which the transaction must be included in a block or become invalid).
+To perform these checks, validators must keep a set of recent nonces and a set of recent block headers available. For scalability and to reduce resource demands on the network, these sets must be limited in size. So verifiers can and should impose reasonable limits on the value of exp (which is the time by which the transaction must be included in a block or become invalid).
 
 To support long-living pre-signed transactions, the protocol allows a nonce to use the blockchain’s initial block ID regardless of the `refscount` limit specified in the block headers.
 
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -407,7 +407,7 @@ Anchors can be created by the [`nonce`](#nonce) instruction or generated from pr
 VM keeps track of the last used anchor and fails if:
 
 1. by the end of the execution, no anchor was used (which means that [transaction ID](#transaction-id) is not unique), or
-2. an [`output`](#output) or [`contract`](#output) are invoked before the anchor is set (since they can’t be made unique).
+2. an [`output`](#output) or [`contract`](#contract) is invoked before the anchor is set (since they can’t be made unique).
 
 Note: using a chain of anchors enables the signer to have a predictable anchor computation, when the signable contract is created transiently from a just-opened UTXO, without the need to know about the entire transaction.
 
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -1277,20 +1277,25 @@ _qty flv metadata pred_ **issue** → _contract_
     T.commit("metadata", metadata)
     flavor = T.challenge_scalar("flavor")
     ```
-6. Checks that the `flv` has unblinded commitment to `flavor` by [deferring the point operation](#deferred-point-operations):
+6. Checks that the `flv` has unblinded commitment to `flavor`
+   by [deferring the point operation](#deferred-point-operations):
     ```
     flv == flavor·B
     ```
-7. Adds a 64-bit range proof for the `qty` to the [constraint system](#constraint-system) (see [Cloak protocol](../../spacesuit/spec.md) for the range proof definition).
+7. Adds a 64-bit range proof for the `qty` to the [constraint system](#constraint-system)
+   (see [Cloak protocol](../../spacesuit/spec.md) for the range proof definition).
 8. Adds an [issue entry](#issue-entry) to the [transaction log](#transaction-log).
-9. Creates a [contract](#contract-type) with the value as the only [payload](#contract-payload), protected by the predicate `pred`.
+9. Creates a [contract](#contract-type) with the value as the only [payload](#contract-payload),
+   protected by the predicate `pred`, consuming [VM’s last anchor](#vm-state)
+   and replacing it with this contract’s [ID](#contract-id).
 
 The value is now issued into the contract that must be unlocked
 using one of the contract instructions: [`signtx`](#signtx), [`delegate`](#delegate) or [`call`](#call).
 
 Fails if:
 * `pred` is not a valid [point](#point),
-* `flv` or `qty` are not [variable types](#variable-type).
+* `flv` or `qty` are not [variable types](#variable-type),
+* VM’s [last anchor](#vm-state) is missing.
 
 
 #### borrow
@@ -1424,10 +1429,8 @@ _prevoutput_ **input** → _contract_
 
 1. Pops a [data](#data-type) `prevoutput` representing the [unspent output structure](#output-structure) from the stack.
 2. Constructs a [contract](#contract-type) based on the `prevoutput` data and pushes it to the stack.
-3. For each decoded [value](#value-type), quantity variable is allocated first, flavor second.
-4. Adds [input entry](#input-entry) to the [transaction log](#transaction-log).
-5. Sets the [VM’s last anchor](#vm-state) to the [contract ID](#contract-id) of the claimed 
-[UTXO](#utxo).
+3. Adds [input entry](#input-entry) to the [transaction log](#transaction-log).
+4. Sets the [VM’s last anchor](#vm-state) to the [contract ID](#contract-id) of the claimed [UTXO](#utxo).
 
 Fails if the `prevoutput` is not a [data type](#data-type) with exact encoding of an [output structure](#output-structure).
 
@@ -1441,11 +1444,11 @@ _items... predicate_ **output:_k_** → ø
     ```
     T = Transcript("ZkVM.ratchet-anchor")
     T.commit("anchor", vm.last_anchor)
-    next_anchor = T.challenge_bytes("vm_anchor")
-    contract_anchor = T.challenge_bytes("contract_anchor")
+    next_anchor = T.challenge_bytes("anchor1")
+    contract_anchor = T.challenge_bytes("anchor2")
     ```
 4. Updates the [VM’s last anchor](#vm-state) with a ratcheted `next_anchor`:
-5. Uses ratcheted `contract_anchor` in the [output structure](#output-structure).
+5. Uses `contract_anchor` in the [output structure](#output-structure).
 6. Adds an [output entry](#output-entry) to the [transaction log](#transaction-log).
 
 Immediate data `k` is encoded as [LE32](#le32).
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -376,9 +376,7 @@ Time bounds are available in the transaction as [expressions](#expression-type) 
 
 ### Contract ID
 
-_Contract ID_ is a 32-byte commitment to the [contract](#contract-type) ([anchor](#anchor), [payload](#contract-payload) and [predicate](#predicate)) that is also a unique identifier.
-
-Contract ID is a hash of the contract’s [output structure](#output-structure):
+_Contract ID_ is a 32-byte unique identifier of the [contract](#contract-type) and a commitment to its contents (see [Output](#output-structure)):
 
 ```
 T = Transcript("ZkVM.contractid")
@@ -386,37 +384,31 @@ T.commit("contract", output_structure)
 id = T.challenge_bytes("id")
 ```
 
-Contract ID makes [`delegate`](#delegate) signatures safe against [predicate](#predicate) key reuse:
-signature covers the unique contract ID, therefore preventing its replay against another contract,
+Contract ID makes signatures safe against reused [predicate](#predicate) keys:
+a signature covers the unique contract ID, therefore preventing its replay against another contract,
 even if containing the same [payload](#contract-payload) and predicate.
 
-Contract ID is made globally unique due to use of [anchor](#anchor).
-
+Uniqueness is provided via the [anchor](#anchor).
 
 ### Anchor
 
 _Anchor_ is a 32-byte unique string that provides uniqueness to the [contract ID](#contract-id).
-
 Anchors can be created by the [`nonce`](#nonce) instruction or generated from previously used unique contract IDs, tracked by the VM via [last anchor](#vm-state):
 
 1. Nonce contract has its anchor computed from the nonce parameters (see [`nonce`](#nonce) instruction).
 2. Claimed UTXO ([`input`](#input)) sets the VM’s [last anchor](#vm-state) to its _ratcheted_ [contract ID](#contract-id) (see [`input`](#input)).
 3. Newly created contracts and outputs ([`contract`](#contract), [`output`](#contract)) consume the VM’s [last anchor](#vm-state) and replace it with its [contract ID](#contract-id).
 
-VM keeps track of the last used anchor and fails if:
+VM fails if:
 
-1. by the end of the execution, no anchor was used (which means that [transaction ID](#transaction-id) is not unique), or
-2. an [`issue`](#issue), [`output`](#output) or [`contract`](#contract) is invoked before the anchor is set (since they can’t be made unique).
+1. an [`issue`](#issue), [`output`](#output) or [`contract`](#contract) is invoked before the anchor is set,
+2. by the end of the execution, no anchor was used (which means that [transaction ID](#transaction-id) is not unique).
 
-#### Note 1
-
-Chaining the anchors this way gives flexibility to the signer:
+Note 1: chaining the anchors this way gives flexibility to the signer:
 if contract A creates contract B, the B's ID can be computed solely
 from the contents of A, without the knowledge of the entire transaction.
 
-#### Note 2
-
-Inputs _ratchet_ the contract ID because this contract ID was already available as an anchor in the _previous transaction_ (where the output was created).
+Note 2: [`input`](#input) _ratchets_ the contract ID because this contract ID was already available as an anchor in the _previous transaction_ (where the output was created).
 
 
 ### Transcript
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -858,7 +858,7 @@ The VM is initialized with the following state:
 
 1. [Transaction](#transaction) as provided by the user.
 2. Extension flag set to `true` or `false` according to the [transaction versioning](#versioning) rules for the transaction version.
-3. Last anchor is set to `None`.
+3. Last anchor is set to ∅.
 4. Data stack is empty.
 5. Program stack is empty.
 6. Current program set to the transaction program; with zero offset.
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -378,11 +378,20 @@ Time bounds are available in the transaction as [expressions](#expression-type) 
 
 _Contract ID_ is a 32-byte commitment to the [contract](#contract-type) ([anchor](#anchor), [payload](#contract-payload) and [predicate](#predicate)) that is also a unique identifier.
 
+Contract ID is a hash of the contract’s [output structure](#output-structure):
+
+```
+T = Transcript("ZkVM.contractid")
+T.commit("contract", output_structure)
+id = T.challenge_bytes()
+```
+
 Contract ID makes [`delegate`](#delegate) signatures safe against [predicate](#predicate) key reuse:
 signature covers the unique contract ID, therefore preventing its replay against another contract,
 even if containing the same [payload](#contract-payload) and predicate.
 
 Contract ID is made globally unique due to use of [anchor](#anchor).
+
 
 ### Anchor
 
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -400,14 +400,13 @@ _Anchor_ is a 32-byte unique string that provides uniqueness to the [contract ID
 Anchors can be created by the [`nonce`](#nonce) instruction or generated from previously used unique contract IDs, tracked by the VM via [last anchor](#vm-state):
 
 1. Nonce contract has its anchor computed from the nonce parameters (see [`nonce`](#nonce) instruction).
-2. Claimed UTXO ([`input`](#input)) sets the VM’s [last anchor](#vm-state) to its [contract ID](#contract-id).
-3. Transient contract ([`contract`](#contract)) consumes the VM’s [last anchor](#vm-state) and replaces it with its [contract ID](#contract-id).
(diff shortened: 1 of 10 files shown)


zkvm/docs/zkvm-spec.md
```diff
@@ -1390,49 +1408,91 @@ Fails if:
...
 _items... predicate_ **output:_k_** → ø
 
 1. Pops [`predicate`](#predicate) from the stack.
 2. Pops `k` items from the stack.
-3. Adds an [output entry](#output-entry) to the [transaction log](#transaction-log).
+3. Use the [VM’s last anchor](#vm-state) as an anchor for the [output structure](#output-structure).
+4. Update the [VM’s last anchor](#vm-state) with a ratcheted [contract ID](#contract-id):
+    ```
+    T = Transcript("ZkVM.ratchet-anchor")
+    T.commit("output", output_id)
```
vickiniu: Here, what is `output_id`? Is this the VM's last anchor?

zkvm/docs/zkvm-spec.md
```diff
@@ -371,6 +374,42 @@ Time bounds are available in the transaction as [expressions](#expression-type)
...
+
+
+### Anchor
+
+_Anchor_ is a 32-byte unique string that provides uniqueness to the [contract ID](#contract-id).
+
+Anchors can be created by the [`nonce`](#nonce) instruction or generated from previously used unique contract IDs, tracked by the VM via [last anchor](#vm-state):
+
+1. Nonce contracts have their anchor computed from the nonce parameters (see [`nonce`](#nonce) instruction).
+2. Claimed UTXOs ([`input`](#input)) carry over the previously created anchor and set the "last anchor" to their [contract ID](#contract-id).
```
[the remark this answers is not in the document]
oleganza: will clarify, thanks

zkvm/docs/zkvm-spec.md
oleganza: rewrote this part here: 
https://github.com/interstellar/slingshot/pull/193/commits/6b995d569476e1bf2b41f03fd3222482f80403bf

zkvm/docs/zkvm-spec.md
```diff
@@ -486,47 +525,25 @@ Payload of a [contract](#contract-type) may contain arbitrary [types](#types),
...
    Predicate  =  <32 bytes>
         Item  =  enum { Data, Value }
         Data  =  0x00  ||  LE32(len)  ||  <bytes>
        Value  =  0x01  ||  <32 bytes> ||  <32 bytes>
 ```
 
+### UTXO
+
+UTXO stands for Unspent Transaction [Output](#output-structure).
+UTXO is uniquely identified by the [contract ID](#contract-id).
```
bobg:
```suggestion
UTXO is uniquely identified by the [ID](#contract-id) of the contract in the output.
```

zkvm/docs/zkvm-spec.md
```diff
@@ -371,6 +374,42 @@ Time bounds are available in the transaction as [expressions](#expression-type)
...
+### Contract ID
+
+_Contract ID_ is a 32-byte commitment to the [contract](#contract-type) ([anchor](#anchor), [payload](#contract-payload) and [predicate](#predicate)) that is also a unique identifier.
+
+Contract ID is a hash of the contract’s [output structure](#output-structure):
+
+```
+T = Transcript("ZkVM.contractid")
+T.commit("contract", output_structure)
+id = T.challenge_bytes()
```
bobg: Does `challenge_bytes` not require a string argument like `"id"`?

If not, why is it used in some places?

zkvm/docs/zkvm-spec.md
```diff
@@ -371,6 +374,42 @@ Time bounds are available in the transaction as [expressions](#expression-type)
...
+
+### Anchor
+
+_Anchor_ is a 32-byte unique string that provides uniqueness to the [contract ID](#contract-id).
+
+Anchors can be created by the [`nonce`](#nonce) instruction or generated from previously used unique contract IDs, tracked by the VM via [last anchor](#vm-state):
+
+1. Nonce contract has its anchor computed from the nonce parameters (see [`nonce`](#nonce) instruction).
+2. Claimed UTXO ([`input`](#input)) sets the VM’s [last anchor](#vm-state) to its [contract ID](#contract-id).
+3. Transient contract ([`contract`](#contract)) consumes the VM’s [last anchor](#vm-state) and replaces it with its [contract ID](#contract-id).
```
bobg: Is it safe for `contract` to use last-anchor as-is? If so, can you say why? Or should it be ratcheted as with `output`?

zkvm/docs/zkvm-spec.md
```diff
@@ -825,7 +843,7 @@ The ZkVM state consists of the static attributes and the state machine attribute
     * `tx_signature`
     * `cs_proof`
 2. Extension flag (boolean)
-3. Uniqueness flag (boolean)
+3. Last anchor (optional [anchor](#anchor): `None` or `Some(anchor)`)
```
bobg: Avoid Rust idioms in the spec. I suggest something like:

```suggestion
3. Last [anchor](#anchor) or ∅ if unset.
```

zkvm/docs/zkvm-spec.md
```diff
@@ -1390,49 +1408,93 @@ Fails if:
...
-2. Pushes a new [contract](#contract-type) with an empty [payload](#contract-payload) and this predicate to the stack.
-3. Adds [nonce entry](#nonce-entry) to the [transaction log](#transaction-log) with the predicate and transaction [maxtime](#time-bounds).
-4. Sets the [VM uniqueness flag](#vm-state) to `true`.
+1. It provides a unique anchor for an issuing transaction if other anchors are not available (i.e., from other values already on the stack);
+2. It binds the entire transaction to a particular blockchain, protecting not only against cross-blockchain replays, but also potential blockchain forks due to compromise of the old block-signing keys. This is a necessary feature for stateless signing devices that rely on blockchain proofs.
 
-Fails if `predicate` is not a valid [point](#point).
+Blockchain state machine performs the following checks:
+
+1. The `blockid` is checked against the IDs of “recent” blocks when the transaction is applied to the blockchain. It must match a recent block, or the initial block of the blockchain.
```
bobg: TxVM also allows it to be 0x00 × 32 bytes. I'm fine with excluding that possibility but first let's make sure we remember why TxVM allows it in the first place. (I don't, do you?)

zkvm/docs/zkvm-spec.md
```diff
@@ -1390,49 +1408,93 @@ Fails if:
...
+**Discussion**
+
+A nonce serves two purposes:
 
-1. Pops [predicate](#predicate) from the stack.
-2. Pushes a new [contract](#contract-type) with an empty [payload](#contract-payload) and this predicate to the stack.
-3. Adds [nonce entry](#nonce-entry) to the [transaction log](#transaction-log) with the predicate and transaction [maxtime](#time-bounds).
-4. Sets the [VM uniqueness flag](#vm-state) to `true`.
+1. It provides a unique anchor for an issuing transaction if other anchors are not available (i.e., from other values already on the stack);
+2. It binds the entire transaction to a particular blockchain, protecting not only against cross-blockchain replays, but also potential blockchain forks due to compromise of the old block-signing keys. This is a necessary feature for stateless signing devices that rely on blockchain proofs.
```
bobg: > block-signing keys

Not clear to me that we're going to have those. Perhaps only in the consensus layer, not in the blockchain data model per se.

zkvm/docs/zkvm-spec.md
```diff
@@ -1906,23 +1968,22 @@ ZkVM ensures transaction uniqueness this way:
...
-
-Handling values becomes much simpler — they are just values. So we can issue, finalize and even “claim” a value in a straightforward manner.
-
 The values are simply pedersen commitments (Q,A) (quantity, flavor), without any extra payload.
+This makes handling them much simpler: we can issue, finalize and even “claim” a value in a straightforward manner.
 
 **Con:**
 
-UTXO IDs are not known until the full transaction log is formed. This could be not a big deal, as we cannot really plan for the next transaction until this one is fully formed and published. Also, in a joint proof scenario, it’s even less reliable to plan the next payment until the MPC is completed, so requirement to wait till transaction ID is determined may not be a big deal.
+UTXO IDs are not generally known without fuller view on transaction flow. This could be not a big deal, as we cannot really plan for the next transaction until this one is fully formed and published. Also, in a joint proof scenario, it’s even less reliable to plan the next payment until the MPC is completed, so requirement to wait till transaction ID is determined may not be a big deal.
```
bobg: Expand MPC?

zkvm/docs/zkvm-spec.md
```diff
@@ -371,6 +374,42 @@ Time bounds are available in the transaction as [expressions](#expression-type)
...
+2. Claimed UTXO ([`input`](#input)) sets the VM’s [last anchor](#vm-state) to its [contract ID](#contract-id).
+3. Transient contract ([`contract`](#contract)) consumes the VM’s [last anchor](#vm-state) and replaces it with its [contract ID](#contract-id).
+4. Newly created outputs ([`output`](#output)) consume the VM’s [last anchor](#vm-state) and replace it with a ratcheted one (because the contract ID will be used as an anchor in a later transaction, via [`input`](#input)).
+
+VM keeps track of the last used anchor and fails if:
+
+1. by the end of the execution, no anchor was used (which means that [transaction ID](#transaction-id) is not unique), or
+2. an [`output`](#output) or [`contract`](#output) are invoked before the anchor is set (since they can’t be made unique).
+
+Note: using a chain of anchors enables the signer to have a predictable anchor computation, when the signable contract is created transiently from a just-opened UTXO, without the need to know about the entire transaction.
```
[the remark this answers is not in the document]
oleganza: How about this?

> Chaining the anchors this way gives flexibility to the signer: if contract A creates contract B, the B's ID can be computed solely from the contents of A, without the knowledge of the entire transaction.

zkvm/docs/zkvm-spec.md
oleganza: it requires, i fixed it and not pushed it yet

zkvm/docs/zkvm-spec.md
oleganza: WDYT?

> Transient contract sets its ID as an anchor because this can be done only once: the contract is soon destroyed by `signtx`, `call` or `delegate`.
> This is not the case with `output`: if the output’s ID was set as the anchor, it'd be available in two transactions: in this one, for the rest of the execution, and in the future one, when the output is claimed by `input` which sets the same output ID as an anchor.

zkvm/docs/zkvm-spec.md
oleganza: good point

zkvm/docs/zkvm-spec.md
oleganza: It was allowed to support hypothetical case where we can do fancy replay of the same transaction on two TxVM chains to do an unspecified shorter cross-chain swap. However, having this possibility, precludes a guarantee that all anchors are globally unique. So if someone somewhere used an zero-bound nonce, its non-uniqueness can shoot someone's foot down the road through the chain of anchor derivation.

Since we are going to make "cross-chain swaps" AKA "payment channels" by the book and be happy with that, I don't think we need to keep this dangerous behavior.

zkvm/docs/zkvm-spec.md
oleganza: yeah, that paragraph is very much out of scope. How about: 
```suggestion
2. It binds the entire transaction to a particular blockchain, protecting not only against cross-blockchain replays, but also potential blockchain forks. This is a necessary feature for stateless signing devices that rely on blockchain proofs.
```

zkvm/docs/zkvm-spec.md
oleganza:
```suggestion
UTXO IDs are not generally known without fuller view on transaction flow. This could be not a big deal, as we cannot really plan for the next transaction until this one is fully formed and published. Also, in a joint proof scenario, it’s even less reliable to plan the next payment until the multi-party protocol is completed, therefore waiting for transaction ID to be determined becomes a requirement.
```

zkvm/docs/zkvm-spec.md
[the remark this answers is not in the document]
bobg: Looks good, thanks!

zkvm/docs/zkvm-spec.md
bobg: OK ✔️

zkvm/docs/zkvm-spec.md
bobg: OK

zkvm/docs/zkvm-spec.md
bobg: That looks good, but note that block signatures are also mentioned in one or two other places.

zkvm/docs/zkvm-spec.md
bobg: :+1:

zkvm/docs/zkvm-spec.md
oleganza: oops

zkvm/src/txlog.rs
```diff
@@ -67,15 +67,16 @@ impl MerkleItem for Entry {
...
             }
-            Entry::Input(utxo) => {
-                t.commit_bytes(b"input", &utxo.0);
+            Entry::Input(id) => {
+                t.commit_bytes(b"input", id.as_bytes());
             }
-            Entry::Nonce(pred, maxtime) => {
-                t.commit_point(b"nonce.p", &pred);
-                t.commit_u64(b"nonce.t", *maxtime);
+            Entry::Output(output) => {
```
vickiniu: Why do we have `Entry::Output` contain the whole output structure if we only log the id? could we have it mirror input & only contain the id?

zkvm/src/vm.rs
```diff
@@ -75,9 +75,8 @@ where
     // we allow treating unassigned opcodes as no-ops.
     extension: bool,
 
-    // set to true by `input` and `nonce` instructions
-    // when the txid is guaranteed to be unique.
-    unique: bool,
+    // updated by input/nonce/contract/output instructions
```
vickiniu: and the `issue` instruction?

zkvm/docs/zkvm-spec.md
```diff
@@ -371,6 +374,42 @@ Time bounds are available in the transaction as [expressions](#expression-type)
...
+
+_Contract ID_ is a 32-byte unique identifier of the [contract](#contract-type) and a commitment to its contents (see [Output](#output-structure)):
+
+```
+T = Transcript("ZkVM.contractid")
+T.commit("contract", output_structure)
+id = T.challenge_bytes("id")
+```
+
+Contract ID makes signatures safe against reused [predicate](#predicate) keys:
```
bobg: What does the phrase "predicate key" mean here?

zkvm/docs/zkvm-spec.md
```diff
@@ -371,6 +374,42 @@ Time bounds are available in the transaction as [expressions](#expression-type)
...
+a signature covers the unique contract ID, therefore preventing its replay against another contract,
+even if containing the same [payload](#contract-payload) and predicate.
+
+Uniqueness is provided via the [anchor](#anchor).
+
+### Anchor
+
+_Anchor_ is a 32-byte unique string that provides uniqueness to the [contract ID](#contract-id).
+Anchors can be created by the [`nonce`](#nonce) instruction or generated from previously used unique contract IDs, tracked by the VM via [last anchor](#vm-state):
```
bobg:
```suggestion
Anchors can be created by the [`nonce`](#nonce) instruction or generated from unique contract IDs used earlier in the same transaction. These are tracked by the VM via [last anchor](#vm-state):
```

zkvm/docs/zkvm-spec.md
```diff
@@ -486,47 +525,24 @@ Payload of a [contract](#contract-type) may contain arbitrary [types](#types),
...
-utxo = T.challenge_bytes("id")
-```
-
-In the above, `previous_txid` is the [transaction ID](#transaction-id) of the transaction where the output was created,
-and `previous_output` is the serialized [output](#output-structure).
-
-
 ### Output structure
 
 Output represents a _snapshot_ of a [contract](#contract-type)
```
bobg: Putting snapshot in italics sort of implies it's a technical term whose definition is nearby. Perhaps it should not be in italics and you can get away with saying something like ..."a snapshot of a contract's state at the moment it executed an `output` instruction."

zkvm/docs/zkvm-spec.md
```diff
@@ -841,7 +858,7 @@ The VM is initialized with the following state:
 
 1. [Transaction](#transaction) as provided by the user.
 2. Extension flag set to `true` or `false` according to the [transaction versioning](#versioning) rules for the transaction version.
-3. Uniqueness flag is set to `false`.
+3. Last anchor is set to `None`.
```
bobg:
```suggestion
3. Last anchor is set to ∅.
```

(To match some earlier text.)

zkvm/docs/zkvm-spec.md
```diff
@@ -866,7 +883,7 @@ Then, the VM executes the current program till completion:
 
 If the execution finishes successfully, VM performs the finishing tasks:
 1. Checks if the stack is empty; fails otherwise.
-2. Checks if the uniqueness flag is set to `true`; fails otherwise.
+2. Checks if the last anchor is present; fails otherwise.
```
bobg:
```suggestion
2. Checks if the last anchor is set to other than ∅; fails otherwise.
```

zkvm/docs/zkvm-spec.md
```diff
@@ -1247,20 +1264,25 @@ _qty flv metadata pred_ **issue** → _contract_
...
+   and replacing it with this contract’s [ID](#contract-id).
 
 The value is now issued into the contract that must be unlocked
 using one of the contract instructions: [`signtx`](#signtx), [`delegate`](#delegate) or [`call`](#call).
 
 Fails if:
 * `pred` is not a valid [point](#point),
-* `flv` or `qty` are not [variable types](#variable-type).
+* `flv` or `qty` are not [variable types](#variable-type),
+* VM’s [last anchor](#vm-state) is missing.
```
bobg:
```suggestion
* VM’s [last anchor](#vm-state) is not set.
```

zkvm/docs/zkvm-spec.md
```diff
@@ -1390,49 +1412,90 @@ Fails if:
...
+2. Constructs a [contract](#contract-type) based on the `prevoutput` data and pushes it to the stack.
+3. Adds [input entry](#input-entry) to the [transaction log](#transaction-log).
+4. Sets the [VM’s last anchor](#vm-state) to the ratcheted [contract ID](#contract-id):
+    ```
+    T = Transcript("ZkVM.ratchet-anchor")
+    T.commit("old", contract_id)
+    new_anchor = T.challenge_bytes("new")
+    ```
+
+Fails if the `prevoutput` is not a [data type](#data-type) with exact encoding of an [output structure](#output-structure).
```
bobg: I think this whole document needs to replace the way it uses "data type." Brought this up [in Slack](https://i10r.slack.com/archives/CCVPR72CR/p1549411686017000?thread_ts=1549408834.013200&cid=CCVPR72CR) originally:

> the way the spec uses the phrase "data type" is causing some cognitive dissonance. e.g. "scalars are encoded as 32-byte data types." those are _strings_ (or _blobs_ or _bytearrays_ or something). a "data type" should be a type, not data

zkvm/docs/zkvm-spec.md
```diff
@@ -1390,49 +1412,90 @@ Fails if:
...
 #### output
 
 _items... predicate_ **output:_k_** → ø
 
 1. Pops [`predicate`](#predicate) from the stack.
 2. Pops `k` items from the stack.
-3. Adds an [output entry](#output-entry) to the [transaction log](#transaction-log).
+3. Creates a contract with the `k` items as a payload, the predicate `pred`, and anchor set to the [VM’s last anchor](#vm-state).
+4. Adds an [output entry](#output-entry) to the [transaction log](#transaction-log).
+5. Update the [VM’s last anchor](#vm-state) with the [contract ID](#contract-id) of the new contract.
```
bobg:
```suggestion
5. Updates the [VM’s last anchor](#vm-state) with the [contract ID](#contract-id) of the new contract.
```

zkvm/docs/zkvm-spec.md
```diff
@@ -1390,49 +1412,90 @@ Fails if:
...
 
 Immediate data `k` is encoded as [LE32](#le32).
 
+Fails if VM’s [last anchor](#vm-state) is missing.
```
bobg:
```suggestion
Fails if VM’s [last anchor](#vm-state) is not set.
```

zkvm/src/txlog.rs
oleganza: Placing whole output there comes at no cost and it's useful in /token project where the output is taken out of the log w/o having to insert the hooks into VM execution to get it out for indexing.

zkvm/docs/zkvm-spec.md
oleganza: yeah, makes sense

zkvm/docs/zkvm-spec.md
oleganza: this reads strangely to me. Maybe simply  "checks if the last anchor is set"?

zkvm/docs/zkvm-spec.md
oleganza: Let's revisit this in another PR. I agree that "data" is confusing, but "blob" and "bytearray" sound clunky and "string" sounds too much like "human text" (e.g. most modern languages have String type for UTF8 text). Motivation for "data" was as opposite to "non-data" linear types like "Values" and "Contracts". Those are not _data types_ because those are _things_. But i don't deny there's confusion alike to would-be one with String-as-bytes VS String-as-humantext.

zkvm/src/vm.rs
oleganza: yup, added

zkvm/docs/zkvm-spec.md
oleganza: Changed to 
```
Contract ID makes signatures safe against reused [verification keys](#verification-key) in [predicates](#predicate):
```

zkvm/docs/zkvm-spec.md
oleganza: changed to

```
Output is a serialized [contract](#contract-type):
```

zkvm/docs/zkvm-spec.md
oleganza:
```
2. Checks if the last anchor is set; fails otherwise.
```

zkvm/docs/zkvm-spec.md
bobg: Sounds good. To get a little pedantic, in the phrase "data type," "type" works as a noun and "data" works as an adjective modifying it; so it's a kind of type, not a kind of data. (The phrase "type data" would be a kind of data, not a kind of type.) My objection is not to the word "data" per se, but to "type" used as a noun to describe data.

oleganza: Filed: 
https://github.com/interstellar/slingshot/issues

Issue #187, opened 2019-03-04
slidechain: possible flakiness in post-pegout tx


bobg:
The post-pegout TxVM transaction retires exported funds after a successful peg-out on the Stellar side. (Alternatively, it repays the TxVM funds to the exporter if pegging out encounters a permanent error.)

This works by invoking the second phase of a smart contract in which the TxVM funds were locked during export. Invoking the second phase means constructing the contract in the exact form it had at the end of its first phase, which ended with `output`. Reconstructing the contract happens [here](https://github.com/interstellar/slingshot/blob/0c51a92260af675a4067154c9b5c370272303081/slidechain/postpegout.go#L44-L66).

A possible problem here is that the `refdata` value on the contract's stack is computed via a call to `json.Marshal` ([here](https://github.com/interstellar/slingshot/blob/0c51a92260af675a4067154c9b5c370272303081/slidechain/postpegout.go#L34)).

This is _probably_ fine. On the other hand, the `encoding/json` package does not make a guarantee about always producing the bytewise identical output for the same input. If a future version of `encoding/json` produced an equivalent JSON object encoded slightly differently, it could lead to a loss-of-funds scenario.

A safer approach would be to store the JSON string from the contract's phase 1 in the database, then recover and double-check it during phase 2, instead of recomputing it.

PR #195, opened 2019-03-05
slidechain: fix post-peg-out flakiness

debnil:
Fixes possible flakiness in the post-peg-out transaction. To do this, this simplifies the `exports` table to now only store the JSON representation of the peg-out struct, rather than all of the struct's constituent fields. Addresses #187.


slidechain/postpegout.go
```diff
@@ -31,10 +32,13 @@ func (c *Custodian) doPostPegOut(ctx context.Context, assetXDR, anchor, txid []b
 		Anchor:   anchor,
 		Pubkey:   pubkey,
 	}
-	refdata, err := json.Marshal(ref)
+	checkRefdata, err := json.Marshal(ref)
 	if err != nil {
 		return errors.Wrap(err, "marshaling reference data")
 	}
+	if !bytes.Equal(refdata, checkRefdata) {
```
bobg: This has the same problem as described in #187 : namely, that we're relying on bytewise equality between two different calls to `json.Marshal`, and we might not get it.

The right fix is is to _unmarshal_ the old refdata string into a `pegOut` structure and make sure all the fields are as we expect.

bobg: > make sure all the fields are as we expect

Sorry, that's misstated. There's nothing to check. We can just rely on the values unmarshaled from the json string for constructing the contract snapshot for `input`, and do away with the now-redundant other fields of the `exports` table.

slidechain/export.go
```diff
@@ -119,83 +119,69 @@ func (c *Custodian) pegOutFromExports(ctx context.Context, pegouts chan<- pegOut
...
-						log.Fatalf("getting error codes from failed submission of tx %x (with horizon err '%s'): %s", txid, herr, rerr)
+						log.Fatalf("getting error codes from failed submission of tx %x (with horizon err '%s'): %s", p.TxID, herr, rerr)
 					}
 					if resultCodes.TransactionCode == xdr.TransactionResultCodeTxBadSeq.String() {
 						peggedOut = pegOutRetry
 					}
 				}
 			}
-			result, err := c.DB.ExecContext(ctx, `UPDATE exports SET pegged_out=$1 WHERE txid=$2`, peggedOut, txid)
+			result, err := c.DB.ExecContext(ctx, `UPDATE exports SET pegged_out=$1 WHERE ref=$2`, peggedOut, ref)
```
bobg: There is no index on `ref` and I don't think there should be, considering that it includes extra information that isn't identifying. I think you need to add back a `txid` column to be the table's primary key.

slidechain/slidechain_test.go
```diff
@@ -564,18 +576,24 @@ func isPostPegOutTx(tx *bc.Tx, asset xdr.Asset, amount int64, tempAddr, exporter
...
+			Amount:   amount,
+			Anchor:   anchor,
+			Pubkey:   pubkey,
+		}
+		refdata, err := json.Marshal(ref)
+		if err != nil {
+			log.Printf("unmarshaling json: %s", err)
+			return false
+		}
+		if !bytes.Equal(refdata, tx.Log[2][2].(txvm.Bytes)) {
```
bobg: This is once again making the same mistake. Instead of marshaling `ref` and comparing it bytewise to the log entry, you should be unmarshaling the log entry into a separate `pegOut` and comparing that with `ref` (e.g. using `reflect.DeepEqual`).

It doesn't matter as much here, since this is just a test, and the same implementation of `json.Marshal` will be used for both this call and the call that produces the log item. The chances of those two marshalings diverging, bytewise, is practically nil. But considering that this is the PR intending to avoid that very possibility, might as well avoid it.

debnil: I was thinking exactly what you said in the second paragraph, but it makes total sense to just avoid this possibility in the test as well, as that is the intention of the PR.

slidechain/slidechain_test.go
```diff
@@ -564,18 +576,25 @@ func isPostPegOutTx(tx *bc.Tx, asset xdr.Asset, amount int64, tempAddr, exporter
...
-		Amount:   amount,
-		Anchor:   anchor,
-		Pubkey:   pubkey,
-	}
-	refdata, err := json.Marshal(ref)
-	if !bytes.Equal(refdata, tx.Log[2][2].(txvm.Bytes)) {
-		return false
+	if foundRetire {
+		var logRef pegOut
+		err := json.Unmarshal(tx.Log[2][2].(txvm.Bytes), logRef)
```
bobg: This needs `&logRef`, not `logRef`. As it is, this should be a no-op - `logRef` should remain the zero `pegOut` object after this call. But that raises the question: why does the test calling this function not fail?

debnil: Great catch! Thanks to looking into this, I realized this logic was, in fact, not being executed. Instead, the "refund" clause was always called. The `State` of the `pegOut p` was not being set correctly to `PegOutOk` before being sent to `doPostPegOut`. As a result, the post-peg-out program always refunded the funds, rather than retiring them. This is fixed now.

slidechain/watch.go
```diff
@@ -207,7 +197,11 @@ func (c *Custodian) watchPegOuts(ctx context.Context, pegouts <-chan pegOut) {
 			if !ok {
 				log.Fatalf("peg-outs channel closed")
 			}
-			err := c.doPostPegOut(ctx, p.AssetXDR, p.Anchor, p.TxID, p.Amount, p.Seqnum, p.State, p.Exporter, p.TempAddr, p.Pubkey)
+			// This peg-out struct is marshalled into a JSON, which must equal the reference data in the export tx.
+			// That requires that the post-peg-out struct txid be empty.
```
bobg: There are a couple of better ways to do this. One is not to export the txid field, which causes the json package to skip over it. The other is to give the field the struct tag `json:"-"` which explicitly instructs the json package to skip it.

slidechain/postpegout.go
```diff
@@ -15,32 +15,35 @@ import (
 	"github.com/stellar/go/xdr"
 )
 
-func (c *Custodian) doPostPegOut(ctx context.Context, assetXDR, anchor, txid []byte, amount, seqnum int64, peggedOut pegOutState, exporter, tempAddr string, pubkey []byte) error {
+func (c *Custodian) doPostPegOut(ctx context.Context, p pegOut, txid []byte) error {
```
bobg: See my other comment about keeping txid out of the json serialization of `pegOut`. If you do that, you won't need a separate `txid` argument here.

slidechain/postpegout.go
```diff
@@ -15,32 +15,35 @@ import (
...
+	var selector int64
+	if p.State == pegOutOK {
+		selector = 1
+	}
+
+	// The refdata stored in the database has pegged-out state equal to 0.
+	// We set the pegOut struct state to this value temporarily.
+	// This lets us correctly reconstruct the contract's stack and call it.
+	currentState := p.State
+	p.State = 0
```
bobg: This has what Eric Rykwalder used to refer to as "code smell."

A slightly better approach would be to leave `p` alone and mutate a duplicate instead, which means we don't have to remember to put anything back the way it was:

```
p2 := p
p2.State = 0
...
refdata, err := json.Marshal(p2)
```

Better still would be to figure out some way that this hack isn't needed to begin with. Why does `p.State` diverge from what's in the db? Does that mean that `State` should be stored separately from the other items in `p`?

slidechain/schema.go
```diff
@@ -30,14 +30,8 @@ CREATE TABLE IF NOT EXISTS pegs (
...
   txid BLOB NOT NULL PRIMARY KEY,
-  exporter TEXT NOT NULL,
-  amount INTEGER NOT NULL,
-  asset_xdr BLOB NOT NULL,
-  temp_addr TEXT NOT NULL,
-  seqnum INTEGER NOT NULL,
   pegged_out INTEGER NOT NULL DEFAULT 0,
-  anchor BLOB NOT NULL,
-  pubkey BLOB NOT NULL
+  ref BLOB NOT NULL
```
bobg: This is json-encoded, right? So it can be `TEXT` (though not sure if that matters), and it should have a name that indicates its purpose and encoding, like `pegout_json` or something.

slidechain/postpegout.go
debnil: I think I was unclear when I explained the issue before, sorry about that.  `State` is already stored separately from the other items in `p`, via the `pegged_out` field in `exports`. `p.State` is consistent with that stored state at the beginning of this function.

However, it differs from the state stored in the JSON encoding of `p`; that is stored in the `pegout_json` field (formerly `ref`)  in the `exports` table.  I think the fix to this is simple and the same as one you suggested elsewhere: instruct the JSON package to skip encoding `State` in the JSON. This will ensure that the JSONs are consistent.

PR #196, opened 2019-03-06
zkvm/predicate: add basic predicate API to program builder

### LINKED ISSUE #145: zkvm: predicate tree traversal API
The VM's `left`/`right` instructions expect a proof of correct branching. In other words, these must be preceded with `push(L)` and `push(R)` instructions. Since the prover's `Predicate` type has all the structure to compute these values, it may be convenient to have Predicate type generate instructions itself:

```rust
impl Predicate {
    pub fn choose_left(&self, program: &mut Program) -> Result<(),VMError> {
        let (L,R) = self.to_disjunction()?;
        program.push(L);
        program.push(R);
        program.left();
    }
}
```

Note: the prover would need to have a reference to the predicate. It is usually available since the prover would construct an `Input` prior to that and can borrow the corresponding predicate from there.

vickiniu:
Adds `choose_left` and `choose_right` to program builder to push correct data items onto the stack for predicate tree traversal.

Fixes #145

### DIFF
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -4,6 +4,7 @@
 use crate::encoding;
 use crate::encoding::SliceReader;
 use crate::errors::VMError;
+use crate::predicate::Predicate;
 use crate::types::Data;
 use core::borrow::Borrow;
 use core::mem;
@@ -398,5 +399,25 @@ impl Program {
     pub fn cloak(&mut self, m: usize, n: usize) -> &mut Program {
         self.0.push(Instruction::Cloak(m, n));
         self
+    }
+
+    /// Adds the left and right predicates, followed by the `left` instruction in the correct order
+    /// to choose the left disjunction in the predicate tree.
+    pub fn choose_left(&mut self, pred: &Predicate) -> Result<&mut Program, VMError> {
+        let (l, r) = pred.to_disjunction()?;
+        self.push(l);
+        self.push(r);
+        self.left();
+        Ok(self)
+    }
+
+    /// Adds the left and right predicates, followed by the `right` instruction in the correct order
+    /// to choose the right disjunction in the predicate tree.
+    pub fn choose_right(&mut self, pred: &Predicate) -> Result<&mut Program, VMError> {
+        let (l, r) = pred.to_disjunction()?;
+        self.push(l);
+        self.push(r);
+        self.right();
+        Ok(self)
     }
 }
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -377,6 +377,13 @@ impl Program {
         self.0
     }
 
+    /// Converts the program to opaque serialized bytes.
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut prog_bytes = Vec::new();
+        self.encode(&mut prog_bytes);
+        prog_bytes
+    }
+
     /// Returns the serialized length of the program.
     pub fn serialized_length(&self) -> usize {
         self.0.iter().map(|p| p.serialized_length()).sum()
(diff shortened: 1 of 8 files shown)


oleganza:
Would be nice if there was a way to make multiple hops, descending down the predicate tree. Something that reads like `.left().left().right().call()` and ideally consumes the parent predicate, passing through one of the selected children. For that it probably should be oriented around Predicate, not Program:

`<Predicate (parent)>.choose_left(&mut Program) -> Predicate (child)` or something along these lines.

vickiniu:
ah, that makes sense! my thought was that having it as part of `Program` makes it jive better with the program builder since I thought it might be a little clunky to have:
```
Program::build(...);
predicate.choose_left(...);
Program::build(...);
```

oleganza:
How about:

```rust
program::build()
.push(stuff)
.push(stuff)
...
.choose_predicate(pred, |tree| {
    tree.left().right().left()
})
.call()
.more()
.things()
```

where `tree` is an aux type `ops::PredicateTree` that holds `&mut program` and owns `pred` and provides left/right traversal API?

zkvm/src/ops.rs
```diff
@@ -399,4 +407,61 @@ impl Program {
...
+        self.prog.push(l);
+        self.prog.push(r);
+        self.prog.left();
+
+        Ok(self)
+    }
+
+    /// Pushes program to the stack and calls the contract protected
+    /// by the program predicate.
+    pub fn call(&mut self, prog: Program) -> &mut Self {
```
oleganza: why does it need a Program? it seems like it supposed to have this program in the `self.pred`'s witness.

zkvm/src/prover.rs
```diff
@@ -57,10 +56,10 @@ impl<'a, 'b> Delegate<r1cs::Prover<'a, 'b>> for Prover<'a, 'b> {
         Ok(run.program.pop_front())
     }
 
-    fn new_run(&self, data: Data) -> Result<Self::RunType, VMError> {
-        Ok(ProverRun {
-            program: data.to_program()?.to_vec().into(),
-        })
+    fn new_run(&self, program: Program) -> Self::RunType {
```
oleganza: Is this change necessary? The verifier intentionally takes raw bytes so it can parse the program on the fly. Otherwise, it'd have to allocate 20x or something memory to store `Vec<Instruction>`

zkvm/src/predicate.rs
```diff
@@ -103,6 +103,14 @@ impl Predicate {
         }
     }
 
+    /// Downcasts the predicate to a disjunction.
+    pub fn to_disjunction(&self) -> Result<(Predicate, Predicate), VMError> {
```
oleganza: why not consume `self` like all other downcasting methods and avoid cloning?

zkvm/src/ops.rs
```diff
@@ -399,4 +407,61 @@ impl Program {
...
+/// Adds data and instructions to traverse a predicate tree.
+pub struct PredicateTree<'a> {
+    prog: &'a mut Program,
+    pred: Predicate,
+}
+
+impl<'a> PredicateTree<'a> {
+    /// Left Predicate branch
+    pub fn left(&mut self) -> Result<&mut Self, VMError> {
+        let (l, r) = self.pred.to_disjunction()?;
```
oleganza: i guess this answers my question about move/borrow. That's tricky since we cannot move out of the field, even if we immediately put something back in. And the thing we want to replace this with is not sitting somewhere separately (could use mem::replace then), but it's referenced by the field self.pred itself.

oleganza: We can overcome this by making PredicateTree consumed instead of &mut-borrowed in this API. This will also help with `call()`.

```rust
pub fn left(self) -> Result<Self, VMError> {
    let (l, r) = self.pred.to_disjunction()?; // assuming self.pred is moved out
    let prog = self.prog;
    prog.push(l).push(r).left();
    Ok(Self{pred: l, prog})  // move left predicate into new instance of the PredicateTree
}
```

zkvm/src/ops.rs
oleganza: And if we move tree in and out, this would be the terminal station:

```rust
pub fn call(self) -> Result<(), ...> {
    let subprog = self.pred.to_program()?;
    self.prog.push(subprog).call();
    Ok(())
}
```

zkvm/src/prover.rs
vickiniu: my thought was that it would be nice to have the process fail earlier if the `Data` that we pop off the stack is not program (rather than having it fail in `new_run`?) but not much reason besides that

oleganza: That won’t happen on verifier anyway - it has opaque data, but it will on the prover - we can mandate that the program is not opaque there if we want.

zkvm/src/ops.rs
vickiniu: Refactored to have the `PredicateTree` consumed, and expect the closure to return some `&mut Program` (rather than `&mut PredicateTree`)!

zkvm/src/ops.rs
```diff
@@ -399,4 +407,59 @@ impl Program {
...
+    pub fn right(self) -> Result<Self, VMError> {
+        let (l, r) = self.pred.to_disjunction()?;
+        let prog = self.prog;
+        prog.push(l).push(r.clone()).right();
+
+        Ok(Self { pred: r, prog })
+    }
+
+    /// Returns Predicate tree program.
+    pub fn program(self) -> &'a mut Program {
```
oleganza: do we use it?

zkvm/src/ops.rs
```diff
@@ -399,4 +407,59 @@ impl Program {
...
+
+    /// Takes predicate and closure to add choose operations for
+    /// predicate tree traversal.
+    pub fn choose_predicate<F>(
+        &mut self,
+        pred: Predicate,
+        choose_fn: F,
+    ) -> Result<&mut Program, VMError>
+    where
+        F: FnOnce(PredicateTree) -> Result<&mut Program, VMError>,
```
oleganza: asking the closure to return `&mut Program` won't work well if it ends with left/right - which return Self. Since we have &mut Program as &mut self already, we can simply return it ourselves, allowing the closure to return anything (which would be `PredicateTree` or `()`, depending on the user's operations):

```rust
pub fn choose_predicate<F,T>(...) where
        F: FnOnce(PredicateTree) -> Result<T, VMError>
```

zkvm/src/ops.rs
```diff
@@ -399,4 +400,55 @@ impl Program {
...
+    prog: &'a mut Program,
+    pred: Predicate,
+}
+
+impl<'a> PredicateTree<'a> {
+    /// Left Predicate branch
+    pub fn left(self) -> Result<Self, VMError> {
+        let (l, r) = self.pred.to_disjunction()?;
+        let prog = self.prog;
+        prog.push(l.clone()).push(r).left();
```
oleganza: Since we know that we are going to consume these predicates in the `left`/`right` instructions, we don't need to clone the entire predicate tree (`l.clone()`): we can convert to an opaque predicate. Moving `r` into push comes for free, but we can make it opaque too for symmetry.

Let's add `Predicate::to_opaque(&self) -> Predicate` method that returns `Predicate::Opaque(self.to_point())` and do this:

```suggestion
        prog.push(l.to_opaque()).push(r.to_opaque()).left();
```

Same thing for `right()`.

WDYT?

zkvm/src/ops.rs
```diff
@@ -399,4 +400,55 @@ impl Program {
...
+        let (l, r) = self.pred.to_disjunction()?;
+        let prog = self.prog;
+        prog.push(l).push(r.clone()).right();
+
+        Ok(Self { pred: r, prog })
+    }
+
+    /// Pushes program to the stack and calls the contract protected
+    /// by the program predicate.
+    pub fn call(self) -> Result<&'a mut Program, VMError> {
```
```suggestion
    pub fn call(self) -> Result<(), VMError> {
```

oleganza: the program is "owned" by the user of PredicateTree and we don't need to mix concerns in the chain-calling: we chain-call on PredicateTree only, and terminate the chain here to avoid any confusion.

zkvm/src/ops.rs
vickiniu: that sounds good to me :)

Issue #189, opened 2019-03-04
zkvm: should timestamps be in milliseconds vs seconds


oleganza:
@bobg raised a valid question regarding time bounds that are currently specced as UNIX timestamp in seconds. TxVM uses milliseconds. Which one should we use?

### Why seconds

1. Biggest reason: this is more "standard", therefore less potential confusion or mistakes by accident. Bitcoin/Ethereum/Stellar use seconds.
2. Precision may be just enough for a wide range of applications.
3. Lesser reason: inequality checks for time intervals can have cheaper range proofs. E.g. 1 month requires 20-bit rangeproof with seconds, but 30-bit one with milliseconds.

### Why milliseconds

@bobg argues:
1. It'll be better to have them and not need them than need them and not have them.
2. http://www.rinkworks.com/said/predictions.shtml

oleganza:
@bobg Feels to me that the extra precision offered by milliseconds is more useful than a vague "standardness" — after all, both versions are just scaled variants of each other with the same starting point. So I'd subscribe under your suggestion to switch to milliseconds. Can you do it in a separate PR, so the blockchain spec does not have to change it?

PR #198, opened 2019-03-06
zkvm/docs: blockchain spec

bobg:
Updates `zkvm-spec.md` to use millisecond-based timestamps.

### DIFF
--- a/zkvm/docs/zkvm-blockchain.md
+++ b/zkvm/docs/zkvm-blockchain.md
@@ -1,0 +1,106 @@
+# ZkVM blockchain specification
+
+This is the specification for the ZkVM blockchain,
+a blockchain containing [ZkVM transactions](zkvm-spec.md).
+
+# Blockchain state
+
+The blockchain state contains:
+
+- `initialheader`: The initial block header (at height 1).
+- `tipheader`: The latest block header.
+- `utxos`: The IDs of all available [utxos](zkvm-spec.md#utxo).
+
+# Block header
+
+A block header contains:
+
+- `version`: integer version number, set to 1.
+- `height`: integer block height. Initial block has height 1. Height increases by 1 with each new block.
+- `previd`: ID of the preceding block. For the initial block (which has no predecessor), an all-zero string of 32 bytes.
+- `timestamp`: integer timestamp of the block in milliseconds since the Unix epoch: 00:00:00 UTC Jan 1, 1970.
+  Each new block must have a time strictly later than the block before it.
+- `txroot`: 32-byte merkle root hash of the transactions in the block.
+- `utxoroot`: 32-byte merkle root hash of the utxo set after applying all transactions in the block.
+
+# Block
+
+A block contains:
+
+- `header`: a [block header](#block-header).
+- `txs`: a list of [transactions](zkvm-spec.md#transaction).
+
+# Block ID
+
+A block ID is computed from a [block header](#block-header) using the [transcript](zkvm-spec.md#transcript) mechanism:
+
+```
+T = Transcript("ZkVM.blockheader")
+T.commit(LE64(version))
+T.commit(LE64(height))
+T.commit(previd)
+T.commit(LE64(timestamp))
+T.commit(txroot)
+T.commit(utxoroot)
+blockid = T.challenge_bytes("id")
+```
+
+# Make initial block
+
+# Join network
+
+# Validate block
+
+Inputs:
+- `block`, the block to validate.
+- `prevheader`, the header of the previous block.
+
+Output:
+- list of [txlog](zkvm-spec.md#transaction-log)+[txid](zkvm-spec.md#transaction-id) pairs,
+  one for each transaction in block.txs.
+
+Procedure:
+1. Verify `block.header.version >= prevheader.version`.
+2. Verify `block.header.height == prevheader.height+1`.
+3. Verify `block.header.previd` equals the [block ID](#block-id) of `prevheader`.
+4. Verify `block.header.timestamp > prevheader.timestamp`.
+5. For each transaction `tx` in block.txs:
+   1. Verify `tx.mintime == 0` or `tx.mintime >= block.header.timestamp`.
+   2. Verify `tx.maxtime == 0` or `tx.maxtime < block.header.timestamp`.
+   3. If `block.header.version == 1`, verify `tx.version == 1`.
+   4. [Execute](zkvm-spec.md#vm-execution) `tx` to produce transaction log `txlog`.
+   5. Compute [transaction ID](zkvm-spec.md#transaction-id) `txid` from `tx` and `txlog`.
+   6. Add the pair `txlog`+`txid` to the list of output pairs.
+
+# Apply block
+
+Inputs:
+- `block`, the block to apply.
+- `state`, the current [blockchain state](#blockchain-state).
+
+Output:
+- New blockchain state.
+
+Procedure:
+1. [Validate](#validate-block) `block` with `prevheader` set to `state.tipheader`.
+2. Let `state′` be `state`.
+3. For each txlog from the validation step, in order:
+   1. [Apply](#apply-transaction-log) the txlog to `state′` to produce `state′′`.
+   2. xxx error check
+   3. Set `state′ <- state′′`.
+4. Compute `utxoroot` from `state′.utxos`.
+5. Verify `block.header.utxoroot == utxoroot`.
+6. Set `state′.tipheader <- block.header`.
+7. Return `state′`.
+
+# Apply transaction
+
+Inputs:
+- `txlog`, a [transaction log](zkvm-spec.md#transaction-log).
+- `state`, a [blockchain state](#blockchain-state).
+
+Output:
+- New blockchain state.
+
+Procedure:
+- 
--- a/zkvm/docs/zkvm-blockchain.md
+++ b/zkvm/docs/zkvm-blockchain.md
@@ -113,12 +113,12 @@ mechanism:
 
 ```
 T = Transcript("ZkVM.blockheader")
-T.commit(LE64(version))
-T.commit(LE64(height))
-T.commit(previd)
-T.commit(LE64(timestampms))
-T.commit(txroot)
-T.commit(utxoroot)
+T.commit("version", LE64(version))
+T.commit("height", LE64(height))
+T.commit("previd", previd)
+T.commit("timestampms", LE64(timestampms))
+T.commit("txroot", txroot)
+T.commit("utxoroot", utxoroot)
 blockid = T.challenge_bytes("id")
 ```
 
@@ -162,7 +162,7 @@ Given a sorted list of n unique inputs,
 D[n] = {d(0), d(1), ..., d(n-1)}
 ```
 
-the MPTH is thus defined as follows:
+the MPTH is thus defined as follows.
 
 The hash of an empty Merkle patricia tree list is a 32-byte challenge string with the label `patricia.empty`:
 
--- a/zkvm/docs/zkvm-blockchain.md
+++ b/zkvm/docs/zkvm-blockchain.md
@@ -10,6 +10,10 @@ The blockchain state contains:
 - `initialheader`: The initial block header (at height 1).
 - `tipheader`: The latest block header.
 - `utxos`: The IDs of all available [utxos](zkvm-spec.md#utxo).
+- `nonces`: The set of current [nonces](#nonce).
+- `refids`: The set of recent [block IDs](#block-id).
+  The size of this set is bounded by `block.header.refscount`.
+  Block IDs in this set may be referenced by nonces.
 
 # Block header
 
@@ -22,6 +26,9 @@ A block header contains:
   Each new block must have a time strictly later than the block before it.
 - `txroot`: 32-byte merkle root hash of the transactions in the block.
 - `utxoroot`: 32-byte merkle root hash of the utxo set after applying all transactions in the block.
+- `nonceroot`: 32-byte merkle root hash of the nonce set after applying all transactions in the block.
+- `refscount`: integer number of recent block IDs to store for reference.
+  A new block may specify a lower `refscount` than its predecessor but may not increase it by more than 1.
 
 # Block
 
@@ -56,7 +63,7 @@ Inputs:
 - `prevheader`, the header of the previous block.
 
 Output:
-- list of [txlog](zkvm-spec.md#transaction-log)+[txid](zkvm-spec.md#transaction-id) pairs,
+- list of [transaction logs](zkvm-spec.md#transaction-log),
   one for each transaction in block.txs.
 
 Procedure:
@@ -64,13 +71,13 @@ Procedure:
 2. Verify `block.header.height == prevheader.height+1`.
 3. Verify `block.header.previd` equals the [block ID](#block-id) of `prevheader`.
 4. Verify `block.header.timestamp > prevheader.timestamp`.
+5. Verify `block.header.refscount >= 0` and `block.header.refscount <= prevheader.refscount + 1`.
 5. For each transaction `tx` in block.txs:
    1. Verify `tx.mintime == 0` or `tx.mintime >= block.header.timestamp`.
    2. Verify `tx.maxtime == 0` or `tx.maxtime < block.header.timestamp`.
    3. If `block.header.version == 1`, verify `tx.version == 1`.
    4. [Execute](zkvm-spec.md#vm-execution) `tx` to produce transaction log `txlog`.
-   5. Compute [transaction ID](zkvm-spec.md#transaction-id) `txid` from `tx` and `txlog`.
-   6. Add the pair `txlog`+`txid` to the list of output pairs.
+   5. Add `txlog` to the list of output logs.
 
 # Apply block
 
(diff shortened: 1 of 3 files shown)


zkvm/src/vm.rs
```diff
@@ -29,10 +29,10 @@ pub struct TxHeader {
     /// Version of the transaction
     pub version: u64,
 
-    /// Timestamp before which tx is invalid (sec)
+    /// Timestamp before which tx is invalid (in milliseconds since the Unix epoch)
```
oleganza: I think I still prefer using seconds as less error-prone. We can bikeshed this here: https://github.com/interstellar/slingshot/issues/189

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,127 @@
...
+- `header`: a [block header](#block-header).
+- `txs`: a list of [transactions](zkvm-spec.md#transaction).
+
+# Block ID
+
+A block ID is computed from a [block header](#block-header) using the [transcript](zkvm-spec.md#transcript) mechanism:
+
+```
+T = Transcript("ZkVM.blockheader")
+T.commit(LE64(version))
```
oleganza: missing labels on each `commit` operartion

bobg:
This spec is still a little bit sparse, but the data types and procedures should be evaluated for completeness and correctness. Note that some parts of the spec implicitly rely on serializing data objects for which we do not yet have a serialization spec. I'm hoping we'll get that spec as part of https://github.com/interstellar/slingshot/pull/216 and that this will then be able to refer to that.

zkvm/src/vm.rs
oleganza: UPD: objection withdrawn

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+     empty set
+   - `nonces`:
+     empty set
+   - `refids`:
+     empty set
+
+## Make initial block
+
+Inputs:
+- `timestampms`,
```
oleganza:
```suggestion
- `timestamp_ms`,
```

otherwise sounds like a typo.

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+Nodes participating in a ZkVM blockchain network must implement the data types and perform the procedures described in this document.
+Specifically:
+
+- Each node maintains a [blockchain state](#blockchain-state).
+- A node creating a new ZkVM network performs the [start new network](#start-new-network) procedure.
+- A node joining an existing ZkVM network performance the [join existing network](#join-existing-network) procedure.
+- Each node performs the [apply block](#apply-block) procedure on the arrival of each new block.
+
+This document does not describe the consensus mechanism,
+which governs the production of authoritative blocks for the blockchain.
```
oleganza: wdym by "authoritative blocks"?

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+## Blockchain state
+
+The state of a ZkVM blockchain is given by the blockchain-state structure.
+Each node maintains a copy of this structure.
+As each new
+[block](#block)
+arrives,
+it is
+[applied](#apply-block)
+to the current state to produce an new,
```
oleganza: i'm not sure if this is a typo or an English gimmick
```suggestion
to the current state to produce a new,
```

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+updated state.
+
+The blockchain state contains:
+
+- `initialheader`:
+  The initial block header
+  (the header with height 1).
```
oleganza: I never liked the kr's argument that number 0 is useful to reserve for "nil" case. In Rust there's Option for exactly that, and we all know that indexing from 1 always strikes back.

oleganza:
```suggestion
  (the header with height 0).
```

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+[Merkle binary tree](zkvm-spec.md#merkle-binary-tree).
+Each node hashes the subtrees beneath it.
+The root node’s hash is a commitment to the full membership of the tree.
+It is possible to create and verify compact proofs of membership.
+
+Unlike a Merkle binary tree,
+a Merkle patricia tree is a radix tree
+(in which subtrees of a given node share a common prefix)
+with variable length branches that allow for efficient updates.
+It is therefore preferable to a Merkle binary tree for sets with churn
```
oleganza: i fail to parse this phrase:

>for sets with churn specifically the utxo set and the nonce set

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+Output:
+- New blockchain state.
+
+Procedure:
+1. Let `state′` be `state`.
+2. For each
+   [nonce entry](zkvm-spec.md#nonce-entry)
+   `n` in `txlog`:
+   1. Verify `n.blockid` is one of the following:
+      - An all-zero 32-byte string,
```
oleganza: we disallow all-zero blockids

zkvm/docs/zkvm-blockchain.md
bobg: In fact I think it was I who first lobbied for the initial block to have height 1 instead of 0, and ISTR several things falling nicely into place once we made that change. I can think of no situation from then until now in which it has "struck back." I still feel strongly that 1 is the right choice.

Remember that the spec should be language-agnostic.

zkvm/docs/zkvm-blockchain.md
bobg: Yes, that was confusing, sorry. Reworded.

zkvm/docs/zkvm-blockchain.md
bobg: Typo. Good 👁 !

zkvm/docs/zkvm-blockchain.md
bobg: This was also confusing, sorry again!

zkvm/docs/zkvm-blockchain.md
oleganza: >the spec should be language-agnostic.

I don't think that's a good argument — similar to LE/BE, most computers are LE and if we just make things BE, it's going to induce more confusion for everyone. Similarly, virtually all languages that matter today index things from 0, and things "fall in place" pretty nicely that way. Having height starting with 1 is an oddity that must be justified with a strong reason, not the other way around. If the only reason is that in Go there is no Option<T> type, while everywhere else there is (NULL in SQL DBs, JS/Ruby having null/nil, optionals in Rust/C++/Swift etc), it's a problem specific to Go and must be worked around there.

Can you remind what was improved by having the height start at 1?

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+mechanism:
+
+```
+T = Transcript("ZkVM.blockheader")
+T.commit("version", LE64(version))
+T.commit("height", LE64(height))
+T.commit("previd", previd)
+T.commit("timestamp_ms", LE64(timestamp_ms))
+T.commit("txroot", txroot)
+T.commit("utxoroot", utxoroot)
```
oleganza: `nonceroot` and `refscount` are missing.

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+  32-byte
+  [Merkle patricia root hash](#merkle-patricia-tree)
+  of the utxo set after applying all transactions in the block.
+- `nonceroot`:
+  32-byte
+  [Merkle patricia root hash](#merkle-patricia-tree)
+  of the nonce set after applying all transactions in the block.
+- `refscount`:
+  integer number of recent block IDs to store for reference.
+  A new block may specify a lower `refscount` than its predecessor but may not increase it by more than 1.
```
oleganza: let's add `ext` variable-length field that we can use to throw in extra fields later as a soft-fork? And also hash it into block-id below.

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+   - `refscount`:
+     `refscount`
+5. Return a block with `header` set to `header` and an empty `txs` list.
+
+## Join existing network
+
+A new node starts here when joining a running network.
+It must either:
+- obtain all historical blocks, [applying](#apply-block) them one by one to reproduce the latest [blockchain state](#blockchain-state); or
+- obtain a recent copy of the blockchain state `state` and begin applying blocks beginning at `state.tipheader.height+1`.
```
oleganza:
```suggestion
- obtain a recent copy of the blockchain state `state` from a trusted source (e.g. from another trusted node that already performed the full validation) and begin applying blocks beginning at `state.tipheader.height+1`.
```

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+
+Output:
+- a
+  [block](#block).
+
+Procedure:
+1. Compute `txroot`,
+   the root hash of an empty
+   [Merkle binary tree](zkvm-spec.md#merkle-binary-tree)
+   with label `transactions`.
```
oleganza: Those labels are repeated across the spec, maybe we need separate sections like we do with [Transaction ID](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#transaction-id) in ZkVM? 

```
## Transaction root

Transaction ID is defined as a [merkle hash](#merkle-binary-tree) of a list consisting of 
a [header entry](#header-entry) followed by all the entries from the [transaction log](#transaction-log):

    T = Transcript("ZkVM.txroot")
    txroot = MerkleHash(T, {tx[0], tx[1], ...})
```

Each transaction is committed to the [transcript](#transcript) via its [transaction ID](zkvm-spec.md#transaction-id):

    T.commit("tx", tx.id)

```

Note that for trees we need to specify how to commit the items.

Similarly, for `## UTXO root` and `## Nonce root`. And then we can xref to these sections throughout the doc.

File: zkvm/docs/zkvm-blockchain.md [hunk 13]
```diff
@@ -0,0 +1,406 @@
...
+2. Verify `block.header.height == prevheader.height+1`.
+3. Verify `block.header.previd` equals the
+   [block ID](#block-id)
+   of `prevheader`.
+4. Verify `block.header.timestamp_ms > prevheader.timestamp_ms`.
+5. Verify `block.header.refscount >= 0` and `block.header.refscount <= prevheader.refscount + 1`.
+6. Compute `txroot`, the [Merkle root hash](zkvm-spec.md#merkle-binary-tree) of `block.txs`, using label `"transactions"`.
+7. Verify `txroot == block.header.txroot`.
+8. For each transaction `tx` in block.txs:
+   1. Verify `tx.mintimems == 0` or `tx.mintimems >= block.header.timestamp_ms`.
```
oleganza:
```suggestion
   1. Verify `tx.mintime_ms <= block.header.timestamp_ms`.
```

from txvm spec:
image

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+   2. Verify `tx.maxtimems == 0` or `tx.maxtimems < block.header.timestamp_ms`.
```
oleganza:
```suggestion
   2. Verify `tx.maxtime_ms >= block.header.timestamp_ms`.
```

We allowed default-zero value in TxVM because we were using varints and were saving unspecified amount of "few bytes", so it was sort-of more convenient to allow 0 as "no limit".

In ZkVM there are no varints anywhere, so we can simplify things by using 0xffffffffffffffff as a default maxtime. It's taking up the same space as zero, and requires no special cases.

zkvm/docs/zkvm-blockchain.md
oleganza: TxVM defined the blockheader as a TxVM tuple with arbitrary number of extra fields, that were all hashed into block ID. Here we can use varlength string and extend it how we want later. The only caveat is that we'd need to hash it in as a whole for compatibility, even if we break it into separate fields in future updates.

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+
+Inputs:
+- `block`,
+  the block to apply.
+- `state`,
+  the current
+  blockchain state.
+
+Output:
+- New blockchain state.
```
oleganza:
```suggestion
- New blockchain state `state′`.
```

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+
+Procedure:
+1. [Validate](#validate-block)
+   `block` with `prevheader` set to `state.tipheader`.
+2. Let `state′` be `state`.
+3. Remove items from `state′.nonces` where the expiration timestamp is earlier than `block.header.timestamp_ms`.
```
oleganza:
```suggestion
3. Remove items from `state′.nonces` where the expiration timestamp is strictly less than `block.header.timestamp_ms`.
```

oleganza: wanted to err on the side of explicitness. Ideally we'd use operator `<` here.

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+4. For each txlog from the validation step,
+   in order:
+   1. [Apply](#apply-transaction-log)
```
oleganza: maybe make the whole phrase "Apply txlog" a xref so it matches the title of that section better?

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,406 @@
...
+3. For each
+   [input entry](zkvm-spec.md#input-entry)
+   or
+   [output entry](zkvm-spec.md#output-entry)
+   in `txlog`:
+   1. If an input entry,
+      verify its ID is in `state′.utxos`,
+      then remove it.
+   2. If an output entry,
+      add its utxo ID to `state′.utxos`.
```
oleganza: Maybe add a note that added utxos may be immediately consumed in the _same transaction_ or in the _same block_. So no one tries to remove and add in bulk, without annihilating such local pairs first.

zkvm/docs/zkvm-blockchain.md
bobg: Nice catch, thanks.

zkvm/docs/zkvm-blockchain.md
bobg: Good call.

zkvm/docs/zkvm-blockchain.md
bobg: OOPS 😅

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,386 @@
...
+   [block ID](#block-id)
+   of `prevheader`.
+5. Verify `block.header.timestamp_ms > prevheader.timestamp_ms`.
+6. Verify `block.header.refscount >= 0` and `block.header.refscount <= prevheader.refscount + 1`.
+7. [Compute txroot](#compute-txroot) from `block.txs`.
+8. Verify `txroot == block.header.txroot`.
+9. For each transaction `tx` in block.txs:
+   1. Verify `tx.mintime_ms <= block.header.timestamp_ms <= tx.maxtime_ms`.
+   2. If `block.header.version == 1`,
+      verify `tx.version == 1`.
```
oleganza: Also verify that `ext` is empty to enforce that it's not abused until there's a version upgrade.

bobg: See step 2 above.

oleganza: 👍

zkvm/docs/zkvm-blockchain.md
```diff
@@ -0,0 +1,386 @@
...
+This is the specification for the ZkVM blockchain,
+a blockchain containing
+[ZkVM transactions](zkvm-spec.md).
+
+Nodes participating in a ZkVM blockchain network must implement the data types and perform the procedures described in this document.
+Specifically:
+
+- Each node maintains a [blockchain state](#blockchain-state).
+- A node creating a new ZkVM network performs the [start new network](#start-new-network) procedure.
+- A node joining an existing ZkVM network performance the [join existing network](#join-existing-network) procedure.
```
cathieyun:
```suggestion
- A node joining an existing ZkVM network performs the [join existing network](#join-existing-network) procedure.
```

zkvm/docs/zkvm-blockchain.md
oleganza: Bitcoin uses height=0: https://blockstream.info/block/000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
(as well as its codebase forks like Zcash, Monero etc)

Ethereum uses height=0: https://etherscan.io/block/0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3

Stellar uses height=1: https://www.stellar.org/developers/guides/concepts/ledger.html

Tezos uses height=0: https://tzscan.io/0

Cosmos uses height=1 for first proposed block, because 0th is a "genesis file": https://www.mintscan.io/blocks/1

oleganza: As someone who knows Cosmos well told me in private conversation: 

> me: do you know why cosmos initial block has height=1, not 0?
> they: Zeroth block is the genesis file
> me: genesis file is somewhat special, or it's just a "baked in block"?
> they: it's special. Just json but I believe you can query the node for height zero state and get genesis info

oleganza: I suggest we use height=0. Another way to look at this: this is the only block that's implicitly agreed upon by everyone, while all the other blocks (with actual transactions) are subject to consensus protocol.

oleganza: i don't care - only delta matters in the end of the day.

PR #199, opened 2019-03-06
zkvm: Implement `unblind` Instruction

### LINKED ISSUE #157: zkvm: implement `unblind` instruction
[`unblind`](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#unblind) is a simple instruction that checks that a given expression has a specific cleartext value and a corresponding unblinded Pedersen commitment for that value.

decentralisedkev:
Closes #157

### DIFF
--- a/zkvm/src/vm.rs
+++ b/zkvm/src/vm.rs
@@ -331,6 +331,42 @@ where
         Ok(())
     }
 
+    fn unblind(&mut self) -> Result<(), VMError> {
+        // Pop expression `expr`
+        let expr = self.pop_item()?.to_expression()?;
+        
+        // Pop commitment `V`
+        let v_commitment = self.pop_item()?.to_data()?.to_commitment()?;
+        let v_point = v_commitment.to_point();
+        
+        // Pop scalar `v`
+        let data = self.pop_item()?.to_data()?.to_bytes();
+        let v_scalar = SliceReader::parse(&data, |r| r.read_scalar())?;
+
+        // Create a detached variable `var` using `V`
+        let var = Variable { commitment: v_commitment };
+
+
+        self.delegate.verify_point_op(|| {
+            // Check V = vB => V-vB = 0
+            PointOp {
+                primary: Some(v_scalar),
+                secondary: None,
+                arbitrary: vec![(Scalar::one(), v_point)],
+            }
+        })?;
+
+        // Check var == expr
+        let expr2 = self.variable_to_expression(var.clone())?;
+
+        // self.delegate
+        //     .cs()
+        //     .constrain(expr.to_r1cs_lc() - expr2.to_r1cs_lc());
+
+        self.push_item(var);
+        Ok(())
+    }
+
     fn r#const(&mut self) -> Result<(), VMError> {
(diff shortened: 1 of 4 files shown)


zkvm/src/vm.rs
```diff
@@ -331,6 +331,38 @@ where
...
+                secondary: None,
+                arbitrary: vec![(Scalar::one(), v_point)],
+            }
+        })?;
+
+        // Add constraint `V == expr`
+        let (_, v) = self.delegate.commit_variable(&v_commitment)?;
+        self.delegate.cs().constrain(expr.to_r1cs_lc() - v);
+
+        // Push detached variable
```
oleganza: there is no longer "attached/detached" terminology.

zkvm/src/vm.rs
```diff
@@ -331,6 +331,38 @@ where
...
+        let v_point = v_commitment.to_point();
+
+        // Pop scalar `v`
+        let data = self.pop_item()?.to_data()?.to_bytes();
+        let v_scalar = SliceReader::parse(&data, |r| r.read_scalar())?;
+
+        self.delegate.verify_point_op(|| {
+            // Check V = vB => V-vB = 0
+            PointOp {
+                primary: Some(v_scalar),
```
```suggestion
                primary: Some(-v_scalar),
```

zkvm/src/vm.rs
```diff
@@ -331,6 +331,38 @@ where
...
+        // Pop expression `expr`
+        let expr = self.pop_item()?.to_expression()?;
+
+        // Pop commitment `V`
+        let v_commitment = self.pop_item()?.to_data()?.to_commitment()?;
+        let v_point = v_commitment.to_point();
+
+        // Pop scalar `v`
+        let data = self.pop_item()?.to_data()?.to_bytes();
+        let v_scalar = SliceReader::parse(&data, |r| r.read_scalar())?;
```
oleganza: We already have `Data::Scalar` case, so it would make sense to do:

```rust
        let scalar_witness = self.pop_item()?.to_data()?.to_scalar()?;
        let v_scalar = scalar_witness.to_scalar();
```

where `Data::to_scalar` would either parse Data::Opaque, or return the contents of Data::Scalar (which is ScalarWitness, so we'd need to convert _that_ to raw scalar again).

And we do that, we should use this scalar witness in `fn r#const` too.

zkvm/src/vm.rs
```diff
@@ -331,9 +331,41 @@ where
...
+            commitment: v_commitment,
+        });
+        Ok(())
+    }
+
     fn r#const(&mut self) -> Result<(), VMError> {
-        let data = self.pop_item()?.to_data()?.to_bytes();
-        let scalar = SliceReader::parse(&data, |r| r.read_scalar())?;
+        let scalar_witness = self.pop_item()?.to_data()?.to_scalar()?;
+        let scalar = scalar_witness.to_scalar();
```
oleganza: This line kills the integer witness. We need to pass ScalarWitness as is to Expression, to preserve its structure (integerness).

```suggestion
```

decentralisedkev: ahh right, sorry missed that

PR #200, opened 2019-03-06
zkvm/merkle: factor out merkle proofs into separate module

### LINKED ISSUE #197: zkvm: factor out binary merkle tree from txid computation
Currently merkle tree logic is tied to txid/txlog computation. We'll need it for blockchain tree-of-transactions commitments (#93) too.

We can replace slices `&[Entry]` with generic slices `&[I]` where `I: MerkleItem`, with `trait MerkleItem` having method to write contents into the `&mut Transcript`: 

```rust
trait MerkleItem {
    fn commit(&self, t: &mut Transcript);
}
```

We can then implement `MerkleItem` for `txlog::Entry` and also implement it for `u64` inside the tests to make them decoupled from txid.

vickiniu:
Fixes #197

### DIFF
--- a/zkvm/src/merkle.rs
+++ b/zkvm/src/merkle.rs
@@ -3,18 +3,20 @@ use subtle::ConstantTimeEq;
 
 use crate::errors::VMError;
 
+/// MerkleItem defines an item in the Merkle tree.
 pub trait MerkleItem: Sized {
+    /// Commits the hash of the item to Transcript.
     fn commit(&self, t: &mut Transcript);
 }
 
-/// MerkleNeighbor represents a step in a Merkle proof of inclusion.
+/// MerkleNeighbor is a step in a Merkle proof of inclusion.
 #[derive(Copy, Clone, PartialEq, Debug)]
 pub enum MerkleNeighbor {
     Left([u8; 32]),
     Right([u8; 32]),
 }
 
-/// MerkleTree represents a Merkle tree of hashes with a given size.
+/// Merkle tree of hashes with a given size.
 pub struct MerkleTree {
     size: usize,
     label: &'static [u8],
@@ -28,10 +30,7 @@ enum MerkleNode {
 
 impl MerkleTree {
     /// Constructs a new MerkleTree based on the input list of entries.
-    pub fn new<M>(label: &'static [u8], list: &[M]) -> Option<MerkleTree>
-    where
-        M: MerkleItem,
-    {
+    pub fn new<M: MerkleItem>(label: &'static [u8], list: &[M]) -> Option<Self> {
         if list.len() == 0 {
             return None;
         }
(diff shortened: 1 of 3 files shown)


zkvm/src/merkle.rs
```diff
@@ -0,0 +1,253 @@
...
+}
+
+enum MerkleNode {
+    Leaf([u8; 32]),
+    Node([u8; 32], Box<MerkleNode>, Box<MerkleNode>),
+}
+
+impl MerkleTree {
+    /// Constructs a new MerkleTree based on the input list of entries.
+    pub fn new<M>(label: &'static [u8], list: &[M]) -> Option<MerkleTree>
```
oleganza: nit:
```suggestion
    pub fn new<M>(label: &'static [u8], list: &[M]) -> Option<Self>
```

zkvm/src/txlog.rs
```diff
@@ -66,43 +47,15 @@ impl UTXO {
 impl TxID {
     /// Computes TxID from a tx log
     pub fn from_log(list: &[Entry]) -> Self {
-        let t = Transcript::new(b"ZkVM.txid");
```
oleganza: this was computing the hash almost in-place, w/o allocating all boxed nodes in memory. I think we should have a separate entry point to MerkleTree to retain this lightweight computation. Something like `MerkleTree::root(list)` in addition to `MerkleTree::build(list)` ("build" probably better reflects the "heaviness" of the structure than a typical `::new`).

oleganza: obviously, `MerkleTree::root()` is a redundant version of `MerkleTree::build().root()`, but it's doing no allocations and then can be even rewritten to be iterative instead of recursive.

zkvm/src/merkle.rs
oleganza: 👍 to the `label` argument!

zkvm/src/merkle.rs
```diff
@@ -0,0 +1,253 @@
...
+        if index >= self.size {
+            return Err(VMError::InvalidMerkleProof);
+        }
+        let t = Transcript::new(self.label);
+        let mut result = Vec::new();
+        self.root.subproof(t, index, self.size, &mut result);
+        Ok(result)
+    }
+
+    pub fn verify_proof<M>(
```
oleganza: documentation missing

zkvm/src/merkle.rs
```diff
@@ -0,0 +1,253 @@
+use merlin::Transcript;
+use subtle::ConstantTimeEq;
+
+use crate::errors::VMError;
+
+pub trait MerkleItem: Sized {
```
oleganza: trait documentation missing

zkvm/src/merkle.rs
```diff
@@ -0,0 +1,253 @@
...
+    fn commit(&self, t: &mut Transcript);
```
oleganza: trait method documentation missing

zkvm/src/merkle.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    pub fn verify_proof<M>(
+        label: &'static [u8],
+        entry: &M,
+        proof: Vec<MerkleNeighbor>,
+        root: &[u8; 32],
+    ) -> Result<(), VMError>
+    where
+        M: MerkleItem,
```
oleganza: can move this M: MerkleItem into `verify_proof<M: MerkleItem>` for succinctness

zkvm/src/merkle.rs
```diff
@@ -0,0 +1,249 @@
...
+        let t = Transcript::new(label);
+        Some(MerkleTree {
+            size: list.len(),
+            label,
+            root: Self::build_tree(t, list),
+        })
+    }
+
+    /// Builds a proof of inclusion for entry at the given index for the Merkle tree.
+    pub fn proof(&self, index: usize) -> Result<Vec<MerkleNeighbor>, VMError> {
```
oleganza: How about `path_to(i)` for this method and `verify_path` instead of `verify_proof`? So we use a more concrete term "path" instead of more abstract "proof" (because checking membership of all the items is also a proof, but not a path proof).

PR #202, opened 2019-03-07
keytree: derive intermediate pubkey

tessr:
This is my first stab at this function. A few outstanding questions: 

1. This function returns an Option because it calls a function that returns an Option and I wasn't sure what else to do.
2. Currently it only accepts Xpubs; I'd love a pointer on how to make it accept Xprvs too. I'm guessing something with generics? _has never see a generic before, :gopher:_
3. This note about not calling the Transcript directly has me a tad spooked: https://doc.dalek.rs/merlin/struct.Transcript.html#note
4. Is the function signature even right here? Among other things, I made it take a label which is a `&'static std::string::String` to satisfy some business about lifetimes.

### DIFF
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -135,7 +135,9 @@ mod tests {
         let seed = [0u8; 32];
         let mut rng = ChaChaRng::from_seed(seed);
         let xprv = Xprv::random(&mut rng);
-        let xpub = xprv.to_xpub().to_intermediate_key(customize);
+        let xpub = xprv.to_xpub().to_intermediate_key(|t| {
+            t.commit_u64("account_id", 42);
+        });
 
         // the following are hard-coded based on the previous seed
         let expected_dk = [
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -48,6 +48,18 @@ impl Xprv {
     }
 }
 
+pub fn intermediate_key(xpub: &Xpub, label: &String, data: &String) -> Xpub {
+    // question: is the following kosher? https://doc.dalek.rs/merlin/struct.Transcript.html#note
+    let mut t = Transcript::new(b"Keytree.intermediate");
+    t.commit_bytes(b"pt", xpub.point.as_bytes());
+    t.commit_bytes(b"dk", &xpub.dk);
+
+    // provide the transcript to the user to commit an arbitrary derivation path or index
+    t.commit_bytes(label.as_bytes(), data.as_bytes());
+
+    unimplemented!();
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -1,11 +1,15 @@
 #![deny(missing_docs)]
 //! Implementation of the key tree protocol, a key blinding scheme for deriving hierarchies of public keys.
 
+use crate::transcript::TranscriptProtocol;
 use curve25519_dalek::constants;
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::ristretto::RistrettoPoint;
 use curve25519_dalek::scalar::Scalar;
+use merlin::Transcript;
 use rand::{CryptoRng, RngCore};
+
+mod transcript;
 
 /// Xprv represents an extended private key.
 pub struct Xprv {
@@ -48,39 +52,33 @@ impl Xprv {
     }
 }
 
-/// Returns a intermediate child pubkey.
-pub fn intermediate_key(
-    xpub: &Xpub,
-    label: &'static std::string::String,
-    data: &String,
-) -> Option<Xpub> {
-    // question: this function returns an option because it calls a function that returns an option
-    // and I didn't know what else to do. is there something better to do?
+impl Xpub {
+    /// Returns a intermediate child pubkey.
+    pub fn to_intermediate_key(&self, customize: impl FnOnce(&mut Transcript)) -> Xpub {
+        let mut t = Transcript::new(b"Keytree.derivation");
+        t.commit_bytes(b"pt", self.precompressed_pubkey.as_bytes());
+        t.commit_bytes(b"dk", &self.dk);
 
-    // question: is the following kosher? https://doc.dalek.rs/merlin/struct.Transcript.html#note
-    let mut t = Transcript::new(b"Keytree.intermediate");
-    t.commit_bytes(b"pt", xpub.point.as_bytes());
-    t.commit_bytes(b"dk", &xpub.dk);
+        // the user can pass customize in so that they can commit arbitrary
+        // data (like an account ID)
+        customize(&mut t);
 
-    // provide the transcript to the user to commit an arbitrary derivation path or index
-    t.commit_bytes(label.as_bytes(), data.as_bytes());
+        // squeeze a challenge scalar
+        let f = t.challenge_scalar(b"f.intermediate");
 
-    // squeeze a challenge scalar
-    let f = t.challenge_scalar(label.as_bytes()); // is this the right input?
+        // squeeze a new derivation key
+        let mut dk = [0u8; 32];
+        t.challenge_bytes(b"dk", &mut dk);
 
-    // squeeze a new derivation key
-    let mut dk = [0u8; 32];
-    t.challenge_bytes(b"dk", &mut dk);
+        let point = self.point + (f * &constants::RISTRETTO_BASEPOINT_POINT);
+        let precompressed_pubkey = point.compress();
 
-    let point;
-    match xpub.point.decompress() {
-        None => return None,
-        Some(p) => {
-            point = (p + (f * &constants::RISTRETTO_BASEPOINT_POINT)).compress();
+        Xpub {
+            point,
+            dk,
+            precompressed_pubkey,
         }
     }
-
-    Some(Xpub { point, dk })
 }
 
 #[cfg(test)]
@@ -129,5 +127,33 @@ mod tests {
 
         assert_eq!(xpub.dk, expected_dk);
         assert_eq!(xpub.point, expected_point);
+        assert_eq!(xpub.precompressed_pubkey, expected_compressed_point);
+    }
+
+    #[test]
+    fn random_xpub_derivation_test() {
+        let seed = [0u8; 32];
+        let mut rng = ChaChaRng::from_seed(seed);
+        let xprv = Xprv::random(&mut rng);
+        let xpub = xprv.to_xpub().to_intermediate_key(customize);
+
+        // the following are hard-coded based on the previous seed
+        let expected_dk = [
+            213, 184, 36, 41, 212, 147, 104, 136, 121, 56, 43, 216, 7, 41, 62, 96, 27, 59, 169,
+            220, 232, 135, 119, 40, 211, 34, 90, 46, 86, 167, 236, 82,
+        ];
+        let expected_compressed_point = CompressedRistretto::from_slice(&[
+            144, 115, 113, 51, 15, 159, 80, 246, 172, 113, 178, 147, 234, 47, 12, 31, 1, 203, 103,
+            10, 174, 22, 233, 179, 59, 228, 17, 206, 65, 206, 151, 3,
+        ]);
+        let expected_point = expected_compressed_point.decompress().unwrap();
+
+        assert_eq!(xpub.dk, expected_dk);
+        assert_eq!(xpub.point, expected_point);
+        assert_eq!(xpub.precompressed_pubkey, expected_compressed_point);
+    }
+
+    fn customize(t: &mut Transcript) {
+        t.commit_bytes(b"test_label", b"test_content");
     }
 }
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -53,14 +53,14 @@ impl Xprv {
 }
 
 impl Xpub {
-    /// Returns a intermediate child pubkey.
+    /// Returns a intermediate child pubkey. Users must provide customize, in order to separate
+    /// sibling keys from one another through unique derivation paths.
     pub fn derive_intermediate_key(&self, customize: impl FnOnce(&mut Transcript)) -> Xpub {
         let mut t = Transcript::new(b"Keytree.derivation");
         t.commit_bytes(b"pt", self.precompressed_pubkey.as_bytes());
         t.commit_bytes(b"dk", &self.dk);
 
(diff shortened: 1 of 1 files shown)


keytree/src/lib.rs
```diff
@@ -37,6 +41,41 @@ impl Xprv {
...
+
+    // squeeze a challenge scalar
+    let f = t.challenge_scalar(label.as_bytes()); // is this the right input?
+
+    // squeeze a new derivation key
+    let mut dk = [0u8; 32];
+    t.challenge_bytes(b"dk", &mut dk);
+
+    let point;
+    match xpub.point.decompress() {
```
oleganza: We can avoid failable decoding by pushing the decoding up the API stack:

1. Xpub can hold decompressed `RistrettoPoint`.
2. Xprv->Xpub conversion does not fail.
3. bytes -> Xpub may fail because it will have to decompress the point.
4. To avoid overhead of recompressing the point, Xpub could also _own_ the transcript instance precomputed with `pt` and `dk`. Then, when we need to derive things, we'd clone the transcript, customize it with user's data and then squeeze the factor. When converting 64 bytes into Xpub, this precomputation would happen with an already compressed point (which is == the first 32 bytes).
5. To reuse the transcript for both intermediate and leaf cases, I suggest changing the spec: the label will be the same for both uses: `"Keytree.derive"`, but domain separation will happen when we squeeze `f`: `"f.intermediate"`  and `"f.leaf"` respectively.
6. As a result, derivation functions will be non-failable.

keytree/src/lib.rs
```diff
@@ -37,6 +41,41 @@ impl Xprv {
...
+/// Returns a intermediate child pubkey.
+pub fn intermediate_key(
+    xpub: &Xpub,
+    label: &'static std::string::String,
+    data: &String,
+) -> Option<Xpub> {
+    // question: this function returns an option because it calls a function that returns an option
+    // and I didn't know what else to do. is there something better to do?
+
+    // question: is the following kosher? https://doc.dalek.rs/merlin/struct.Transcript.html#note
```
oleganza: the note means that the ZKP protocol should be composable with others. This derivation logic is not a ZKP protocol and is not meant to be composed with the externally-provided transcript. However, it is composable with the user's choice of "derivation path", by letting the user commit arbitrary structured data to the transcript in-between Keytree committing the xpub (`pt`,`dk`) and squeezing the blinding factor `f`.

keytree/src/lib.rs
```diff
@@ -37,6 +41,41 @@ impl Xprv {
     }
 }
 
+/// Returns a intermediate child pubkey.
+pub fn intermediate_key(
+    xpub: &Xpub,
+    label: &'static std::string::String,
```
oleganza: Instead of passing a label, we should pass a closure `F: FnOnce(&mut Transcript)` where the user can commit arbitrary key-value pairs as a "derivation path". Note: the user may commit more than one labeled input.

oleganza: The closure is supposed to be called by keytree in-between committing the xpub (`pt`,`dk`) and squeezing the blinding factor `f`.

keytree/src/lib.rs
```diff
@@ -37,6 +41,41 @@ impl Xprv {
...
+    // question: this function returns an option because it calls a function that returns an option
+    // and I didn't know what else to do. is there something better to do?
+
+    // question: is the following kosher? https://doc.dalek.rs/merlin/struct.Transcript.html#note
+    let mut t = Transcript::new(b"Keytree.intermediate");
+    t.commit_bytes(b"pt", xpub.point.as_bytes());
+    t.commit_bytes(b"dk", &xpub.dk);
+
+    // provide the transcript to the user to commit an arbitrary derivation path or index
+    t.commit_bytes(label.as_bytes(), data.as_bytes());
```
oleganza: Call the closure here:

```suggestion
    customize(&mut t);
```

keytree/src/lib.rs
oleganza: Private function `Xpub::from_parts(RistrettoPoint, [u8;32])` can be used by Xprv->Xpub conversion and by derivation function.

oleganza:
>I'd love a pointer on how to make it accept Xprvs too.

Xprv simply has the similar function signature for deriving intermediate and leaf keys, but returns Xprv and Scalar respectively (Xpub derives Xpub and CompressedRistretto). Common derivation logic can be factored out into private helper methods/functions.

There is no use of generics apart from closure (`FnOnce`) for derivation path customization (see notes above).

keytree/src/lib.rs
tessr: I have done the very first part of this at https://github.com/interstellar/slingshot/pull/207

tessr: Sorry I've been going through this so slowly. Another question: 

are you suggesting that we add another field to the xpub for the transcript, and when the xpub is generated from the xprv, we also produce a new transcript instance using that xpub’s point and derivation key?

oleganza: >are you suggesting that we add another field to the xpub for the transcript, 

Yes. `Xpub {point, dk, transcript}`

>and when the xpub is generated from the xprv, we also produce a new transcript instance using that xpub’s point and derivation key?

Not quite: simply clone xprv.transcript to assign it to xpub.transcript.

oleganza: ^ this is because the transcript holds the derivation that's the same for xpubs and xprvs, so it must be exactly the same in xpub and xprv

tessr: Ah, ok, so we also store the transcript on Xprv? And we create that instance of the transcript when we create the Xprv?

oleganza: yup!

oleganza: And it's supposed to be precomputed:

```
t = Transcript::new()
t.commit(pub)
t.commit(dk)
Xpub.transcript = t
```

oleganza: so that when you derive a child key, you can simply do:

```
t = self.transcript.clone()
t.commit(user provided derivation path)
f = t.challenge_scalar("f.leaf") // or "f.intermediate"
...
```

tessr: In this pseudo code: 

```
t = Transcript::new()
t.commit(pub)
t.commit(dk)
Xpub.transcript = t
```

Did you mean to set the transcript on the Xprv instead? And if so, what are you committing? Finally, I assume you need to use `commit_bytes` to commit things to the transcript.... unless we should do the closure trick you mentioned above ^ ?

oleganza: sorry for not being clear. 

When you initialize _either_ Xpub, or Xprv, they both need to instantiate the same precomputed transcript.

* the `commit(pub)` is actually `commit_point("pt", point.compress())`, point is computed on the fly for Xprv (from `scalar*B`) and for Xpub it's already there.
* the `commit(dk)` is actually `commit_bytes("dk", &dk)`

Then, the transcript is stored inside the structure (Xpub or Xprv).

When you derive Xpub from Xprv, the transcript is already good, you just need to compute point `xpub.point = xprv.scalar*B` and clone the transcript.

oleganza: edited ^

oleganza: When you derive a child, you use a closure on Xprv/Xpub to which you clone `t=self.transcript.clone()` and provide `&mut t` to let user customize it with derivation path.

After that, you squeeze `f` and (if it's an intermediate key) `dk2` from that transcript. Then, you instantiate a new (child) Xprv/Xpub and precompute its own transcript from scratch using its pubkey (adjusted by `f`) and `dk2`.

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
     }
 }
 
+impl Xpub {
+    /// Returns a intermediate child pubkey.
+    pub fn to_intermediate_key(&self, customize: impl FnOnce(&mut Transcript)) -> Xpub {
```
oleganza: `to_***` is used for converting type A to type B. Here we do not convert, we derive child keys, leaving the parent happily alone.

Suggestion:
```suggestion
    pub fn derive_intermediate_key(&self, customize: impl FnOnce(&mut Transcript)) -> Xpub {
```

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
+
+        // the user can pass customize in so that they can commit arbitrary
+        // data (like an account ID)
+        customize(&mut t);
+
+        // squeeze a challenge scalar
+        let f = t.challenge_scalar(b"f.intermediate");
+
+        // squeeze a new derivation key
+        let mut dk = [0u8; 32];
```
oleganza: maybe call it `dk2` or `child_dk` to disambiguate from `self.dk`?
```suggestion
        let mut child_dk = [0u8; 32];
```

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
+        t.challenge_bytes(b"dk", &mut dk);
+
+        let point = self.point + (f * &constants::RISTRETTO_BASEPOINT_POINT);
```
oleganza: same comment as above:
```suggestion
        let child_point = self.point + (f * &constants::RISTRETTO_BASEPOINT_POINT);
```

keytree/src/lib.rs
```diff
@@ -94,5 +127,33 @@ mod tests {
...
         assert_eq!(xpub.point, expected_point);
+        assert_eq!(xpub.precompressed_pubkey, expected_compressed_point);
+    }
+
+    #[test]
+    fn random_xpub_derivation_test() {
+        let seed = [0u8; 32];
+        let mut rng = ChaChaRng::from_seed(seed);
+        let xprv = Xprv::random(&mut rng);
+        let xpub = xprv.to_xpub().to_intermediate_key(customize);
```
oleganza: the usage is going to be mostly via a closure, so maybe something like this is more instructive:

```suggestion
        let xpub = xprv.to_xpub().to_intermediate_key(|t| {
            t.commit_u64("account_id", 42);
        });
```

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
+        let mut dk = [0u8; 32];
+        t.challenge_bytes(b"dk", &mut dk);
+
+        let point = self.point + (f * &constants::RISTRETTO_BASEPOINT_POINT);
+        let precompressed_pubkey = point.compress();
+
+        Xpub {
+            point,
+            dk,
+            precompressed_pubkey,
```
oleganza: maybe it's more readable if we don't declare one more variable, but simply say 

```suggestion
            precompressed_pubkey: point.compress(),
```

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
+
+        // squeeze a new derivation key
+        let mut child_dk = [0u8; 32];
+        t.challenge_bytes(b"dk", &mut dk);
+
+        let child_point = self.point + (f * &constants::RISTRETTO_BASEPOINT_POINT);
+        let precompressed_pubkey = point.compress();
+
+        Xpub {
+            point,
```
oleganza: `child_point`?

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
+
+        // squeeze a challenge scalar
+        let f = t.challenge_scalar(b"f.intermediate");
+
+        // squeeze a new derivation key
+        let mut child_dk = [0u8; 32];
+        t.challenge_bytes(b"dk", &mut dk);
+
+        let child_point = self.point + (f * &constants::RISTRETTO_BASEPOINT_POINT);
+        let precompressed_pubkey = point.compress();
```
oleganza: `child_point`?

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
+
+        Xpub {
+            point,
+            dk,
+            precompressed_pubkey: point.compress(),
```
oleganza: `child_point`?

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
+        // squeeze a new derivation key
+        let mut child_dk = [0u8; 32];
+        t.challenge_bytes(b"dk", &mut dk);
+
+        let child_point = self.point + (f * &constants::RISTRETTO_BASEPOINT_POINT);
+        let precompressed_pubkey = point.compress();
+
+        Xpub {
+            point,
+            dk,
```
oleganza: `child_dk`?

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
+            precompressed_pubkey: point.compress(),
+        }
+    }
```
oleganza: sorry

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
+}
```
oleganza: sorry again

keytree/src/lib.rs
```diff
@@ -94,5 +127,35 @@ mod tests {
...
+            10, 174, 22, 233, 179, 59, 228, 17, 206, 65, 206, 151, 3,
+        ]);
+        let expected_point = expected_compressed_point.decompress().unwrap();
+
+        assert_eq!(xpub.dk, expected_dk);
+        assert_eq!(xpub.point, expected_point);
+        assert_eq!(xpub.precompressed_pubkey, expected_compressed_point);
+    }
+
+    fn customize(t: &mut Transcript) {
```
oleganza: no need for this one if we use closures

keytree/src/lib.rs
```diff
@@ -48,6 +52,35 @@ impl Xprv {
...
 }
 
+impl Xpub {
+    /// Returns a intermediate child pubkey.
+    pub fn derive_intermediate_key(&self, customize: impl FnOnce(&mut Transcript)) -> Xpub {
+        let mut t = Transcript::new(b"Keytree.derivation");
+        t.commit_bytes(b"pt", self.precompressed_pubkey.as_bytes());
+        t.commit_bytes(b"dk", &self.dk);
+
+        // the user can pass customize in so that they can commit arbitrary
```
oleganza: i'd change this comment to reflect better that the "derivation path" is provided only through this closure. Meaning, it's not something that the user "can use" as an optionality, but "must use" in order to separate siblings from each other. I don't know what's the best way to put it yet.

tessr: I added something to the function-level comment. Is that better?

oleganza: Yes!

PR #204, opened 2019-03-08
zkvm/patricia: implement Merkle patricia tree

vickiniu:
Fixes #204

### DIFF
--- a/zkvm/src/patricia.rs
+++ b/zkvm/src/patricia.rs
@@ -1,8 +1,7 @@
-use merlin::Transcript;
+use std::cmp::min;
 use subtle::ConstantTimeEq;
 
 use crate::errors::VMError;
-// use crate::merkle::MerkleItem;
 
 pub struct PatriciaTree {
     root: PatriciaNode,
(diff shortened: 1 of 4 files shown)


zkvm/src/patricia.rs
```diff
@@ -0,0 +1,520 @@
...
+    /// Left (bit 0) and right (bit 1) nodes
+    children: Option<(Box<Node>, Box<Node>)>,
+}
+
+#[derive(Clone)]
+struct PatriciaItem {
+    /// Key contained at this item: with the appropriate
+    /// bits included, represents the data represented
+    /// by this path of the Patricia tree.
+    key: &'static [u8],
```
oleganza: `'static` is not going to work with actual data. Since all our use cases are for 32-byte keys, we can hardcode the key as `[u8;32]`.

zkvm/src/patricia.rs
```diff
@@ -0,0 +1,520 @@
...
+
+    /// Position of the last bit to include in the key, starting
+    /// from the most-significant bit. An offset of 7 includes
+    /// the whole bit
+    bits: u8,
```
oleganza: The semantics of this are not clear. Seems like the `key`'s slice contains the portion of the bitslice, while this field contains the remainder within a partial byte. Then maybe we need a borrowed slice, not an array? (with non-static lifetime).

vickiniu: My mistake, that is unclear! `key` is `>=` the whole bitslice (for this node), and `bits` indicates the position of the last bit to read in the `key`

oleganza:
We are most likely moving onto utreexo design and using a simple binary merkle tree as a result, so the prefix tree is not going to be needed (See #249, #258, #271).

PR #207, opened 2019-03-08
keytree: store decompressed ristretto point on Xpub

tessr:


keytree/src/lib.rs
```diff
@@ -75,10 +76,14 @@ mod tests {
             159, 7, 231, 190, 85, 81, 56, 122, 152, 186, 151, 124, 115, 45, 8, 13, 203, 15, 41,
             160, 72, 227, 101, 105, 18, 198, 83, 62, 50, 238, 122, 237,
         ];
-        let expected_point = CompressedRistretto::from_slice(&[
+        let expected_compressed_point = CompressedRistretto::from_slice(&[
             156, 102, 163, 57, 200, 52, 79, 146, 47, 195, 32, 108, 181, 218, 232, 20, 165, 148,
             192, 23, 125, 211, 35, 92, 37, 77, 156, 64, 154, 101, 184, 8,
         ]);
+        let expected_point = match expected_compressed_point.decompress() {
+          None => panic!("should be a valid compressed point"),
```
oleganza: Since you know that it's not going to fail (you just created a valid point yourself, not received it from elsewhere), you can simplify the annoying error handling with `.decompress().unwrap()` which is totally kosher in tests and helps keeping the logic cleaner.

oleganza:
an annoying `cargo fmt`:

image

keytree/src/lib.rs
tessr: TIL! thank you.

tessr:
OK, all fixed, PTAL 🍨

PR #208, opened 2019-03-08
zkvm: implement `alloc`

### LINKED ISSUE #150: zkvm: implement `alloc` instruction
[`alloc`](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#alloc) instruction creates a free variable in R1CS without an individual Pedersen commitment.

1. Change `Instruction::Alloc` to `Instruction::Alloc(Option<ScalarWitness>)` to let the prover specify the secret assignment.
2. When parsing bytecode (on the verifier's end) use `Alloc(None)`.
3. Implement `alloc` per [specification](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#alloc) by allocating a low-level variable in R1CS and using the provided optional assignment.

debnil:
Implements alloc instruction. Closes #150.


zkvm/src/ops.rs
```diff
@@ -241,7 +242,13 @@ impl Instruction {
             }
             Instruction::Const => write(Opcode::Const),
             Instruction::Var => write(Opcode::Var),
-            Instruction::Alloc => write(Opcode::Alloc),
+            Instruction::Alloc(osw) => {
+                write(Opcode::Alloc);
+                match osw {
```
oleganza: The assignment is the private data held by the prover, it's not part of the bytecode. Only the Opcode::Alloc needs to be written to the bytecode string.

zkvm/src/ops.rs
```diff
@@ -241,7 +242,13 @@ impl Instruction {
             }
             Instruction::Const => write(Opcode::Const),
             Instruction::Var => write(Opcode::Var),
-            Instruction::Alloc => write(Opcode::Alloc),
+            Instruction::Alloc(osw) => {
```
```suggestion
            Instruction::Alloc(_) => {
```

PR #209, opened 2019-03-09
utils: add pre-commit hook checking cargo-fmt

oleganza:
This adds a folder `utils` for development tools we use.

To check `cargo fmt` automatically before you commit, do this in the root of your copy:

```
cp utils/pre-commit .git/hooks/pre-commit
```

Note: this will not auto-format the code — only run a check whether it's formatted. The same that CI does.


oleganza:
nvm

PR #213, opened 2019-03-11
zkvm: removes qty and flavor instructions

### LINKED ISSUE #153: zkvm: implement `qty` and `flavor` instructions
[`qty`](
https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#qty) and [`flavor`](
https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#flavor) permit placing constraints on existing value or [reblinding](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#reblind) their commitments.

A good task for a beginner.

debnil:
Removes the qty and flavor instructions. Closes #153.

### DIFF
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -983,23 +983,21 @@ Code | Instruction                | Stack diagram                              |
 0x13 | [`issue`](#issue)          |    _qty flv data pred_ → _contract_        | Modifies [CS](#constraint-system), [tx log](#transaction-log), [defers point ops](#deferred-point-operations)
 0x14 | [`borrow`](#borrow)        |         _qty flv_ → _–V +V_                | Modifies [CS](#constraint-system)
 0x15 | [`retire`](#retire)        |           _value_ → ø                      | Modifies [CS](#constraint-system), [tx log](#transaction-log)
-0x16 | [`qty`](#qty)              |           _value_ → _value qtyvar_         |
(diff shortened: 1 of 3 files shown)


zkvm/docs/zkvm-spec.md
```diff
@@ -1985,13 +1967,8 @@ UTXO IDs are not generally known without fuller view on transaction flow. This c
 
 That said, for contracts created from another contract, the contract ID is determined locally by the parent contract’s ID.
 
+### Do we need qty/flavor introspection ops?
+Previously, we thought we needed them to reblind the received value. However, we normally use `borrow` instead of placing some bounds on a received value. If we only ever mix all values and borrow necessary payments, we do not need to expose these variables at all.
```
oleganza: Correction: reblinding and constraints are orthogonal concepts:

1. "reblinding" was a (now scrapped) protocol to "encrypt a value to a pubkey" in one tx, and then decrypt it in another tx. We will use elgamal encryptions and codify encrypted values or data in specific types, so we won't be operating on individual variables or commitments. Therefore qty/flavor introspection won't be needed there.
2. "constraints on values" are still relevant as always, but with `borrow` we do not need to inspect a Value object — we can simply constrain raw commitments/variables and then `borrow` a value with such commitment. Hence, no need for qty/flavor for this use-case.

debnil: Thanks, this was super helpful!

PR #216, opened 2019-03-13
zkvm: serialize Tx object

### LINKED ISSUE #131: zkvm: serialization for `Tx` object
We need to serialize `Tx` object (header, program, signature, r1cs proof) to send it around the network.

R1CS proof serialization: https://github.com/dalek-cryptography/bulletproofs/pull/243

Use the R1CS as an example: `to_bytes`, `from_bytes` and `serde` support on top of those.

debnil:
Implements serialization of Tx object. Closes #131 .

### DIFF
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -1605,10 +1605,10 @@ A [Transaction](#transaction) is serialized as follows:
 
 ```
         SerializedTx = TxHeader || LE32(len(Program)) || Program || Signature || Proof
-        TxHeader = <24 bytes>
+        TxHeader = LE64(version) || LE64(mintime) || LE64(maxtime)
         Program = <len(Program) bytes>
         Signature = <64 bytes>
-        Proof = <14 * 32 + len(InnerProductProof) bytes>
+        Proof = <14·32 + len(InnerProductProof) bytes>
 ```
 
 ## Examples
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -57,7 +57,7 @@ ZkVM defines a procedural representation for blockchain transactions and the rul
     * [Constraint system instructions](#constraint-system-instructions)
     * [Value instructions](#value-instructions)
     * [Contract instructions](#contract-instructions)
-* [Transaction Encoding](#tx-encoding)
+* [Transaction Encoding](#transaction-encoding)
 * [Examples](#examples)
     * [Lock value example](#lock-value-example)
     * [Unlock value example](#unlock-value-example)
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -57,6 +57,7 @@ ZkVM defines a procedural representation for blockchain transactions and the rul
     * [Constraint system instructions](#constraint-system-instructions)
     * [Value instructions](#value-instructions)
     * [Contract instructions](#contract-instructions)
+* [Transaction Encoding](#tx-encoding)
 * [Examples](#examples)
     * [Lock value example](#lock-value-example)
     * [Unlock value example](#unlock-value-example)
@@ -1598,7 +1599,17 @@ All unassigned instruction codes are interpreted as no-ops. This are reserved fo
 
 See [Versioning](#versioning).
 
+## Transaction Encoding
 
+This section contains the rules for serializing a [Transaction](#transaction) object.
+
+```
+        SerializedTx = TxHeader || LE32(len(Program)) || Program || Signature || Proof
+        TxHeader = <24 bytes>
+        Program = <len(Program) bytes>
+        Signature = <64 bytes>
+        Proof = <14 * 32 + len(InnerProductProof) bytes>
+```
 
 ## Examples
 
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -1601,7 +1601,7 @@ See [Versioning](#versioning).
 
 ## Transaction Encoding
 
-This section contains the rules for serializing a [Transaction](#transaction) object.
+A [Transaction](#transaction) is serialized as follows:
 
 ```
         SerializedTx = TxHeader || LE32(len(Program)) || Program || Signature || Proof
--- a/zkvm/src/vm.rs
+++ b/zkvm/src/vm.rs
@@ -3,9 +3,10 @@ use bulletproofs::r1cs::R1CSProof;
 use curve25519_dalek::ristretto::CompressedRistretto;
 use curve25519_dalek::scalar::Scalar;
 use merlin::Transcript;
+use serde::de::Visitor;
+use serde::{self, Deserialize, Deserializer, Serialize, Serializer};
 use spacesuit;
 use spacesuit::BitRange;
-use serde::Serialize;
 use std::iter::FromIterator;
 use std::mem;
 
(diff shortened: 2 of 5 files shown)


bobg:
Please do not merge this PR without updating the spec with serialization rules (which could be part of this PR).

zkvm/src/vm.rs
```diff
@@ -51,6 +56,99 @@ pub struct Tx {
...
+    }
+
+    /// Deserializes the tx from a byte slice.
+    ///
+    /// Returns an error if the byte slice cannot be parsed into a `Tx`.
+    pub fn from_bytes(slice: &[u8]) -> Result<Tx, VMError> {
+        if slice.len() < (24 + 4 + 64 + 14 * 32) {
+            return Err(VMError::FormatError);
+        }
+        let mut reader = SliceReader::new(slice);
```
oleganza: Sorry, I should've documented the reason for keeping`::new` private. Instead, one should use `::parse` which ensures that the slice is fully consumed and no trailing garbage bytes are left unparsed. See the usage of `::parse` in other places in the code.

zkvm/src/vm.rs
```diff
@@ -51,6 +56,99 @@ pub struct Tx {
...
+        let version = reader.read_u64()?;
```
oleganza: TxHeader may implement `decode`/`encode` like Contract/Output and other types do, so we can simply do `let header = TxHeader::decode(reader)`.

oleganza: For that matter, `from_bytes` should probably be a thin wrapper method around `Tx::decode(reader)`.

zkvm/src/vm.rs
```diff
@@ -51,6 +56,99 @@ pub struct Tx {
...
+    }
+
+    /// Returns the size in bytes required to serialize the `Tx`.
+    pub fn serialized_size(&self) -> usize {
+        // header is 8 bytes * 3 fields = 24 bytes
+        // program length is 4 bytes
+        // program is self.program.len() bytes
+        // signature is 64 bytes
+        // proof is 14*32 + the ipp bytes
+        8 * 3 + 4 + self.program.len() + 64 + self.proof.serialized_size()
```
```suggestion
        self.header.serialized_size() + 4 + self.program.len() + 64 + self.proof.serialized_size()
```

maybe factor `8*3` into the TxHeader's method?

zkvm/src/vm.rs
```diff
@@ -51,6 +56,99 @@ pub struct Tx {
...
+    }
+
+    /// Deserializes the tx from a byte slice.
+    ///
+    /// Returns an error if the byte slice cannot be parsed into a `Tx`.
+    pub fn from_bytes(slice: &[u8]) -> Result<Tx, VMError> {
+        if slice.len() < (24 + 4 + 64 + 14 * 32) {
```
oleganza: this check is taken care by SliceReader::parse already

zkvm/src/vm.rs
```diff
@@ -51,6 +56,99 @@ pub struct Tx {
     pub proof: R1CSProof,
 }
 
+impl Tx {
+    /// Serializes the tx into a byte array.
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut buf = Vec::with_capacity(self.serialized_size());
+        encoding::write_u64(self.header.version, &mut buf);
```
oleganza: Same as with parsing `TxHeader`: we can factor this out into `TxHeader::encode()`.

zkvm/docs/zkvm-spec.md
```diff
@@ -1598,7 +1599,17 @@ All unassigned instruction codes are interpreted as no-ops. This are reserved fo
 
 See [Versioning](#versioning).
 
+## Transaction Encoding
 
+A [Transaction](#transaction) is serialized as follows:
+
+```
+        SerializedTx = TxHeader || LE32(len(Program)) || Program || Signature || Proof
+        TxHeader = <24 bytes>
```
```suggestion
        TxHeader = LE64(version) || LE64(mintime) || LE64(maxtime)
```

zkvm/docs/zkvm-spec.md
```diff
@@ -1598,7 +1599,17 @@ All unassigned instruction codes are interpreted as no-ops. This are reserved fo
...
+        Program = <len(Program) bytes>
+        Signature = <64 bytes>
+        Proof = <14 * 32 + len(InnerProductProof) bytes>
```
oleganza: I try to use `·` for multiplication everywhere in the docs for the sake of typography.
```suggestion
        Proof = <14·32 + len(InnerProductProof) bytes>
```

zkvm/src/vm.rs
```diff
@@ -36,6 +39,32 @@ pub struct TxHeader {
...
+    }
+
+    fn encode(&self, buf: &mut Vec<u8>) {
+        encoding::write_u64(self.version, buf);
+        encoding::write_u64(self.mintime, buf);
+        encoding::write_u64(self.maxtime, buf);
+    }
+
+    fn decode<'a>(reader: &mut SliceReader<'a>) -> Result<Self, VMError> {
+        let (header, _) = reader.slice(|r| {
```
oleganza: `.slice` method is only needed when you need the surrounding slice, e.g. to hash the serialized object w/o re-serializing it. Here we do not need that slice (hence use of `_` in this line), so the whole `.slice` call is not needed. We can directly do `reader.read_u64()` etc.

oleganza:
```rust
fn decode(reader: ...) {
    Ok(TxHeader {
        version: r.read_u64()?,
        mintime: r.read_u64()?,
        maxtime: r.read_u64()?,
    })
}
```

zkvm/src/vm.rs
```diff
@@ -51,6 +80,96 @@ pub struct Tx {
...
+    fn encode(&self, buf: &mut Vec<u8>) {
+        self.header.encode(buf);
+        encoding::write_size(self.program.len(), buf);
+        buf.extend(&self.program);
+        buf.extend_from_slice(&self.signature.to_bytes());
+        buf.extend_from_slice(&self.proof.to_bytes());
+    }
+
+    fn decode<'a>(reader: &mut SliceReader<'a>) -> Result<Tx, VMError> {
+        let (tx, _) = reader.slice(|r| {
```
oleganza: as in the above comment, no need for `.slice` wrapper, since we do not want to use that slice in any way.

zkvm/src/vm.rs
```diff
@@ -51,6 +80,96 @@ pub struct Tx {
...
+            let header = TxHeader::decode(r)?;
+            let prog_len = r.read_size()?;
+            let program = r.read_bytes(prog_len)?.to_vec();
+
+            let mut sig_buf = [0u8; 64];
+            sig_buf.copy_from_slice(r.read_bytes(64)?);
```
oleganza: there's already `read_u8x64` for reading `[u8; 64]`.

zkvm/src/vm.rs
```diff
@@ -51,6 +80,96 @@ pub struct Tx {
...
+        // signature is 64 bytes
+        // proof is 14*32 + the ipp bytes
+        self.header.serialized_size() + 4 + self.program.len() + 64 + self.proof.serialized_size()
+    }
+
+    /// Deserializes the tx from a byte slice.
+    ///
+    /// Returns an error if the byte slice cannot be parsed into a `Tx`.
+    pub fn from_bytes(slice: &[u8]) -> Result<Tx, VMError> {
+        let tx = SliceReader::parse(slice, |r| Self::decode(r))?;
```
oleganza: no need for extra variables - the result of SliceReader::parse is of the same type as we need.
```suggestion
pub fn from_bytes(slice: &[u8]) -> Result<...> {
    SliceReader::parse(slice, |r| Self::decode(r))
}
```

oleganza:
Seems correct 👍 , just left a few nitpicks.

zkvm/src/vm.rs
debnil: Got it, this was super helpful, thank you!

PR #217, opened 2019-03-13
zkvm/constraints: adds boolean `not` instruction

### LINKED ISSUE #117: zkvm: boolean `not` instruction
There's a cool trick to implement logical negation described in [Setty, Vu, Panpalia, Braun, Ali, Blumberg, Walfish (2012)](https://eprint.iacr.org/2012/598.pdf) (appendix D.1), later [implemented in libsnark](https://github.com/scipr-lab/libsnark/blob/dfa74ff270ca295619be1fdf7661f76dff0ae69e/libsnark/gadgetlib1/gadgets/basic_gadgets.hpp#L162-L169). It lets us use just two multipliers:

```
x - the input secret scalar mod |G|.
y - the output of NOT(x): if x == 0 then 1 else 0.
w - additional free variable chosen by the prover.
```

Statements in R1CS:

```
x·y = 0
x·w = 1-y
```

* The first statement implies that `y == 0` when `x != 0`.
* The second statement implies that `y == 1` when `x == 0`. 
* To satisfy both statements, the prover must choose `w` to be `x^-1 mod |G|` if `x != 0`.
* If `x == 0` the `w` could be set to an arbitrary value.

### Spec draft

Stack diagram for the `not` instruction:

_constr1_ **not** → _constr2_

1. Allocate two multipliers.
2. Constrain the first mult's left wire to the input constraint (`x`).
3. Constrain the second mult's left wire to first mult's left wire (i.e. `x == x` on both multipliers).
4. Assign the second mult's right wire (`w`) to the inverse of input constraint `x` or zero if it's zero. No constraints needed for `w`.
5. Constrain the first mult's output wire to `0`.
6. Constrain the second mult's output wire to `1-y`, where `y` is the first mult's right wire.
7. Wrap the first mult's right wire (`y`) in a Constraint `constr2` and push it to the stack.

image

(4 constraints)

### Implementation notes

1. Add `Constraint::Not(Box<Constraint>)` variant to existing `::Eq/Or/And`.
2. Note that truth is represented by equations `stuff == 0`, therefore, the truth of NOT is represented by `y == 0` (which means that the corresponding input to the NOT must satisfy `stuff ≠ 0`).

### Alternative

@cathieyun made a good point that if we simply need to check that something is not zero, we could use a _single multiplier_ using statement `x·w == 1`, where `w` must be chosen to be `x^-1 mod |G|`. 

This could be used with a hypothetical instruction `notverify` that inverses the composition of boolean functions (And/Or). Unfortunately, this does not permit nice compositions like `not(x) and y`, `not(x) or y`. Bringing `not` outside in these would bring back `not` to the other term.

2-multiplier `not` instruction seems to be of a minor overhead and is easily composable with the rest of boolean operations, so probably the tradeoff is worth it.

vickiniu:
Adds boolean `not` instruction. Modifies `Constraint.flatten` to also return an optional assignment, used to calculate the inputs for `not`.

Fixes #117

### DIFF
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -977,9 +977,9 @@ Code | Instruction                | Stack diagram                              |
 0x0e | [`range:n`](#range)        |            _expr_ → _expr_                 | Modifies [CS](#constraint-system)
 0x0f | [`and`](#and)              | _constr1 constr2_ → _constr3_              |
 0x10 | [`or`](#or)                | _constr1 constr2_ → _constr3_              |
-0x11 | [`not`](#not)              | _constr1_ → _constr2_                      | Modifies [CS](#constraint-system)
+0x11 | [`not`](#not)              |         _constr1_ → _constr2_              | Modifies [CS](#constraint-system)
 0x12 | [`verify`](#verify)        |      _constraint_ → ø                      | Modifies [CS](#constraint-system) 
-0x13 | [`unblind`](#unblind)      |        _V v_ → _V_                         | [Defers point ops](#deferred-point-operations)
+0x13 | [`unblind`](#unblind)      |             _V v_ → _V_                    | [Defers point ops](#deferred-point-operations)
  |                                |                                            |
  |     [**Values**](#value-instructions)              |                        |
 0x14 | [`issue`](#issue)          |    _qty flv data pred_ → _contract_        | Modifies [CS](#constraint-system), [tx log](#transaction-log), [defers point ops](#deferred-point-operations)
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -977,28 +977,29 @@ Code | Instruction                | Stack diagram                              |
 0x0e | [`range:n`](#range)        |            _expr_ → _expr_                 | Modifies [CS](#constraint-system)
 0x0f | [`and`](#and)              | _constr1 constr2_ → _constr3_              |
 0x10 | [`or`](#or)                | _constr1 constr2_ → _constr3_              |
-0x11 | [`verify`](#verify)        |      _constraint_ → ø                      | Modifies [CS](#constraint-system) 
-0x12 | [`unblind`](#unblind)      |        _V v_ → _V_                         | [Defers point ops](#deferred-point-operations)
+0x11 | ['not'](#not)              | _constr1_ → _constr2_                      | Modifies [CS](#constraint-system)
+0x12 | [`verify`](#verify)        |      _constraint_ → ø                      | Modifies [CS](#constraint-system) 
+0x13 | [`unblind`](#unblind)      |        _V v_ → _V_                         | [Defers point ops](#deferred-point-operations)
  |                                |                                            |
  |     [**Values**](#value-instructions)              |                        |
-0x13 | [`issue`](#issue)          |    _qty flv data pred_ → _contract_        | Modifies [CS](#constraint-system), [tx log](#transaction-log), [defers point ops](#deferred-point-operations)
-0x14 | [`borrow`](#borrow)        |         _qty flv_ → _–V +V_                | Modifies [CS](#constraint-system)
-0x15 | [`retire`](#retire)        |           _value_ → ø                      | Modifies [CS](#constraint-system), [tx log](#transaction-log)
-0x16 | [`cloak:m:n`](#cloak)      | _widevalues commitments_ → _values_        | Modifies [CS](#constraint-system)
-0x17 | [`import`](#import)        |   _proof qty flv_ → _value_                | Modifies [CS](#constraint-system), [tx log](#transaction-log), [defers point ops](#deferred-point-operations)
-0x18 | [`export`](#export)        |       _value ???_ → ø                      | Modifies [CS](#constraint-system), [tx log](#transaction-log)
+0x14 | [`issue`](#issue)          |    _qty flv data pred_ → _contract_        | Modifies [CS](#constraint-system), [tx log](#transaction-log), [defers point ops](#deferred-point-operations)
+0x15 | [`borrow`](#borrow)        |         _qty flv_ → _–V +V_                | Modifies [CS](#constraint-system)
+0x16 | [`retire`](#retire)        |           _value_ → ø                      | Modifies [CS](#constraint-system), [tx log](#transaction-log)
+0x17 | [`cloak:m:n`](#cloak)      | _widevalues commitments_ → _values_        | Modifies [CS](#constraint-system)
+0x18 | [`import`](#import)        |   _proof qty flv_ → _value_                | Modifies [CS](#constraint-system), [tx log](#transaction-log), [defers point ops](#deferred-point-operations)
+0x19 | [`export`](#export)        |       _value ???_ → ø                      | Modifies [CS](#constraint-system), [tx log](#transaction-log)
  |                                |                                            |
  |     [**Contracts**](#contract-instructions)        |                        |
-0x19 | [`input`](#input)          |      _prevoutput_ → _contract_             | Modifies [tx log](#transaction-log)
-0x1a | [`output:k`](#output)      |   _items... pred_ → ø                      | Modifies [tx log](#transaction-log)
-0x1b | [`contract:k`](#contract)  |   _items... pred_ → _contract_             | 
-0x1c | [`nonce`](#nonce)          |    _pred blockid_ → _contract_             | Modifies [tx log](#transaction-log)
-0x1d | [`log`](#log)              |            _data_ → ø                      | Modifies [tx log](#transaction-log)
-0x1e | [`signtx`](#signtx)        |        _contract_ → _results..._           | Modifies [deferred verification keys](#transaction-signature)
-0x1f | [`call`](#call)            | _contract bf prog_ → _results..._          | [Defers point operations](#deferred-point-operations)
-0x20 | [`left`](#left)            |    _contract A B_ → _contract’_            | [Defers point operations](#deferred-point-operations)
-0x21 | [`right`](#right)          |    _contract A B_ → _contract’_            | [Defers point operations](#deferred-point-operations)
-0x22 | [`delegate`](#delegate)    |_contract prog sig_ → _results..._          | [Defers point operations](#deferred-point-operations)
+0x1a | [`input`](#input)          |      _prevoutput_ → _contract_             | Modifies [tx log](#transaction-log)
+0x1b | [`output:k`](#output)      |   _items... pred_ → ø                      | Modifies [tx log](#transaction-log)
+0x1c | [`contract:k`](#contract)  |   _items... pred_ → _contract_             | 
+0x1d | [`nonce`](#nonce)          |    _pred blockid_ → _contract_             | Modifies [tx log](#transaction-log)
+0x1e | [`log`](#log)              |            _data_ → ø                      | Modifies [tx log](#transaction-log)
+0x1f | [`signtx`](#signtx)        |        _contract_ → _results..._           | Modifies [deferred verification keys](#transaction-signature)
+0x20 | [`call`](#call)            |_contract bf prog_ → _results..._           | [Defers point operations](#deferred-point-operations)
+0x21 | [`left`](#left)            |    _contract A B_ → _contract’_            | [Defers point operations](#deferred-point-operations)
+0x22 | [`right`](#right)          |    _contract A B_ → _contract’_            | [Defers point operations](#deferred-point-operations)
+0x23 | [`delegate`](#delegate)    |_contract prog sig_ → _results..._          | [Defers point operations](#deferred-point-operations)
   —  | [`ext`](#ext)              |                 ø → ø                      | Fails if [extension flag](#vm-state) is not set.
 
 
@@ -1201,6 +1202,23 @@ _constraint1 constraint2_ **or** → _constraint3_
 No changes to the [constraint system](#constraint-system) are made until [`verify`](#verify) is executed.
 
 Fails if `c1` and `c2` are not [constraints](#constraint-type).
+
+#### not
+
+_constr1_ **not** → _constr2_
+
+1. Pops [constraint](#constraint-type) `c1`.
+2. Create two constraints:
+   ```
+   x * y = 0
+   x * w = 1-y
+   ```
+   where `w` is a random challenge and `x` is the expression from constraint `c1`.
+3. Wrap the output `y` in a constraint `c2`.
+4. Push `c2` to the stack.
+
+This implements the boolean `not` trick from [Setty, Vu, Panpalia, Braun, Ali, Blumberg, Walfish (2012)](https://eprint.iacr.org/2012/598.pdf) and implemented in [libsnark](https://github.com/scipr-lab/libsnark/blob/dfa74ff270ca295619be1fdf7661f76dff0ae69e/libsnark/gadgetlib1/gadgets/basic_gadgets.hpp#L162-L169).
+
 
 #### verify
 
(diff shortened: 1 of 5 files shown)


zkvm/src/constraints.rs
```diff
@@ -89,21 +95,60 @@ impl Constraint {
...
+                        let w = Scalar::conditional_select(&x_assg, &Scalar::one(), is_zero.into());
+                        let w = w.invert();
+
+                        let (x, y, o) = cs.multiply(x, y_assg.into());
+                        cs.constrain(o.into());
+                        let (_x, _w, xw) = cs.multiply(x.into(), w.into());
+                        cs.constrain(xw - Scalar::one() + y);
+                        Ok((r1cs::LinearCombination::from(y), Some(y_assg)))
+                    }
+                    None => Err(VMError::WitnessMissing),
```
oleganza: if the x assignment is missing, then we are inside the verifier, so we simply propagate None to all other assignments. Therefore, this whole method can remain being non-failable. 

As of `y_assg`, we should compute it only if x is Some, but keep it None if x is unassigned.

zkvm/src/constraints.rs
```diff
@@ -89,21 +95,60 @@ impl Constraint {
...
+                        Ok((r1cs::LinearCombination::from(o), Some(a_assg * b_assg)))
+                    }
+                    (_, _) => Ok((r1cs::LinearCombination::from(o), None)),
+                }
+            }
+            Constraint::Not(c1) => {
+                let (x, x_assg) = c1.flatten(cs)?;
+                match x_assg {
+                    Some(x_assg) => {
+                        let is_zero = (x_assg == Scalar::zero()) as u8;
```
oleganza: this should probably be `ct_eq` (consttime equality check returning Choice).

zkvm/src/constraints.rs
```diff
@@ -89,21 +95,60 @@ impl Constraint {
...
+                    (Some(a_assg), Some(b_assg)) => {
+                        Ok((r1cs::LinearCombination::from(o), Some(a_assg * b_assg)))
+                    }
+                    (_, _) => Ok((r1cs::LinearCombination::from(o), None)),
+                }
+            }
+            Constraint::Not(c1) => {
+                let (x, x_assg) = c1.flatten(cs)?;
+                match x_assg {
+                    Some(x_assg) => {
```
oleganza: How about this branch of `match x_assg` returns `(Some(y),Some(w))` and the `None` branch returns `(None,None)`?

oleganza: And then constraints and multipliers are allocated after the `match` with these optional assignments?

zkvm/src/constraints.rs
```diff
@@ -246,6 +291,16 @@ impl Expression {
             Expression::LinearCombination(terms, _) => r1cs::LinearCombination::from_iter(terms),
         }
     }
+
+    fn to_scalar(&self) -> Option<Scalar> {
```
oleganza: IMHO, we should preserve integerness of the assignment and make it clear that it is the _evaluation_ of the expression, not some sort of alternative representation (`to_*` is usually for conversion methods).
```suggestion
    fn eval(&self) -> Option<ScalarWitness> {
```

zkvm/src/constraints.rs
```diff
@@ -89,21 +95,60 @@ impl Constraint {
...
+                        let is_zero = (x_assg == Scalar::zero()) as u8;
+                        let y_assg = Scalar::conditional_select(
+                            &Scalar::zero(),
+                            &Scalar::one(),
+                            is_zero.into(),
+                        );
+                        let w = Scalar::conditional_select(&x_assg, &Scalar::one(), is_zero.into());
+                        let w = w.invert();
+
+                        let (x, y, o) = cs.multiply(x, y_assg.into());
```
oleganza: we should not use `multiply` API. We wouldn't be able to convert `Option<Scalar>` to a linear combination. Let's use `allocate_multiplier` for both multipliers with explicit assignments and explicit constraints for all inputs and outputs.

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,61 @@ impl Constraint {
...
+                            is_zero.into(),
+                        );
+                        let w = Scalar::conditional_select(&x_assg, &Scalar::one(), is_zero.into());
+                        let w = w.invert();
+                        (Some((x_assg, y_assg)), Some((x_assg, w)), Some(y_assg))
+                    }
+                    None => (None, None, None),
+                };
+                let (x, y, o) = cs.allocate_multiplier(xy)?;
+                cs.constrain(o.into());
```
oleganza: maybe clarify the intent?
```suggestion
                cs.constrain(o.into()); // enforce `x == 0`
```

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,61 @@ impl Constraint {
...
+                cs.constrain(x - x_lc);
+                let (_x, _w, xw) = cs.allocate_multiplier(xw)?;
```
oleganza: these are simply unconstrained variables, maybe rename those to `(l1,r1,o1)` from the first multiplier and to `(l2,r2,o2)` from the second one? Then you'd need `cs.constrain(l1 - l2)` and `cs.constrain(x_lc - l1)`.

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,61 @@ impl Constraint {
...
+                cs.constrain(xw - Scalar::one() + y);
```
oleganza:
```suggestion
                cs.constrain(xw - Scalar::one() + y); // enforce `x*w == 1 - y` (y must be 1 if x == 0)
```

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,61 @@ impl Constraint {
         .map_err(|e| VMError::R1CSError(e))
     }
 
-    fn flatten<CS: r1cs::RandomizedConstraintSystem>(self, cs: &mut CS) -> r1cs::LinearCombination {
+    fn flatten<CS: r1cs::RandomizedConstraintSystem>(
+        self,
+        cs: &mut CS,
+    ) -> Result<(r1cs::LinearCombination, Option<Scalar>), r1cs::R1CSError> {
```
oleganza: this is not failable anymore?

vickiniu: we need to be able to handle the errors from `allocate_multiplier`

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,64 @@ impl Constraint {
...
+            }
             Constraint::And(c1, c2) => {
-                let a = c1.flatten(cs);
-                let b = c2.flatten(cs);
+                let (a, a_assg) = c1.flatten(cs)?;
+                let (b, b_assg) = c2.flatten(cs)?;
                 let z = cs.challenge_scalar(b"ZkVM.verify.and-challenge");
-                a + z * b
+                match (a_assg, b_assg) {
+                    (Some(a_assg), Some(b_assg)) => Ok((a + z * b, Some(a_assg + z * b_assg))),
```
oleganza: can we factor LC logic out of assignment logic? So assignment mapping is completely separate. 

e.g.

```rust
let assignment = match (a_assg, b_assg) {
    (Some(a), Some(b)) => Some(a + z * b),
    _ => None
};
(a + z*b, assignment)
```

oleganza: Alternatively:

```rust
let assignment = a_assg.and_then(|a| b_assg.map(|b| a + z * b ));
(a + z*b, assignment)
```

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,64 @@ impl Constraint {
...
+                let (x_lc, x_assg) = c1.flatten(cs)?;
+                let (xy, xw, y_assg) = match x_assg {
+                    Some(x_assg) => {
+                        let is_zero = x_assg.ct_eq(&Scalar::zero());
+                        let y_assg = Scalar::conditional_select(
+                            &Scalar::zero(),
+                            &Scalar::one(),
+                            is_zero.into(),
+                        );
+                        let w = Scalar::conditional_select(&x_assg, &Scalar::one(), is_zero.into());
```
oleganza:
```suggestion
                        let w = Scalar::conditional_select(&x_assg, &Scalar::one(), is_zero);
```

`is_zero` is already a `Choice`.

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,64 @@ impl Constraint {
...
+                    (_, _) => Ok((r1cs::LinearCombination::from(o), None)),
+                }
+            }
+            Constraint::Not(c1) => {
+                let (x_lc, x_assg) = c1.flatten(cs)?;
+                let (xy, xw, y_assg) = match x_assg {
+                    Some(x_assg) => {
+                        let is_zero = x_assg.ct_eq(&Scalar::zero());
+                        let y_assg = Scalar::conditional_select(
+                            &Scalar::zero(),
```
oleganza: shouldn't `y` be 1 if x==0?

zkvm/src/constraints.rs
oleganza:
```suggestion
                        let w = Scalar::conditional_select(&Scalar::one(), &x_assg, is_zero);
```

Does `choice==true` select the first or the second variant?

oleganza: if x==0, we need w=1 to make it invertible

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,64 @@ impl Constraint {
...
+                match (a_assg, b_assg) {
+                    (Some(a_assg), Some(b_assg)) => {
+                        Ok((r1cs::LinearCombination::from(o), Some(a_assg * b_assg)))
+                    }
+                    (_, _) => Ok((r1cs::LinearCombination::from(o), None)),
+                }
+            }
+            Constraint::Not(c1) => {
+                let (x_lc, x_assg) = c1.flatten(cs)?;
+                let (xy, xw, y_assg) = match x_assg {
```
oleganza: those are all assignments, so using `_assg` consistently would help the reader.

zkvm/src/constraints.rs
oleganza: ouch. `r1cs::Prover` needs those assignments.

image

zkvm/src/constraints.rs
vickiniu: til about `and_then`!

zkvm/src/constraints.rs
vickiniu: hm, I see inconsistent documentation — the function doc says that it chooses `a` if `Choice(0)` but then the subtle docs online say the opposite. will investigate

vickiniu: based on https://doc.dalek.rs/src/subtle/lib.rs.html#364 I think that `conditional_select` chooses `a` on `Choice(0)`, so `a` on `false` and `b` on `true`

zkvm/src/constraints.rs
vickiniu: (see above, the assignments are a little backwards!)

zkvm/src/constraints.rs
oleganza: omg

zkvm/src/constraints.rs
oleganza: 👍

zkvm/docs/zkvm-spec.md
```diff
@@ -1201,6 +1202,23 @@ No changes to the [constraint system](#constraint-system) are made until [`verif
...
+
+_constr1_ **not** → _constr2_
+
+1. Pops [constraint](#constraint-type) `c1`.
+2. Create two constraints:
+   ```
+   x * y = 0
+   x * w = 1-y
+   ```
+   where `w` is a random challenge and `x` is the expression from constraint `c1`.
```
oleganza: `w` is not a random challenge, it's a free variable. 

Let's rephrase`x` to "`x` is the evaluation of the constraint `c1`". 

I'll add a section on evaluating expressions and constraints in the spec as a follow-up.

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,58 @@ impl Constraint {
...
+    fn flatten<CS: r1cs::RandomizedConstraintSystem>(
+        self,
+        cs: &mut CS,
+    ) -> Result<(r1cs::LinearCombination, Option<Scalar>), r1cs::R1CSError> {
         match self {
-            Constraint::Eq(expr1, expr2) => expr1.to_r1cs_lc() - expr2.to_r1cs_lc(),
+            Constraint::Eq(expr1, expr2) => {
+                let assignment = expr1
+                    .eval()
+                    .and_then(|x| expr2.eval().map(|y| x.to_scalar() - y.to_scalar()));
```
oleganza:
```suggestion
                    .and_then(|x| expr2.eval().map(|y| (x - y).to_scalar()));
```

this probably would be a bit easier to read.

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,58 @@ impl Constraint {
...
+                    }
+                    None => (None, None, None),
+                };
+                let (l1, r1, o1) = cs.allocate_multiplier(xy_assg)?;
+                // constraint l1 to x
+                cs.constrain(l1 - x_lc);
+                // enforce `x*y == 0` (y is 0 if x != 0)
+                cs.constrain(o1.into());
+                let (_l2, _r2, o2) = cs.allocate_multiplier(xw_assg)?;
+                // enforce `x*w == 1 - y` (y is 1 if x == 0)
```
oleganza: need to constrain `l2 == l1` (`x` is connected to left wires on both multipliers):

```
cs.constrain(l1 - l2);
```

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,58 @@ impl Constraint {
...
+                        (Some((x_assg, y_assg)), Some((x_assg, w_assg)), Some(y_assg))
+                    }
+                    None => (None, None, None),
+                };
+                let (l1, r1, o1) = cs.allocate_multiplier(xy_assg)?;
+                // constraint l1 to x
+                cs.constrain(l1 - x_lc);
+                // enforce `x*y == 0` (y is 0 if x != 0)
+                cs.constrain(o1.into());
+                let (_l2, _r2, o2) = cs.allocate_multiplier(xw_assg)?;
```
oleganza: maybe for clarity, allocate both multipliers first, then add 4 constraints and additionally comment that r2 is left unconstrained intentionally.

zkvm/src/constraints.rs
```diff
@@ -89,21 +92,63 @@ impl Constraint {
...
+                    None => (None, None, None),
+                };
+                // Allocate multipliers
+                let (l1, r1, o1) = cs.allocate_multiplier(xy_assg)?;
+                let (_l2, _r2, o2) = cs.allocate_multiplier(xw_assg)?;
+
+                // Add constraints: r2 left unconstrained since w is
+                // a free varaible
+
+                // `l1 == x`
```
oleganza: still missing l1=l2 constraint: 
https://github.com/interstellar/slingshot/pull/217#discussion_r265351704

oleganza: total - 4 constraints

zkvm/src/constraints.rs
```diff
@@ -119,35 +119,45 @@ impl Constraint {
...
                             &Scalar::one(),
                             is_zero.into(),
                         );
-                        let w_assg =
-                            Scalar::conditional_select(&x_assg, &Scalar::one(), is_zero.into());
-                        let w_assg = w_assg.invert();
-                        (Some((x_assg, y_assg)), Some((x_assg, w_assg)), Some(y_assg))
+                        let w = Scalar::conditional_select(&x, &Scalar::one(), is_zero.into());
+                        let w = w.invert();
+                        (Some((x, w)), Some((x, w)), Some(y))
```
oleganza: i introduced a bug here (`w` instead of `y` in the first tuple), oops (fixed in a later commit).

PR #220, opened 2019-03-14
keytree: store precomputed transcripts on Xprv, Xpub

tessr:
This is Step 4 in https://github.com/interstellar/slingshot/pull/202#discussion_r263216987, 
working to make sure that intermediate key derivation isn't failable. 
This PR also updates the spec to match this change.

### DIFF
--- a/keytree/keytree.md
+++ b/keytree/keytree.md
@@ -42,18 +42,20 @@ Stands for _extended private key_: consists of a secret [scalar](#scalar) and a 
 ```rust
 struct Xprv {
   scalar: Scalar,
-  dk: [u8; 32]
+  dk: [u8; 32],
+  transcript: Transcript,
 }
 ```
 
 ### Xpub
 
-Stands for _extended public key_: consists of a [point](#point) and a [derivation key](#derivation-key).
+Stands for _extended public key_: consists of a [point](#point), a [derivation key](#derivation-key), and a transcript.
 
 ```rust
 struct Xpub {
   point: CompressedRistretto,
-  dk: [u8; 32]
+  dk: [u8; 32],
+  Transcript: transcript,
 }
 ```
 
@@ -74,6 +76,7 @@ If you need to share an individual public key, use [leaf key derivation](#derive
 	```
 	xprv = Xprv { scalar, dk }
 	```
+5. Create Merlin `t = Transcript::new("Keytree.derivation")`.
 5. Return the resulting extended private key `xprv`.
 
 ### Convert Xprv to Xpub
@@ -89,22 +92,21 @@ Xpub {
 
 ### Derive an intermediate key
 
-1. Create Merlin `t = Transcript::new("Keytree.intermediate")`.
-2. Commit [xpub](#xpub) to transcript (if xprv is provided, compute xpub from xprv):
+1. Commit [xpub](#xpub) to the provided key's transcript (if xprv is provided, compute xpub from xprv):
 	```
 	t.commit_bytes("pt", xpub.point)
 	t.commit_bytes("dk", xpub.dk)
 	```
-3. Provide the transcript to the user to commit an arbitrary derivation path or index:
+2. Provide the transcript to the user to commit an arbitrary derivation path or index:
 	```
-	t.commit_bytes(label, data) 
+	t.commit_bytes(label, data)
 	```
 	E.g. `t.commit_u64("account", account_id)` for an account within a hierarchy of keys.
-4. Squeeze a blinding factor `f`:
+3. Squeeze a blinding factor `f`:
 	```
-	f = t.challenge_scalar("f")
+	f = t.challenge_scalar("f.intermediate")
 	```
-5. Squeeze a new derivation key `dk2` (32 bytes):
+4. Squeeze a new derivation key `dk2` (32 bytes):
 	```
 	dk2 = t.challenge_bytes("dk")
 	```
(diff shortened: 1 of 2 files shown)


keytree/src/lib.rs
```diff
@@ -25,7 +31,19 @@ impl Xprv {
         let scalar = Scalar::random(&mut rng);
         let mut dk = [0u8; 32];
         rng.fill_bytes(&mut dk);
-        Xprv { scalar, dk }
+
+        let mut transcript = Transcript::new(b"Keytree.intermediate");
```
oleganza: Since it is going to be used for both intermediate nodes and leafs, i’d change label to “.derivation”, and adjust the spec accordingly (see also my comments on the domain separation in another PR).

tessr: Right, thanks.

tessr:
also, another question for you about this design, @oleganza: why does the Xprv need to know about the transcript? it seems like only the Xpub uses it, @cathieyun points out.

cathieyun:
^ the reason I was wondering, is because it seems much more intuitive for the `to_xpub()` function to make the transcript itself, if the `Xprv` does not care about the transcript. Did you suggest the current design because of a performance improvement (faster to clone a pre-made transcript, rather than to instantiate it again?), or because the `Xprv` actually does need the transcript, or something else?

keytree/keytree.md
```diff
@@ -42,18 +42,20 @@ Stands for _extended private key_: consists of a secret [scalar](#scalar) and a
...
-  dk: [u8; 32]
+  dk: [u8; 32],
+  transcript: Transcript,
 }
 ```
 
 ### Xpub
 
-Stands for _extended public key_: consists of a [point](#point) and a [derivation key](#derivation-key).
+Stands for _extended public key_: consists of a [point](#point), a [derivation key](#derivation-key), and a transcript.
```
oleganza: i'd remove the `transcript` field from the spec - that's the implementation detail. From the user's perspective, there's only a key and dk, and derivation rules from them. We simply precompute and reuse transcript for efficiency.

tessr: Hmm. I see where you're going with that, but doesn't this spec include implementation details? E.g. steps 1-6 for "derive an intermediate key" seem like an implementation detail.

oleganza:
Xprv should support the same derivation logic "in parallel" with Xpub, because the signer is going to be using it to derive signing keys. While the "accounting software" would derive only Xpub->Xpub.

tessr:
> because the signer is going to be using it to derive signing keys.

in other words, the signer needs access to the transcript?

oleganza:
>Did you suggest the current design because of a performance improvement (faster to clone a pre-made transcript, rather than to instantiate it again?)

Yeah, if you don't have this transcript in the parent, for each child key, you'd have to:
1. create a new transcript object
2. compress the decompressed pubkey and write it in (`pk` - compression is expensive)
3. mix in `dk` (although, that's cheap)

If that's precomputed in the parent, then derivation of each child is faster (2x for xpub->xpub: only one scalar-point mul, and 1000x faster for xprv->xprv because no expensive ops are made).

oleganza:
>in other words, the signer needs access to the transcript?

No, that's still an internal detail. Signer needs only a derived xprv, which is derived using the same transcript as with xpub->xpub.

tessr:
Oh, maybe what I'm missing here is that you can derive Xprvs from a parent Xprv? Otherwise, I'm still only seeing the transcript used for deriving Xpub -> Xpub.

oleganza:
If we had stored compressed pubkey in the parent (and xprv too), we'd have a slightly uglier failable API, and would not be precomputing transcript - making it up and hashing things in is pretty cheap. 

But since we store uncompressed point, we can go an extra mile and cache the compressed point as part of the transcript state, so we can even avoid redoing all the hashing. Even nicer!

oleganza:
>Oh, maybe what I'm missing here is that you can derive Xprvs from a parent Xprv? 

Yes! Sorry the spec draft doesn't make this clear, but it sort-of mentions it: you can derive both xpub->xpub and xprv->xprv.

oleganza:
Definitely, need to clarify it.

oleganza:
image

tessr:
Aha! I think that was what both @cathieyun and I were missing. I assumed that the flow was like:

```
Xprv -> Xpub -> intermediate Xpub 0
          or -> intermediate Xpub 1
          or -> intermediate Xpub n
```

but really, it's:

```
Xprv -> Xpub -> intermediate Xpub 0
          or -> intermediate Xpub n
  or -> intermediate Xprv 0 
  or -> intermediate Xprv 1
```

oleganza:
it's more like this:
image

tessr:
Oh good, this is very helpful.

keytree/keytree.md
oleganza: _trying to articulate it better..._

The spec is focused on the _protocol_, while where we put the transcript is an implementation detail. We can instead store compressed pubkey (along the decompressed one) and recompute transcript anew on each derivation, that'd be almost as efficient. Or we could compress it on the fly and make things slower. Or we could store the compressed only, and avoid redundancy, but then derivation API would be failable which isn't nice. The protocol does not change regardless of those choices, and i think it's better to keep those choices out of the spec to allow other implementors focus on what actually matters for users, instead of micromanaging the implementation.

PR #223, opened 2019-03-14
zkvm: encapsulate bulletproofs dependencies

### LINKED ISSUE #210: zkvm: remove bulletproofs dependency from API
Would be good if BulletproofsGens and other entities from BP were encapsulated inside ZkVM or at least re-exported, so that the users can add ZkVM as a dependency w/o also bringing in bulletproofs. See `token` crate as an example.

1. `BulletproofsGens`: those can be instantiated by ZkVM with a capacity parameter. R1CS can take them at the `Prover::prove` method, so ZkVM can just create a necessary capacity based on number of multiplicative constraints used. This can also be a part of the `Tx` object.
2. `PedersenGens` are already baked into Commitment and Predicate API, and we can do the same for Signature/VerificationKey API, so users do not create those things with raw scalar operations. If they need to do something custom and unsupported by ZkVM APIs, then they can import Ristretto or Bulletproofs explicitly.
3. R1CS instance is already encapsulated inside ZkVM prover/verifier.

vickiniu:
Fixes #210 - uses capacity of 256 for bulletprooofs generator.

### DIFF
--- a/zkvm/src/verifier.rs
+++ b/zkvm/src/verifier.rs
@@ -79,10 +79,11 @@ impl<'a, 'b> Delegate<r1cs::Verifier<'a, 'b>> for Verifier<'a, 'b> {
 impl<'a, 'b> Verifier<'a, 'b> {
     /// Verifies the `Tx` object by executing the VM and returns the `VerifiedTx`.
     /// Returns an error if the program is malformed or any of the proofs are not valid.
-    pub fn verify_tx<'g>(tx: Tx, bp_gens: &'g BulletproofGens) -> Result<VerifiedTx, VMError> {
+    pub fn verify_tx<'g>(tx: Tx) -> Result<VerifiedTx, VMError> {
+        let bp_gens = BulletproofGens::new(256, 1);
         let mut r1cs_transcript = Transcript::new(b"ZkVM.r1cs");
         let pc_gens = PedersenGens::default();
-        let cs = r1cs::Verifier::new(bp_gens, &pc_gens, &mut r1cs_transcript);
+        let cs = r1cs::Verifier::new(&bp_gens, &pc_gens, &mut r1cs_transcript);
 
         let mut verifier = Verifier {
             signtx_keys: Vec::new(),
--- a/zkvm/src/verifier.rs
+++ b/zkvm/src/verifier.rs
@@ -79,7 +79,7 @@ impl<'a, 'b> Delegate<r1cs::Verifier<'a, 'b>> for Verifier<'a, 'b> {
 impl<'a, 'b> Verifier<'a, 'b> {
     /// Verifies the `Tx` object by executing the VM and returns the `VerifiedTx`.
     /// Returns an error if the program is malformed or any of the proofs are not valid.
-    pub fn verify_tx<'g>(tx: Tx) -> Result<VerifiedTx, VMError> {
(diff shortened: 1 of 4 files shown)


zkvm/src/verifier.rs
```diff
@@ -79,10 +79,11 @@ impl<'a, 'b> Delegate<r1cs::Verifier<'a, 'b>> for Verifier<'a, 'b> {
 impl<'a, 'b> Verifier<'a, 'b> {
     /// Verifies the `Tx` object by executing the VM and returns the `VerifiedTx`.
     /// Returns an error if the program is malformed or any of the proofs are not valid.
-    pub fn verify_tx<'g>(tx: Tx, bp_gens: &'g BulletproofGens) -> Result<VerifiedTx, VMError> {
+    pub fn verify_tx(tx: Tx) -> Result<VerifiedTx, VMError> {
+        let bp_gens = BulletproofGens::new(128, 1);
```
oleganza: We cannot hardcode this for two reasons:

1. Bigger txs will have larger circuits and more than 128 multipliers, failing the verification if 128 is hardcoded.
2. Computing generators is expensive. Validator would want to precompute them and use over its whole lifetime to validate arbitrary amount of transactions, so we need to pass it in.

zkvm/src/verifier.rs
```diff
@@ -79,10 +79,11 @@ impl<'a, 'b> Delegate<r1cs::Verifier<'a, 'b>> for Verifier<'a, 'b> {
 impl<'a, 'b> Verifier<'a, 'b> {
     /// Verifies the `Tx` object by executing the VM and returns the `VerifiedTx`.
     /// Returns an error if the program is malformed or any of the proofs are not valid.
-    pub fn verify_tx<'g>(tx: Tx, bp_gens: &'g BulletproofGens) -> Result<VerifiedTx, VMError> {
+    pub fn verify_tx(tx: Tx) -> Result<VerifiedTx, VMError> {
+        let bp_gens = BulletproofGens::new(256, 1);
```
oleganza: We cannot hardcode this for two reasons:

1. Bigger txs will have larger circuits and more than 128 multipliers, failing the verification if 128 is hardcoded.
2. Computing generators is expensive. Validator would want to precompute them and use over its whole lifetime to validate arbitrary amount of transactions, so we need to pass it in.

vickiniu: Got it, so on the prover we'd want to calculate the capacity that we need?

If the verifier might want to precompute generators, would we need to keep `bp_gens` as a parameter on `verify_tx`?

oleganza: Yeah, that's why this discussion was prompted: https://dalek-cryptography.slack.com/archives/CBKMRC8DN/p1552425846018800

oleganza: As of exposing BP API, we could re-export only `BulletproofGens` in the lib.rs to allow user precompute them and pass into verifier. This way if they don't care about other parts of R1CS, they can avoid explicitly linking with `bulletproofs` crate.

vickiniu: got it, that makes sense! reading slack thread now :)

oleganza:
We should probably simply do `pub use bulletproofs;` in the lib.rs to allow users pick various bulletproofs types (such as generators) directly via `zkvm::bulletproofs::...` to avoid adding BP as a dependency.

PR #224, opened 2019-03-14
zkvm: commit blinding factor with program in predicate tree

### LINKED ISSUE #214: zkvm: add salt to program commitment in a predicate tree
Currently the program is committed to a [Predicate](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#predicate) with an unsalted hash:

```
P = hash(program)·B2
```

The theory is that the program may include `push:<data> drop` instruction that provides entropy without having an overhead of verifying a blinding factor in a Pedersen commitment:

```
P = hash(program)·B2 + blinding·B
```

The efficient alternative would be to instead salt the hash:

```
P = hash(salt, program)·B2
```

This way the program can remain clean, the salt can be generated by the Predicate API automatically, and the `call` instruction must be changed to accept one extra data argument `salt` popped from the stack.

Also seems like the salt can be safely made 16-byte long for a typical 128-bit security level. We can not mandate any specific length for the salt and leave this to applications, since it's just a matter of privacy and not of network integrity.

debnil:
Closes #214 .

### DIFF
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -995,7 +995,7 @@ Code | Instruction                | Stack diagram                              |
 0x1c | [`nonce`](#nonce)          |    _pred blockid_ → _contract_             | Modifies [tx log](#transaction-log)
 0x1d | [`log`](#log)              |            _data_ → ø                      | Modifies [tx log](#transaction-log)
 0x1e | [`signtx`](#signtx)        |        _contract_ → _results..._           | Modifies [deferred verification keys](#transaction-signature)
-0x1f | [`call`](#call)            |   _contract prog_ → _results..._           | [Defers point operations](#deferred-point-operations)
+0x1f | [`call`](#call)            | _contract salt prog_ → _results..._        | [Defers point operations](#deferred-point-operations)
 0x20 | [`left`](#left)            |    _contract A B_ → _contract’_            | [Defers point operations](#deferred-point-operations)
 0x21 | [`right`](#right)          |    _contract A B_ → _contract’_            | [Defers point operations](#deferred-point-operations)
 0x22 | [`delegate`](#delegate)    |_contract prog sig_ → _results..._          | [Defers point operations](#deferred-point-operations)
@@ -1506,10 +1506,10 @@ is deferred until the end of VM execution.
 
 #### call
 
-_contract(P) prog_ **call** → _results..._
+_contract(P) salt prog_ **call** → _results..._
 
-1. Pops the [data](#data-type) `prog` and a [contract](#contract-type) `contract`.
-2. Reads the [predicate](#predicate) `P` from the contract.
+1. Pops the [data](#data-type) `prog`, its associated [data](#data-type) `salt`, and a [contract](#contract-type) `contract`.
+2. Reads the [predicate](#predicate) `P` from the contract and its salt.
 3. Forms a statement for [program predicate](#program-predicate) of `prog` being equal to `P`:
     ```
     0 == -P + h(prog)·B2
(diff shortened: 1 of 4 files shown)


zkvm/src/predicate.rs
```diff
@@ -95,6 +95,18 @@ impl Predicate {
...
+    /// Creates a program with a random salt.
+    pub fn blinded_program<T: Into<Program>>(x: T) -> Self {
+        let salt: [u8; 16] = rand::random();
+        Predicate::Program(x.into(), salt.to_vec())
+    }
+
+    /// Creates a program with an empty salt.
+    pub fn unblinded_program<T: Into<Program>>(x: T) -> Self {
+        let salt: [u8; 0] = [];
+        Predicate::Program(x.into(), salt.to_vec())
```
oleganza:
```suggestion
        Predicate::Program(x.into(), Vec::new())
```

zkvm/src/predicate.rs
```diff
@@ -95,6 +95,18 @@ impl Predicate {
         }
     }
 
+    /// Creates a program with a random salt.
+    pub fn blinded_program<T: Into<Program>>(x: T) -> Self {
+        let salt: [u8; 16] = rand::random();
```
oleganza: Actually, `salt` is not a good name. It's private, so it should be called "pepper". But since we have exhausted our meme budget with a motocrab, maybe call it "blinding" (short for "blinding factor")?

zkvm/src/predicate.rs
```diff
@@ -173,17 +186,19 @@ mod tests {
     #[test]
     fn valid_program_commitment() {
         let prog = Program::build(|p| p.drop());
-        let pred = Predicate::Program(prog.clone());
-        let op = pred.prove_program_predicate(&bytecode(&prog));
+        let pred = Predicate::unblinded_program(prog.clone());
+        let salt: [u8; 0] = [];
+        let op = pred.prove_program_predicate(&bytecode(&prog), &salt);
```
oleganza: maybe we can avoid using any extra variable?
```suggestion
        let op = pred.prove_program_predicate(&bytecode(&prog), &[]);
```

zkvm/src/predicate.rs
```diff
@@ -173,17 +186,19 @@ mod tests {
...
 
     #[test]
     fn invalid_program_commitment() {
         let prog = Program::build(|p| p.drop());
         let prog2 = Program::build(|p| p.dup(1));
-        let pred = Predicate::Program(prog);
-        let op = pred.prove_program_predicate(&bytecode(&prog2));
+        let pred = Predicate::unblinded_program(prog);
+        let salt: [u8; 0] = [];
+        let op = pred.prove_program_predicate(&bytecode(&prog2), &salt);
```
oleganza:
```suggestion
        let op = pred.prove_program_predicate(&bytecode(&prog2), &[]);
```

oleganza:
this looks perfect, only a bikeshed about "salt" left

zkvm/docs/zkvm-spec.md
```diff
@@ -995,7 +995,7 @@ Code | Instruction                | Stack diagram                              |
 0x1c | [`nonce`](#nonce)          |    _pred blockid_ → _contract_             | Modifies [tx log](#transaction-log)
 0x1d | [`log`](#log)              |            _data_ → ø                      | Modifies [tx log](#transaction-log)
 0x1e | [`signtx`](#signtx)        |        _contract_ → _results..._           | Modifies [deferred verification keys](#transaction-signature)
-0x1f | [`call`](#call)            |   _contract prog_ → _results..._           | [Defers point operations](#deferred-point-operations)
+0x1f | [`call`](#call)            | _contract salt prog_ → _results..._        | [Defers point operations](#deferred-point-operations)
```
```suggestion
0x1f | [`call`](#call)            | _contract blinding prog_ → _results..._    | [Defers point operations](#deferred-point-operations)
```

zkvm/src/ops.rs
```diff
@@ -438,7 +438,8 @@ impl<'a> PredicateTree<'a> {
     /// Pushes program to the stack and calls the contract protected
     /// by the program predicate.
     pub fn call(self) -> Result<(), VMError> {
-        let subprog = self.pred.to_program()?;
+        let (subprog, salt) = self.pred.to_program()?;
```
```suggestion
        let (subprog, blinding) = self.pred.to_program()?;
```

PR #226, opened 2019-03-15
zkvm: simplify `unblind`

### LINKED ISSUE #212: zkvm: simplify `unblind`
Currently [`unblind`](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-spec.md#unblind) does a lot of things:

1. Checks whether a commitment `V` is unblinded `v*B`.
2. Creates a variable with `V`.
3. Adds a constraint to a provided expression `expr` that enforces `expr == V`.

We probably do not need steps (2) and (3). Just having commitment `V` and its opening `v` on the input is enough. The additional code can then construct constraints to `V` or `borrow` it, as needed.

debnil:
Closes #212

### DIFF
--- a/zkvm/src/vm.rs
+++ b/zkvm/src/vm.rs
@@ -440,11 +440,10 @@ where
         let scalar_witness = self.pop_item()?.to_data()?.to_scalar()?;
         let v_scalar = scalar_witness.to_scalar();
 
-        // Peek commitment `V`
-        if self.stack.len() == 0 {
-            return Err(VMError::StackUnderflow);
-        }
-        let v_commitment = &self.stack[0].to_data()?.to_commitment()?;
+        // Pop commitment `V`
+        let v_commitment_data = self.pop_item()?.to_data()?;
+        let v_commitment_push = v_commitment_data.clone();
+        let v_commitment = v_commitment_data.to_commitment()?;
         let v_point = v_commitment.to_point();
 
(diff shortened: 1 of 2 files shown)


zkvm/src/vm.rs
```diff
@@ -436,17 +436,15 @@ where
...
-        // Pop expression `expr`
-        let expr = self.pop_item()?.to_expression()?;
-
-        // Pop commitment `V`
-        let v_commitment = self.pop_item()?.to_data()?.to_commitment()?;
-        let v_point = v_commitment.to_point();
-
         // Pop scalar `v`
         let scalar_witness = self.pop_item()?.to_data()?.to_scalar()?;
         let v_scalar = scalar_witness.to_scalar();
```
oleganza: let's make this impl more readable by reducing number of named entities.

E.g. 

```rust
let v = self.pop_item()?.to_data()?.to_scalar()?.to_scalar();
let V = self.pop_item()?.to_data()?.to_commitment()?.to_point();
... verify V == v*B
self.push_item(Data::Opaque(V.as_bytes().to_vec()));
```

oleganza: `V` is `Copy`, so it'll be copied for the last line automatically via `memcpy` by compiler.

debnil: Nice, thanks!

Issue #179, opened 2019-03-01
keytree: serialization for xpub, xprv


oleganza:
For both `Xprv` and `Xpub` types:

* `::to_bytes(&self) -> [u8; 64]`
* `::from_bytes(&[u8]) -> Option<Self>` (checking that slice is exactly 64-byte-long)

For later: implement Serde on top of these ([example](https://github.com/dalek-cryptography/ed25519-dalek/blob/master/src/public.rs#L249-L285))

Note: the `from_bytes` won't benefit from using fixed-length array (`[u8; 64]`) because at least `Xprv` would still have to check if scalar is canonically encoded. So we might as well make decoding of all of these failable and take byteslices to let user not to make extra 64-byte copies.

tessr:
What should be serialized/deserialized here? For the private key, is the deserialized form composed of the deserialized scalar (a u8 of length 32) followed by the deserialized derivation key (also a u8 of length 32)? For the pub key, is it the same thing but the deserialized point instead of the scalar?

oleganza:
Yes:

`xprv = <scalar 32 bytes> <dk 32 bytes>`
`xpub = <point 32 bytes> <dk 32 bytes>`

Now that Xpub holds a decompressed point, the `Option<Xpub>` returned by decoding `&[u8]` can also fail because `decompress()` fails (not only on len()!=64).

oleganza:
(clicked wrong button)

tessr:
Resolved by https://github.com/interstellar/slingshot/pull/228

PR #228, opened 2019-03-18
keytree: add serialization for xpub, xprv

tessr:
Addresses https://github.com/interstellar/slingshot/issues/179

### DIFF
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -67,7 +67,7 @@ impl Xprv {
         }
 
         let pieces = bytes.split_at(32);
-        let mut scalar_bytes: [u8; 32] = Default::default();
+        let mut scalar_bytes = [0u8; 32];
         scalar_bytes.copy_from_slice(&pieces.0[..]);
         let scalar = match Scalar::from_canonical_bytes(scalar_bytes) { Some(x)=>x, None => return None };
         let mut dk: [u8; 32] = Default::default();
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -126,7 +126,7 @@ impl Xpub {
 
         let pieces = bytes.split_at(32);
         let precompressed_pubkey = CompressedRistretto::from_slice(&pieces.0[..]);
-        let mut dk: [u8; 32] = Default::default();
+        let mut dk = [0u8; 32];
         dk.copy_from_slice(&pieces.1[..]);
 
         match precompressed_pubkey.decompress() {
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -69,7 +69,7 @@ impl Xprv {
         let pieces = bytes.split_at(32);
         let mut scalar_bytes: [u8; 32] = Default::default();
         scalar_bytes.copy_from_slice(&pieces.0[..]);
-        let scalar = Scalar::from_bits(scalar_bytes);
+        let scalar = match Scalar::from_canonical_bytes(scalar_bytes) { Some(x)=>x, None => return None };
         let mut dk: [u8; 32] = Default::default();
         dk.copy_from_slice(&pieces.1[..]);
         let precompressed_pubkey = (scalar * &constants::RISTRETTO_BASEPOINT_POINT).compress();
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -69,7 +69,10 @@ impl Xprv {
         let pieces = bytes.split_at(32);
         let mut scalar_bytes = [0u8; 32];
         scalar_bytes.copy_from_slice(&pieces.0[..]);
-        let scalar = match Scalar::from_canonical_bytes(scalar_bytes) { Some(x)=>x, None => return None };
+        let scalar = match Scalar::from_canonical_bytes(scalar_bytes) {
+            Some(x) => x,
+            None => return None,
+        };
         let mut dk = [0u8; 32];
         dk.copy_from_slice(&pieces.1[..]);
         let precompressed_pubkey = (scalar * &constants::RISTRETTO_BASEPOINT_POINT).compress();
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -127,7 +127,7 @@ impl Xpub {
             return None;
         }
 
-        let pieces = bytes.split_at(32);
+        let (pkslice, dkslice) = bytes.split_at(32);
         let precompressed_pubkey = CompressedRistretto::from_slice(&pieces.0[..]);
         let mut dk = [0u8; 32];
         dk.copy_from_slice(&pieces.1[..]);
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -132,17 +132,15 @@ impl Xpub {
         let mut dk = [0u8; 32];
         dk.copy_from_slice(&pieces.1[..]);
 
-        match precompressed_pubkey.decompress() {
+        let point = match precompressed_pubkey.decompress() {
+            Some(p) => p,
             None => return None,
-            Some(p) => {
-                let point = p;
-                return Some(Xpub {
-                    point,
-                    dk,
-                    precompressed_pubkey,
-                });
-            }
-        }
+        };
+        Some(Xpub {
+            point,
+            dk,
+            precompressed_pubkey,
+        })
     }
 }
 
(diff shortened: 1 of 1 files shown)


keytree/src/lib.rs
```diff
@@ -50,6 +50,36 @@ impl Xprv {
...
+    /// exactly 64 bytes.
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        if bytes.len() != 64 {
+            return None;
+        }
+
+        let pieces = bytes.split_at(32);
+        let mut scalar_bytes: [u8; 32] = Default::default();
+        scalar_bytes.copy_from_slice(&pieces.0[..]);
+        let scalar = Scalar::from_bits(scalar_bytes);
```
oleganza: `Scalar::from_bits` is a function for specific use of bit-twiddling-infested protocols originally spawned by djb (x25519 and EdDSA). I'd recommend using `Scalar::from_canonical_bytes` that checks that a scalar is canonical. Our decoding func is already failable, so it's not going to make API uglier to fail on non-canonical scalars.

```suggestion
        let scalar = match Scalar::from_canonical_bytes(scalar_bytes) { Some(x)=>x, None => return None };
```

keytree/src/lib.rs
```diff
@@ -50,6 +50,36 @@ impl Xprv {
...
+
+    /// Decodes an Xprv from a 64-byte array, and fails if the provided array is not
+    /// exactly 64 bytes.
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        if bytes.len() != 64 {
+            return None;
+        }
+
+        let pieces = bytes.split_at(32);
+        let mut scalar_bytes: [u8; 32] = Default::default();
```
oleganza: Arrays have this syntax for initializing with a repeated value:
```rust
        let mut scalar_bytes: [u8; 32] = [0; 32];
```
And if you clarify the integer type, you can even avoid declaring the variable type separately:

```suggestion
        let mut scalar_bytes = [0u8; 32];
```

keytree/src/lib.rs
```diff
@@ -50,6 +50,36 @@ impl Xprv {
...
+        scalar_bytes.copy_from_slice(&pieces.0[..]);
+        let scalar = Scalar::from_bits(scalar_bytes);
+        let mut dk: [u8; 32] = Default::default();
```
oleganza: same as above:
```suggestion
        let mut dk = [0u8; 32];
```

keytree/src/lib.rs
```diff
@@ -78,6 +108,39 @@ impl Xpub {
...
+    /// Decodes an Xpub from a 64-byte array, and fails if the provided array is not
+    /// exactly 64 bytes, or if the compressed point fails to decompress.
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        if bytes.len() != 64 {
+            return None;
+        }
+
+        let pieces = bytes.split_at(32);
+        let precompressed_pubkey = CompressedRistretto::from_slice(&pieces.0[..]);
+        let mut dk: [u8; 32] = Default::default();
```
oleganza:
```suggestion
        let mut dk = [0u8; 32];
```

keytree/src/lib.rs
```diff
@@ -78,6 +108,39 @@ impl Xpub {
...
+        dk.copy_from_slice(&pieces.1[..]);
+
+        match precompressed_pubkey.decompress() {
```
oleganza: i'd factor out decompressing from Xpub creation to make code less entangled and easier to read:

```rust
let point = match precompressed_pubkey.decompress() { Some(p)=>p, None => return None };
Some(Xpub {
    point, dk, precompressed_pubkey
})
```

keytree/src/lib.rs
```diff
@@ -106,6 +169,42 @@ mod tests {
...
 
+    #[test]
+    fn serialize_xprv_test() {
+        let seed = [0u8; 32];
+        let mut rng = ChaChaRng::from_seed(seed);
+        let xprv = Xprv::random(&mut rng);
+        let xprv_bytes = xprv.to_bytes();
+
+        // hardcoded, but happens to be expected_scalar concatenated with expected_dk
+        let expected_bytes = [
```
oleganza: there's `hex` crate with `hex::encode` method so we can write a more readable test (and then copy this over to the spec as Test Vectors):

```rust
assert_eq!(hex::encode(&xprv.to_bytes()), "df0da2b88d88de1ad7678d8f012...");
```

oleganza: `hex` is only needed in `dev-dependencies` in Cargo.toml. So you'd include it only under the `mod test { }`, not in the top of lib.rs.

oleganza:
looks good, just left a bunch of stylistic rusty nitpicks

tessr:
Awesome. I'm going to commit your suggestions, fix any breakages, and then make the bigger changes. Thanks very much!

keytree/src/lib.rs
tessr: I just tried this out, and it's neat, but I'm not sure it actually produces a more readable test, especially since all the other tests in this package (for creating xprvs and xpubs) use non-hex encoded byte arrays. When I switched over to a hex string, I could no longer eyeball that these were the same. 

WDYT?

oleganza: i think all have to be hex, across the board. But if it's a PITA to update, we can do that in a separate PR later.

oleganza: we'd definitely want that to post the test vectors in the spec in hex. But before that we need to implement derivation for both xpub and xprv.

keytree/src/lib.rs
```diff
@@ -78,6 +111,37 @@ impl Xpub {
...
+    }
+
+    /// Decodes an Xpub from a 64-byte array, and fails if the provided array is not
+    /// exactly 64 bytes, or if the compressed point fails to decompress.
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        if bytes.len() != 64 {
+            return None;
+        }
+
+        let pieces = bytes.split_at(32);
```
oleganza:
```suggestion
        let (pkslice, dkslice) = bytes.split_at(32);
```

then you can use those instead of .0 and .1 - easier to read.

keytree/src/lib.rs
```diff
@@ -50,6 +50,39 @@ impl Xprv {
...
+    }
+
+    /// Decodes an Xprv from a 64-byte array, and fails if the provided array is not
+    /// exactly 64 bytes.
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        if bytes.len() != 64 {
+            return None;
+        }
+
+        let pieces = bytes.split_at(32);
```
oleganza:
```suggestion
        let (pkslice, dkslice) = bytes.split_at(32);
```

see comment below

keytree/src/lib.rs
tessr: I'll do it all in another PR, then. (It's fine to do, I should just change them all together.)

keytree/src/lib.rs
tessr: oh, good! this is much nicer.

tessr:
Ready for an (I think) final pass!

keytree/src/lib.rs
```diff
@@ -50,6 +50,39 @@ impl Xprv {
...
+    }
+
+    /// Decodes an Xprv from a 64-byte array, and fails if the provided array is not
+    /// exactly 64 bytes.
+    pub fn from_bytes(bytes: &[u8]) -> Option<Self> {
+        if bytes.len() != 64 {
+            return None;
+        }
+
+        let (scslice, dkslice) = bytes.split_at(32);
```
oleganza: sorry, i overlooked it previously. You probably don't need this explicit splitting since you can slice things up in place:

```rust
scalar_bytes.copy_from_slice(&bytes[..32]);
...
dk.copy_from_slice(&bytes[32..]);
```

And the same for xpub parsing.

keytree/src/lib.rs
```diff
@@ -50,6 +50,39 @@ impl Xprv {
...
+        let mut scalar_bytes = [0u8; 32];
+        scalar_bytes.copy_from_slice(&scslice[..]);
```
oleganza: `scslice` is already a slice, so it can be used directly w/o extra reslicing (it's cheap, but still needless operation):

```suggestion
        scalar_bytes.copy_from_slice(scslice);
```

however, see the above comment - i suggest slicing the `bytes` in place instead of `split_at`.

tessr:
Ready for another pass! (I will not jinx it this time by suggesting it's final.)

PR #229, opened 2019-03-19
MuSig: add counterparties

cathieyun:
- [x] fill out counterparty details in eng design doc
- [x] implement counterparty states & transitions
- [x] small fix: rename error to `MuSigShareError`

MuSig work outline here:
https://github.com/interstellar/slingshot/issues/92

### DIFF
--- a/zkvm/src/signature/counterparty.rs
+++ b/zkvm/src/signature/counterparty.rs
@@ -82,7 +82,7 @@ impl CounterpartyPrecommitted {
 
 impl CounterpartyCommitted {
     pub fn sign(
-        &self,
+        self,
         share: Scalar,
         challenge: Scalar,
         multikey: &Multikey,
--- a/zkvm/src/signature/counterparty.rs
+++ b/zkvm/src/signature/counterparty.rs
@@ -15,12 +15,20 @@ pub struct NoncePrecommitment([u8; 32]);
 pub struct NonceCommitment(RistrettoPoint);
 
 impl NonceCommitment {
-    fn precommit(&self) -> NoncePrecommitment {
+    pub fn new(commitment: RistrettoPoint) -> Self {
+        NonceCommitment(commitment)
+    }
+
+    pub fn precommit(&self) -> NoncePrecommitment {
         let mut h = Transcript::new(b"MuSig.nonce-precommit");
         h.commit_point(b"R", &self.0.compress());
         let mut precommitment = [0u8; 32];
         h.challenge_bytes(b"precommitment", &mut precommitment);
         NoncePrecommitment(precommitment)
+    }
+
+    pub fn sum(commitments: &Vec<Self>) -> RistrettoPoint {
+        commitments.iter().map(|R_i| R_i.0).sum()
     }
 }
 
@@ -54,7 +62,7 @@ impl Counterparty {
 impl CounterpartyPrecommitted {
     pub fn commit_nonce(
         self,
-        commitment: NonceCommitment,
+        commitment: &NonceCommitment,
     ) -> Result<CounterpartyCommitted, VMError> {
         // Check H(commitment) =? precommitment
         let received_precommitment = commitment.precommit();
@@ -66,7 +74,7 @@ impl CounterpartyPrecommitted {
         }
 
         Ok(CounterpartyCommitted {
-            commitment,
+            commitment: *commitment,
             pubkey: self.pubkey,
         })
     }
@@ -74,11 +82,11 @@ impl CounterpartyPrecommitted {
 
 impl CounterpartyCommitted {
     pub fn sign(
-        self,
+        &self,
         share: Scalar,
         challenge: Scalar,
         multikey: &Multikey,
-    ) -> Result<(Scalar, NonceCommitment), VMError> {
+    ) -> Result<Scalar, VMError> {
         // Check if s_i * G == R_i + c * a_i * X_i.
         // - s_i = share
         // - G = RISTRETTO_BASEPOINT_POINT
@@ -96,6 +104,6 @@ impl CounterpartyCommitted {
             });
         }
 
-        Ok((share, self.commitment))
+        Ok(share)
     }
 }
--- a/zkvm/src/signature/counterparty.rs
+++ b/zkvm/src/signature/counterparty.rs
@@ -15,11 +15,11 @@ pub struct NoncePrecommitment([u8; 32]);
 pub struct NonceCommitment(RistrettoPoint);
 
 impl NonceCommitment {
-    pub fn new(commitment: RistrettoPoint) -> Self {
+    pub(super) fn new(commitment: RistrettoPoint) -> Self {
         NonceCommitment(commitment)
     }
 
-    pub fn precommit(&self) -> NoncePrecommitment {
+    pub(super) fn precommit(&self) -> NoncePrecommitment {
         let mut h = Transcript::new(b"MuSig.nonce-precommit");
         h.commit_point(b"R", &self.0.compress());
         let mut precommitment = [0u8; 32];
@@ -27,7 +27,7 @@ impl NonceCommitment {
         NoncePrecommitment(precommitment)
     }
 
-    pub fn sum(commitments: &Vec<Self>) -> RistrettoPoint {
+    pub(super) fn sum(commitments: &Vec<Self>) -> RistrettoPoint {
         commitments.iter().map(|R_i| R_i.0).sum()
     }
 }
@@ -47,11 +47,14 @@ pub struct CounterpartyCommitted {
 }
 
 impl Counterparty {
-    pub fn new(pubkey: VerificationKey) -> Self {
+    pub(super) fn new(pubkey: VerificationKey) -> Self {
         Counterparty { pubkey }
     }
 
-    pub fn precommit_nonce(self, precommitment: NoncePrecommitment) -> CounterpartyPrecommitted {
+    pub(super) fn precommit_nonce(
+        self,
+        precommitment: NoncePrecommitment,
+    ) -> CounterpartyPrecommitted {
         CounterpartyPrecommitted {
             precommitment,
             pubkey: self.pubkey,
@@ -60,7 +63,7 @@ impl Counterparty {
 }
 
 impl CounterpartyPrecommitted {
-    pub fn commit_nonce(
+    pub(super) fn commit_nonce(
         self,
         commitment: NonceCommitment,
     ) -> Result<CounterpartyCommitted, VMError> {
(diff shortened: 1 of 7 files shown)


zkvm/src/signature/signer.rs
```diff
@@ -165,28 +155,15 @@ impl PartyAwaitingShares {
...
-            let X_i = X_i.0.decompress().ok_or(VMError::InvalidPoint)?;
-
-            // Check that S_i = R_i + c * a_i * X_i
-            if S_i != R_i + self.c * a_i * X_i {
-                return Err(VMError::ShareError { index: i });
-            }
-        }
+        let validated_shares = self
+            .counterparties
+            .iter()
```
cathieyun: Help! When I try to take ownership of `self.counterparties` by using `into_iter()` instead of `iter`, I then no longer have ownership over `self` in future steps. This prevents me from using `self.c`, `self.multikey` inside the map, as well as prevents me from calling `self.receive_trusted_shares(...)`. 

So, for now I am changing `CounterpartyCommitted::sign(...)` to take `&self` instead of `self`. This is not great (since then you can call this function on the struct multiple times, which is not *dangerous* but is just incorrect behavior). Is there a way to take ownership of `self.counterparties` without taking ownership of all of `self`, and therefore being able to take ownership of other fields of `self`?

(To get the errors I'm talking about, remove the `&` from `&self` in the inputs of `CounterpartyCommitted::sign(...)`.)

JeremyRubin: Simple solution: std::mem::swap with an empty vector

zkvm/src/signature/counterparty.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    pub fn precommit(&self) -> NoncePrecommitment {
+        let mut h = Transcript::new(b"MuSig.nonce-precommit");
+        h.commit_point(b"R", &self.0.compress());
+        let mut precommitment = [0u8; 32];
+        h.challenge_bytes(b"precommitment", &mut precommitment);
+        NoncePrecommitment(precommitment)
+    }
+
+    pub fn sum(commitments: &Vec<Self>) -> RistrettoPoint {
```
oleganza: most rusty (eye-rolling) way to do this is

```suggestion
    pub fn sum<I>(commitments: I) -> RistrettoPoint 
    where 
    I: IntoIterator,
    I::Item: Borrow<RistrettoPoint>
    {
```

and then `commitments.into_iter().map(|p|p.borrow()).sum()`. You can then still pass &Vec into this method since it implements IntoIterator.

zkvm/src/signature/counterparty.rs
```diff
@@ -0,0 +1,109 @@
...
+        let received_precommitment = commitment.precommit();
+        let equal = self.precommitment.0.ct_eq(&received_precommitment.0);
+        if equal.unwrap_u8() == 0 {
+            return Err(VMError::MuSigShareError {
+                pubkey: self.pubkey.0.to_bytes(),
+            });
+        }
+
+        Ok(CounterpartyCommitted {
+            commitment: *commitment,
```
oleganza: Since you are copying the commitment here, you can as well move it in and let the user scope to copy it if needed.

zkvm/src/signature/counterparty.rs
```diff
@@ -0,0 +1,109 @@
...
+            precommitment,
+            pubkey: self.pubkey,
+        }
+    }
+}
+
+impl CounterpartyPrecommitted {
+    pub fn commit_nonce(
+        self,
+        commitment: &NonceCommitment,
```
oleganza:
```suggestion
        commitment: NonceCommitment,
```

see comment below

zkvm/src/signature/counterparty.rs
```diff
@@ -0,0 +1,109 @@
...
+        Ok(CounterpartyCommitted {
+            commitment: *commitment,
+            pubkey: self.pubkey,
+        })
+    }
+}
+
+impl CounterpartyCommitted {
+    pub fn sign(
+        &self,
```
oleganza: Do we need to preserve the counterparty state after receiving the signature? If it gave us bad signature, we need to restart the whole protocol from scratch (since nonces are burned), and if it succeeded, we can return the share (and other data if needed), destroying the counterparty state.

zkvm/src/signature/signer.rs
oleganza: The issue here is not that you need `self.c` and `self.multikey` later - you can move out fields in arbitrary order. The issue is that the `map`'s closure captures `self`, not its individual fields like it seems at first.

So to avoid this problem, you'd need to first move out `self.c` and `self.multikey` in local variables outside the closure, and then capture them in the closure, so `self` is not referenced by anyone else. This should allow you to do `self.counterparties.into_iter()` and have clean move semantics throughout.

oleganza: Of course, there's an RFC for that already: https://github.com/rust-lang/rfcs/pull/2229 + https://github.com/rust-lang/rust/issues/53488 

Meanwhile, this is what i found:
* http://smallcultfollowing.com/babysteps/blog/2018/04/24/rust-pattern-precise-closure-capture-clauses/
* https://github.com/rust-lang/rust/issues/19004
* https://stevedonovan.github.io/rustifications/2018/08/18/rust-closures-are-hard.html

Especially a neat trick in the first article:

image

TL;DR: you move out those fields into local variables inside the block (not closure) located where the closure needs to be, which returns a closure that uses those local variables. So it feels like it the closure with "explicit capture annotations" using standard syntax features.

zkvm/src/signature/counterparty.rs
cathieyun: done!

zkvm/src/signature/counterparty.rs
cathieyun: Makes sense, done!

zkvm/src/signature/counterparty.rs
cathieyun: Is there a benefit (performance, memory-wise) to this approach?
Or is it just for the sake of the rusty eye-rolls? :P

zkvm/src/signature/counterparty.rs
cathieyun: The only reason this takes `&self` instead of `self` is to help with the trouble I was having here:
https://github.com/interstellar/slingshot/pull/229#discussion_r267110317

(I would very much like to figure out a way to make this take `self` and still compile correctly!)

zkvm/src/signature/signer.rs
cathieyun: I tried what you suggested (moving out the fields into local variables), but the problem is that after the map, I still want to be able to call `self.receive_trusted_shares(...)`. And I can't do that because the map's closure captures `self`, so I no longer have ownership of `self` to call `self.receive_trusted_shares(...)` on.

I will read into the articles you linked, maybe those will have something! Otherwise it might just make sense to copy the logic in `receive_trusted_shares()` and put it in `receive_shares()` to avoid the problem of calling `self.receive_trusted_shares` without ownership of `self`.

zkvm/src/signature/counterparty.rs
oleganza: if it's going to be a public API, it allows users pass a wider variety of things in. E.g. if they already have an iterator of owned or borrowed points, they can pass it in w/o doing any allocations (perf benefit) or conversions (ergonomics benefit).

oleganza: if it's a private API (maybe it should be - it's only used within Party impl?) - then `&[u8]` is simple and unassuming enough.

zkvm/src/signature/signer.rs
oleganza: @cathieyun don't call `receive_trusted_shares` :-) it's a one-liner that sums up all shares, so you can just do that directly in the `receive_shares`.

oleganza: moreover, it's more efficient that collecting into Result<Vec<_>>, if you simply end the giant iterator with .sum(). And the sum over iteration of Results is defined in stdlib to return Result of the sum!

image

zkvm/src/signature/counterparty.rs
cathieyun: done!

zkvm/src/signature/signer.rs
```diff
@@ -165,28 +155,22 @@ impl PartyAwaitingShares {
...
-                return Err(VMError::ShareError { index: i });
-            }
-        }
+        // Check that all shares are valid
+        let validated_shares = self
+            .counterparties
+            .into_iter()
+            .zip(shares)
+            .map(|(counterparty, share)| counterparty.sign(share, challenge, multikey))
+            .collect::<Result<Vec<_>, _>>()?;
```
oleganza: wouldn't this work directly, without allocating another Vec?
```suggestion
            .sum()?;
```

zkvm/src/signature/signer.rs
```diff
@@ -165,28 +155,22 @@ impl PartyAwaitingShares {
...
-            let X_i = pubkeys[i];
-            let R_i = self.nonce_commitments[i].0;
-
-            // Make a_i = H(L, X_i)
-            let a_i = self.multikey.factor_for_key(&VerificationKey(X_i.0));
-
-            let X_i = X_i.0.decompress().ok_or(VMError::InvalidPoint)?;
+    pub fn receive_shares(self, shares: Vec<Scalar>) -> Result<Signature, VMError> {
+        // Create local copies of challenge and &multikey,
+        // since iterator takes ownership of `self` later.
```
oleganza: These are not local copies, these are moved-out fields, because compiler is not smart enough to capture them as moves in the map's closure.

how about:

```rust
// Move out self's fields because `self.c` inside `map`'s closure would
// lead to capturing `self` by reference, while we want
// to move `self.counterparties` out of it.
// See also RFC2229: https://github.com/rust-lang/rfcs/pull/2229
```

zkvm/src/signature/counterparty.rs
cathieyun: I actually think that all of the counterparty structs & API should be only usable from within the module, which I denoted by making them all `pub(super)`. So, this should take care of the concern about flexibility in the case of a public API (which it's not)

zkvm/src/signature/signer.rs
cathieyun: thanks for clarification, will update!

zkvm/src/signature/signer.rs
cathieyun: Done!

zkvm/src/signature/signer.rs
cathieyun: That's awesome, I had no idea you could do a sum over iter of Results! Nice :)

PR #231, opened 2019-03-19
keytree: express test data as hex strings

tessr:
Per conversation with @oleganza, test data is more readable as a hex string (rather than a byte slice). This does introduce a bit more boilerplate around encoding and decoding, so I'm a bit apathetic on that.

### DIFF
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -271,11 +271,11 @@ mod tests {
         assert_eq!(xpub.precompressed_pubkey, expected_compressed_point);
     }
 
-    fn to_hex_32(input: [u8; 32])->String {
+    fn to_hex_32(input: [u8; 32]) -> String {
         return hex::encode(&input[..]);
     }
 
-    fn to_hex_64(input: [u8; 64])->String {
+    fn to_hex_64(input: [u8; 64]) -> String {
         return hex::encode(&input[..]);
     }
 
--- a/keytree/src/lib.rs
+++ b/keytree/src/lib.rs
@@ -146,10 +146,9 @@ impl Xpub {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use hex;
     use rand::SeedableRng;
     use rand_chacha::ChaChaRng;
-
-    use hex;
 
     #[test]
     fn random_xprv_test() {
@@ -158,17 +157,14 @@ mod tests {
         let xprv = Xprv::random(&mut rng);
 
         // the following are hard-coded based on the previous seed
-        let dk_hex = "9f07e7be5551387a98ba977c732d080dcb0f29a048e3656912c6533e32ee7aed";
-        let scalar_hex = "4a53c3fbbc59970ee5f85af813875dffc13a904a2e53ae7e65fa0dea6e62c901";
-
-        let expected_dk = hex::decode(dk_hex).unwrap();
-
-        let mut expected_scalar_bytes = [0u8; 32];
-        expected_scalar_bytes[..].copy_from_slice(&hex::decode(scalar_hex).unwrap());
-        let expected_scalar = Scalar::from_bits(expected_scalar_bytes);
-
-        assert_eq!(expected_dk, xprv.dk);
-        assert_eq!(expected_scalar, xprv.scalar);
+        assert_eq!(
+            hex::encode(&xprv.dk),
+            "9f07e7be5551387a98ba977c732d080dcb0f29a048e3656912c6533e32ee7aed"
+        );
+        assert_eq!(
+            hex::encode(xprv.scalar.as_bytes()),
+            "4a53c3fbbc59970ee5f85af813875dffc13a904a2e53ae7e65fa0dea6e62c901"
+        );
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


keytree/src/lib.rs
```diff
@@ -148,21 +149,23 @@ mod tests {
...
-            160, 72, 227, 101, 105, 18, 198, 83, 62, 50, 238, 122, 237,
-        ];
-        let expected_scalar = Scalar::from_bits([
-            74, 83, 195, 251, 188, 89, 151, 14, 229, 248, 90, 248, 19, 135, 93, 255, 193, 58, 144,
-            74, 46, 83, 174, 126, 101, 250, 13, 234, 110, 98, 201, 1,
-        ]);
+        let dk_hex = "9f07e7be5551387a98ba977c732d080dcb0f29a048e3656912c6533e32ee7aed";
+        let scalar_hex = "4a53c3fbbc59970ee5f85af813875dffc13a904a2e53ae7e65fa0dea6e62c901";
+
+        let expected_dk = hex::decode(dk_hex).unwrap();
```
oleganza: Generally, in tests one does not need to decode from hex, only encode.

Why not writing something like this for brevity:

```rust
assert_eq!(
    hex::encode(&xprv.dk),
    "9f07e7be5551387a98ba977c732d080dcb0f29a048e3656912c6533e32ee7aed"
);
assert_eq!(
    hex::encode(xprv.scalar.as_bytes()),
    "4a53c3fbbc59970ee5f85af813875dffc13a904a2e53ae7e65fa0dea6e62c901"
);
```

tessr: > Why not writing something like this

I took some pains in this PR to try to make all of the assert statements as short and obvious as possible (i.e. always comparing exactly the thing I was testing). But you're right, that makes these tests very long.

keytree/src/lib.rs
```diff
@@ -174,25 +173,18 @@ mod tests {
         let mut rng = ChaChaRng::from_seed(seed);
         let xprv = Xprv::random(&mut rng);
         let xprv_bytes = xprv.to_bytes();
+        let xprv_hex = hex::encode(&xprv_bytes[..]);
```
oleganza: My personal preference would be to not create extra variable for something mundane as convert-to-hex, and inline it inside `assert_eq!`, wdyt?

```rust
assert_eq!(
    hex::encode(&xprv.to_bytes()[..]),
    "4a53c3fbbc59970ee5f85af813875dffc13a904a2e53ae7e65fa0dea6e62c9019f07e7be5551387a98ba977c732d080dcb0f29a048e3656912c6533e32ee7aed"
);
```

tessr: Oh sure, I just missed this one

oleganza: or factor out a `fn to_hex` converter inside `mod test` so it looks like this:

```
assert_eq!(
    to_hex(xprv.to_bytes()),
    "4a53c3fbbc59970ee5f85af813875dffc13a904a2e53ae7e65fa0dea6e62c9019f07e7be5551387a98ba977c732d080dcb0f29a048e3656912c6533e32ee7aed"
);
```

keytree/src/lib.rs
```diff
@@ -174,25 +173,18 @@ mod tests {
...
     #[test]
     fn deserialize_xprv_test() {
-        let xprv_bytes = [
-            74, 83, 195, 251, 188, 89, 151, 14, 229, 248, 90, 248, 19, 135, 93, 255, 193, 58, 144,
-            74, 46, 83, 174, 126, 101, 250, 13, 234, 110, 98, 201, 1, 159, 7, 231, 190, 85, 81, 56,
-            122, 152, 186, 151, 124, 115, 45, 8, 13, 203, 15, 41, 160, 72, 227, 101, 105, 18, 198,
-            83, 62, 50, 238, 122, 237,
-        ];
+        let xprv_hex =  "4a53c3fbbc59970ee5f85af813875dffc13a904a2e53ae7e65fa0dea6e62c9019f07e7be5551387a98ba977c732d080dcb0f29a048e3656912c6533e32ee7aed";
+        let xprv_bytes = hex::decode(xprv_hex).unwrap();
```
tessr: @oleganza I know you said there shouldn't be any decoding in these tests, but I think it's unavoidable here, given that `from_bytes` expects, well, a byte array. 

(I'll also update this to inline `xprv_hex`.)

keytree/src/lib.rs
```diff
@@ -211,18 +203,16 @@ mod tests {
...
-            160, 72, 227, 101, 105, 18, 198, 83, 62, 50, 238, 122, 237,
-        ];
-        let expected_compressed_point = CompressedRistretto::from_slice(&[
-            156, 102, 163, 57, 200, 52, 79, 146, 47, 195, 32, 108, 181, 218, 232, 20, 165, 148,
-            192, 23, 125, 211, 35, 92, 37, 77, 156, 64, 154, 101, 184, 8,
-        ]);
+        // hard-coded based on the previous seed
+        let compressed_pt_hex = "9c66a339c8344f922fc3206cb5dae814a594c0177dd3235c254d9c409a65b808";
+        let expected_compressed_point =
+            CompressedRistretto::from_slice(&hex::decode(compressed_pt_hex).unwrap());
```
tessr: Likewise, I don't think I can avoid this decoding. (And, again, I'll inline `compressed_pt_hex`.)

keytree/src/lib.rs
tessr: Oh! I like this very much.

keytree/src/lib.rs
oleganza: A hypothetical `to_hex64` helper can take `[u8;64]`. But `&to_bytes()[..]` is not too bad imho.

tessr:
alright, PTAL

keytree/src/lib.rs
```diff
@@ -211,18 +200,17 @@ mod tests {
...
+                .unwrap(),
+        );
         let expected_point = expected_compressed_point.decompress().unwrap();
 
-        assert_eq!(xpub.dk, expected_dk);
+        assert_eq!(
+            to_hex_32(xpub.dk),
+            "9f07e7be5551387a98ba977c732d080dcb0f29a048e3656912c6533e32ee7aed"
+        );
         assert_eq!(xpub.point, expected_point);
```
oleganza: why not simply do?

```rust
assert_eq!(xpub.point.compress(), xpub.precompressed_pubkey); // checks internal consistency
assert_eq!(hex::encode(&xpub.precompressed_pubkey.to_bytes()[..]), "9c66a339c8344f922fc3206cb5dae814a594c0177dd3235c254d9c409a65b808");
```

tessr:
PTAL 💃

PR #232, opened 2019-03-20
zkvm: make VerificationKey uncompressed

### LINKED ISSUE #219: zkvm: make VerificationKey uncompressed
We can streamline a few APIs around `VerificationKey` by making them non-failable by pushing decoding/decompression further from the protocol logic. However, point compression remains being relatively pricey, so if we initialize with a compressed point, we need to preserve it.

Suggestion:

```rust
#[derive(Copy,Clone,Debug)]
struct VerificationKey {
    point: RistrettoPoint,
    compressed: Option<CompressedRistretto>, // available only if instantiated with compressed one
}

impl VerificationKey {
    /// Creates new key from a compressed form, remembers the compressed point.
    pub fn with_compressed(p: CompressedRistretto) -> Option<Self>;
    /// Creates with a decompressed point, does not eagerly compress it.
    pub fn with_decompressed(p: RistrettoPoint) -> Self;
    
    /// Compresses the point if needed (no-op if `self.compressed == Some`)
    pub fn compress_in_place(&mut self);

    pub fn to_point(self) -> RistrettoPoint;

    /// Compresses if needed
    pub fn to_compressed_point(self) -> CompressedRistretto;

    /// Compresses if needed, avoids copying uncompressed one, but cannot cache in place.
    pub fn as_compressed_point(&self) -> &CompressedRistretto;
}

impl PartialEq {
   /// TBD: implement partial eq over self.point, ignoring optional compressed form
}
```

Once we have this API, we can revisit all cases where we do `key.0.decompress().ok_or(...)?` and try to make those callers non-failable. Like, `Multikey`, for instance.

Let's do this after #190.

decentralisedkev:
Closes #219


zkvm/src/signature/mod.rs
```diff
@@ -16,11 +16,59 @@ mod multikey;
...
+pub struct VerificationKey {
+    point: RistrettoPoint,
+    // Available only if instantiated using `with_compressed`
+    // or if caller directly calls a method that returns a compressed point
+    compressed: Option<CompressedRistretto>,
+}
 
+impl VerificationKey {
+    /// Creates new key from a compressed form,remembers the compressed point.
+    pub fn with_compressed(p: CompressedRistretto) -> Option<Self> {
```
oleganza:
```suggestion
    pub fn from_point(p: CompressedRistretto) -> Option<Self> {
```

maybe this sounds better?

zkvm/src/signature/mod.rs
```diff
@@ -16,11 +16,59 @@ mod multikey;
...
+        Some(VerificationKey {
+            point: p.decompress()?,
+            compressed: Some(p),
+        })
+    }
+    /// Creates a new key from a ristretto point. Does not eagerly compress.
+    pub fn with_decompressed(p: RistrettoPoint) -> Self {
```
oleganza: let's remove this method and `impl From<RistrettoPoint> for VerificationKey` instead?

zkvm/src/signature/mod.rs
```diff
@@ -16,11 +16,59 @@ mod multikey;
...
-/// Verification key (aka "pubkey") is a wrapper type around a Ristretto point
+/// Verification key (aka "pubkey") is a wrapper type around a Ristretto point and it's compressed variation
 /// that lets the verifier to check the signature.
 #[derive(Copy, Clone, PartialEq, Eq, Debug)]
-pub struct VerificationKey(pub CompressedRistretto);
+pub struct VerificationKey {
+    point: RistrettoPoint,
+    // Available only if instantiated using `with_compressed`
+    // or if caller directly calls a method that returns a compressed point
+    compressed: Option<CompressedRistretto>,
```
oleganza: i'm thinking that we often have to use compressed point because we commit it somewhere - Keytree needs it to compute the additive factor, Multikey needs it to compute multiplicative factors. So let's simplify the logic and always store compressed point alongside decompressed one? And rename the field to reflect the fact that it's an optimization.

```suggestion
    precompressed: CompressedRistretto,
```

Then `From<RistrettoPoint>` would have to compress, but that's fine.

zkvm/src/signature/mod.rs
```diff
@@ -16,11 +16,59 @@ mod multikey;
...
+    }
+    /// Creates a new key from a ristretto point. Does not eagerly compress.
+    pub fn with_decompressed(p: RistrettoPoint) -> Self {
+        VerificationKey {
+            point: p,
+            compressed: Some(p.compress()),
+        }
+    }
+    /// Compresses the point if needed (no-op if `self.compressed == Some`)
+    pub fn compress_in_place(&mut self) {
```
oleganza: if we always store compressed point, we can remove this method (see above)

zkvm/src/signature/mod.rs
```diff
@@ -16,11 +16,59 @@ mod multikey;
...
+    }
+
+    /// Converts the Verification key to a point and returns a reference to it
+    pub fn as_point(&self) -> &RistrettoPoint {
+        &self.point
+    }
+
+    /// Converts Verification key to a compressed point.
+    /// If the compressed form is not available, the compression is completed
+    pub fn to_compressed_point(&self) -> CompressedRistretto {
```
oleganza: Instead of to/as compressed/decompressed point we can impl `Deref<>` and `Into<>` for each of these types.

zkvm/src/signature/mod.rs
decentralisedkev: With this, I think we can make most of the methods that fail with a compressedPoint non-failable; instead of passing the compressedPoint which could fail with `from_point`, we can pass the RistrettoPoint instead, since the compression is done regardless. Is this correct?

cathieyun:
We've changed a few things around in the signature crate, could you update this PR to get it back in sync? Thanks!

oleganza:
Hi, i've fixed some conflicts after merging the main in this branch: `oleg/232-fixup` (https://github.com/interstellar/slingshot/tree/oleg/232-fixup), if you update your PR with this, we could merge it. Thanks!

PR #237, opened 2019-03-21
zkvm/src/signature: add Signature::sign_single, handle 1-key Multikey specially

bobg:
Addresses part of https://github.com/interstellar/slingshot/issues/92.

### DIFF
--- a/zkvm/src/signature/multikey.rs
+++ b/zkvm/src/signature/multikey.rs
@@ -6,12 +6,23 @@ use merlin::Transcript;
 
 #[derive(Clone)]
 pub struct Multikey {
-    transcript: Transcript,
+    transcript: Option<Transcript>,
     aggregated_key: VerificationKey,
 }
 
 impl Multikey {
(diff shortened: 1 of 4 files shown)


zkvm/src/signature/multikey.rs
```diff
@@ -1,17 +1,29 @@
...
 #[derive(Clone)]
 pub struct Multikey {
-    transcript: Transcript,
+    transcript: Option<Transcript>,
     aggregated_key: VerificationKey,
 }
 
 impl Multikey {
-    pub fn new(pubkeys: Vec<VerificationKey>) -> Option<Self> {
+    pub fn new(pubkeys: Vec<VerificationKey>) -> Result<Self, VMError> {
```
oleganza: if we are going with Result<> i'd prefer not tying MuSig things into VMError/ZkVM too much. Maybe not in this PR, but it'd be cool to have MuSigError type, and keep all the things in one module. And then maybe move it out into a separate crate. I'll file an issue.

bobg: sgtm

PR #238, opened 2019-03-21
keytree: derive intermediate xprv

tessr:


keytree/src/lib.rs
```diff
@@ -167,6 +195,28 @@ mod tests {
...
+            t.commit_u64(b"account_id", 34);
+        });
+
+        assert_eq!(
+            hex::encode(xprv.scalar.as_bytes()),
+            "4a53c3fbbc59970ee5f85af813875dffc13a904a2e53ae7e65fa0dea6e62c901"
+        );
+        assert_eq!(
+            to_hex_32(xprv.dk),
+            "36e435eabc2a562ef228b82b399fbd004b2cc64103313fa673bd1fca0971f59d"
```
oleganza: This corresponds to the tests for xpub derivation, cool. Would be useful to also test that deriving `xprv->childxprv->childxpub` gives the same thing as `xprv->xpub->childxpub`.

tessr: Yes! Ok. I'll add another test at some point (maybe not in this PR so as not to mess up the approval ✅ )

oleganza:
Note: for `Xprv::derive_key` (for the leaf Scalar), we can use precomputed parent point to hash it, and avoid any overhead of computing the derived pubkey and its compression. This means, the signer can reuse the "account xprv" and perform any number of leaf derivations for free.

tessr:
Nice.

PR #241, opened 2019-03-21
zkvm: move `Program` into it's own module

### LINKED ISSUE #240: zkvm: move program builder API outside `ops` mod
Currently `ops` module contains definition of instructions and serialization code for them, but also the convenience API for building programs.

Suggestion: move the `Program` type and related APIs into `program.rs` module.

decentralisedkev:
Closes #240


tessr:
Hi, Kev! Thank you again for your contributions to Slingshot. 

You might've noticed that Oleg often changes your PR titles. He's just editing them into a shared format: 

`<crate name>: <description of what's happening>`

We have two other format guidelines, which are that the description is lowercase, and that it is in the [imperative mood](https://en.wikipedia.org/wiki/Imperative_mood). It might seem picky to be so stringent, but it makes our git history really nice to read! 

If this is a bit much, no worries--the Slingshot team can keep editing the titles. But I thought I'd let you know, in case you wanted to do it yourself.

decentralisedkev:
@tessr No worries, thanks for the information! I'll give it a whiz in the next PR

zkvm/src/program.rs
```diff
@@ -0,0 +1,176 @@
...
+        })
+    }
+
+    /// Converts the program to a plain vector of instructions.
+    pub fn to_vec(self) -> Vec<Instruction> {
+        self.0
+    }
+
+    /// Returns the serialized length of the program.
+    pub fn serialized_length(&self) -> usize {
```
vickiniu: Since we're exposing `program` as the public module, it seems like we do not need to expose `serialized_length` and `encode` as `pub` since a caller interested in building programs shouldn't need to call these functions!

vickiniu: same with `parse`?

PR #242, opened 2019-03-21
keytree: derive leaf keys

tessr:


keytree/src/lib.rs
```diff
@@ -79,6 +79,22 @@ impl Xprv {
         }
     }
 
+    /// Returns a leaf Xprv. (Is this really an xprv, or just a prv?)
```
tessr: One question here. Are these leaf keys really _extended_ keys, or just regular ole keys?

oleganza: Leaf secret is Scalar. Leaf pubkey is CompressedRistretto (we’ll change it later to VerificationKey)

tessr: Ohhh. Oops!

tessr: Ok. Does that mean I should just refer to this as "leaf secret" and "leaf pubkey"?

oleganza: Maybe “leaf private/public key”? If api is simply called `derive_key` then you can talk about private/public _keys_ and extended private/public keys.

keytree/src/lib.rs
```diff
@@ -79,6 +79,22 @@ impl Xprv {
...
+    /// Users must provide customize, in order to separate sibling keys from one another
+    /// through unique derivation paths.
+    pub fn derive_leaf_key(&self, customize: impl FnOnce(&mut Transcript)) -> Scalar {
```
oleganza: maybe:?
```suggestion
    pub fn derive_key(&self, customize: impl FnOnce(&mut Transcript)) -> Scalar {
```

keytree/src/lib.rs
```diff
@@ -139,6 +155,22 @@ impl Xpub {
         }
     }
 
+    /// Returns a leaf Xpub, which can safely be shared. (Is this really an xprv, or just a prv?)
+    /// Users must provide customize, in order to separate sibling keys from one another
+    /// through unique derivation paths.
+    pub fn derive_leaf_key(&self, customize: impl FnOnce(&mut Transcript)) -> CompressedRistretto {
```
oleganza: similarly,

```suggestion
    pub fn derive_key(&self, customize: impl FnOnce(&mut Transcript)) -> CompressedRistretto {
```

keytree/src/lib.rs
```diff
@@ -139,6 +155,22 @@ impl Xpub {
...
+        let mut t = Transcript::new(b"Keytree.derivation");
+        t.commit_bytes(b"pt", self.precompressed_pubkey.as_bytes());
+        t.commit_bytes(b"dk", &self.dk);
+
+        // change the derivation path for this key
+        customize(&mut t);
+
+        // squeeze a challenge scalar
+        let f = t.challenge_scalar(b"f.leaf");
+        (self.point + (f * &constants::RISTRETTO_BASEPOINT_POINT)).compress()
```
oleganza: maybe add a comment about future plans?
```suggestion
        // TBD: change this to VerificationKey after we factor out Musig (#239) 
        //      in order to preserve uncompressed point.
        (self.point + (f * &constants::RISTRETTO_BASEPOINT_POINT)).compress()
```

keytree/src/lib.rs
```diff
@@ -217,6 +249,20 @@ mod tests {
         );
     }
 
+    #[test]
+    fn random_xprv_leaf_test() {
+        let seed = [0u8; 32];
+        let mut rng = ChaChaRng::from_seed(seed);
+        let xprv = Xprv::random(&mut rng).derive_leaf_key(|t| {
+            t.commit_u64(b"account_id", 34);
```
oleganza: let's maybe do a more leaf-like example? Accounts are themselves key-trees, but individual invoices aren't:

```suggestion
            t.commit_u64(b"invoice_id", 1000024);
```

(because paperwork identifiers often long strings of zeroes in the middle ¯\\\_(ツ)_/¯ )

oleganza:
Nitpicks added, but essence-wise it's wholesome 👍

oleganza:
Are you planning to add a "rectangle test" in a separate PR? I mean, the test that `xprv->child->xpub` gives the same result as `xprv->xpub->child`?

keytree/src/lib.rs
tessr: oh, yes, good point!

tessr:
> essence-wise it's wholesome

this is the nicest thing anyone has said about my code maybe ever

> Are you planning to add a "rectangle test" in a separate PR?

oh yes, good idea!

PR #245, opened 2019-03-22
musig: use &mut Transcript comme il faut

oleganza:
Also cleans up warnings and missing docs.


cathieyun:
For the plebians among us (I had to look it up), "comme il faut" means "as necessary"

thanks for figuring this out, I'm glad it ended up not being as difficult as it seemed :)

oleganza:
It’s more like “as it supposed to be”

Issue #248, opened 2019-03-22
zkvm: redesign predicate tree as Taproot


debnil:
#### Overview
This issue redesigns the predicate tree as an implementation of Taproot. This transforms it into a normal Merkle tree of programs, rather than predicates, and a one-shot `call` instruction that checks a Merkle proof of program inclusion and then executes it.

#### Context: Taproot

From [Gregory Maxwell's Taproot proposal](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2018-January/015614.html):

>Interest in merkelized scriptPubKeys (e.g. MAST) is driven by two main
>areas: efficiency and privacy. Efficiency because unexecuted forks of
>a script can avoid ever hitting the chain, and privacy because hiding
>unexecuted code leaves scripts indistinguishable to the extent that
>their only differences are in the unexecuted parts.
>
>As Mark Friedenbach and others have pointed out before it is almost
>always the case that interesting scripts have a logical top level
>branch which allows satisfaction of the contract with nothing other
>than a signature by all parties.  Other branches would only be used
>where some participant is failing to cooperate. More strongly stated,
>I believe that _any_ contract with a fixed finite participant set
>upfront can be and should be represented as an OR between an N-of-N
>and whatever more complex contract you might want to represent.

The Taproot proposal uses the above insights and contains two features. First, it commits a signing key `X` and the alternative "clauses" `{C_i}` in a single signing key `P`, such that `P = X + h1(X, {C_i})*G`. Through this, it is possible to sign for P if parties wish to sign for X, thus not revealing alternative clauses. Second, the alternative clauses are organized as _programs_ compressed in a binary Merkle tree.

#### Current Approach
Our current approach diverges from this in two significant ways. First, rather than a Merkle tree of programs, we have a Merkle tree of Predicates (Ristretto points). Second, each node does not simply hash its children, but forms the same disjunction `L + hash(L, R)*G`. This was motivated by the elegant symmetry of its building blocks. Additionally, the nice Merkle Tree API now used for transactions did not exist at the time.

However, this design also poses two issues. First, to make proof verification batchable, both neighbor commitments at each level must be presented, thus wasting more than double the bandwidth of a single commitment. Second, if not redundantly presenting both neighbors, the computation of each commitment must happen in-place (as in any Merkle tree protocol). But this computation is expensive, because it requires the group operation `L+hash(L, R)*G`, rather than a simple hash in a standard Merkle tree.

### Solution
[...]

PR #250, opened 2019-03-22
zkvm/tests: use predicate chooser in tests

vickiniu:
Use `choose_predicate` API in tests (some small bug fixes also)

### DIFF
--- a/zkvm/tests/zkvm.rs
+++ b/zkvm/tests/zkvm.rs
@@ -527,10 +527,7 @@ fn predicate_disjunction_happy_path() {
     let prog = Program::build(|p| {
         p.push(Output::new(prev_output))
             .input()
-            .choose_predicate(disjunction, |t| {
-                t.select(0)?;
-                Ok(())
-            })
+            .choose_predicate(disjunction, |t| t.select(0).map(|_| ()))
             .unwrap()
             .sign_tx()
             .cloak_helper(1, vec![(qty, flavor)])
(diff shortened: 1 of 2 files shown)


zkvm/tests/zkvm.rs
```diff
@@ -527,9 +527,8 @@ fn predicate_disjunction_happy_path() {
     let prog = Program::build(|p| {
         p.push(Output::new(prev_output))
             .input()
-            .push(key_pred)
-            .push(program_pred)
-            .select(2, 0)
+            .choose_predicate(disjunction, |t| t.select(0).map(|_| ()))
```
vickiniu: I'm not sure if there's a better workaround here: effectively, we need `choose_fn` to return `Result<(), VMError>` since when we return the `Result<PredicateTree, VMError>` that comes from `select()`, there is an error in specifying lifetimes because returning the borrowed program in a new `PredicateTree` conflicts with the input lifetime

vickiniu: might be worth just changing the `choose_fn` signature to specify that and adding a function other than `call` that returns the `Result<(), VMError>`, unless there's a better solution!

oleganza:
Sorry for not making it clearer, but this issue will bring Taproot and effectively replace `select+call` with just `call`: https://github.com/interstellar/slingshot/issues/248 and make this change undone.

oleganza:
This is made obsolete by Taproot API merged in #268

PR #253, opened 2019-03-25
musig: move musig files into separate crate

cathieyun:
As described here: https://github.com/interstellar/slingshot/issues/92

- [x] Make own MuSigError
- [x] Make own TranscriptProtocol
- [x] Make own PointOp (instead of direct verification), for compatibility with ZkVM


musig/src/counterparty.rs
```diff
@@ -1,6 +1,5 @@
-use super::multikey::Multikey;
-use super::VerificationKey;
-use crate::errors::VMError;
+use super::errors::MuSigError;
```
oleganza: that's a bit JuMpiNg, let's call it `MusigError`? And use `Musig` instead of `MuSig` everywhere in the source code for that matter?

cathieyun: haha SounDs GoOd

cathieyun: done!

musig/README.md
```diff
@@ -42,10 +42,10 @@ Functions:
 
 ### Signature
 
-A signature is comprised of a scalar `s`, and a RistrettoPoint `R`. In the simple Schnorr signature case, `s` represents the Schnorr signature scalar and `R` represents the nonce commitment. In the MuSig signature case, `s` represents the sum of the Schnorr signature scalars of each party, or `s = sum_i (s_i)`. `R` represents the sum of the nonce commitments of each party, or `R = sum_i (R_i)`. 
+A signature is comprised of a scalar `s`, and a RistrettoPoint `R`. In the simple Schnorr signature case, `s` represents the Schnorr signature scalar and `R` represents the nonce commitment. In the Musig signature case, `s` represents the sum of the Schnorr signature scalars of each party, or `s = sum_i (s_i)`. `R` represents the sum of the nonce commitments of each party, or `R = sum_i (R_i)`.
```
bobg: Request: run this file through `mdfmt` (from i10r.io/bin).

musig/src/errors.rs
```diff
@@ -0,0 +1,22 @@
...
+        /// The pubkey corresponding to the share that failed fo verify correctly
+        pubkey: [u8; 32],
+    },
+
+    /// This error occurs when an individual point operation failed.
+    #[fail(display = "Point operation failed.")]
+    PointOperationFailed,
+
+    /// This error occurs when a function is called with bad arguments.
+    #[fail(display = "Bad arguments")]
```
bobg: Is there style guidance about these strings? Should they be capitalized? Some of these end in periods and some don't, should they be consistent?

cathieyun: I couldn't find an official style guide, but just looking through the examples, it seems like capitalize + no period at the end is the norm. I'll update that!
Example: https://doc.rust-lang.org/std/fmt/struct.Error.html

musig/README.md
cathieyun: I couldn't figure out how to run `mdfmt`, but I added line breaks manually, following https://sembr.org/

PR #255, opened 2019-03-26
zkvm/src: make MerkleTree::build always produce a MerkleTree (never None)

bobg:


zkvm/src/merkle.rs
```diff
@@ -32,16 +32,32 @@ enum MerkleNode {
...
+            root: root,
+        }
+    }
+
+    /// Returns the root hash of the Merkle tree.
+    pub fn hash(&self) -> [u8; 32] {
+        match &self.root {
+            None => {
+                let t = Transcript::new(self.label);
+                let mut result = [0u8; 32]; // xxx why must i preinitialize this?
```
oleganza: Because you cannot pass a readable reference to uninitialized data.

zkvm/src/merkle.rs
```diff
@@ -32,16 +32,32 @@ enum MerkleNode {
...
+                let t = Transcript::new(label);
+                Some(Self::build_tree(t, list))
+            }
+        };
+        MerkleTree {
             size: list.len(),
             label,
-            root: Self::build_tree(t, list),
-        })
+            root: root,
```
oleganza: Can be written simply
```suggestion
            root,
```

zkvm/src/merkle.rs
```diff
@@ -22,7 +22,7 @@ pub enum MerkleNeighbor {
 pub struct MerkleTree {
     size: usize,
     label: &'static [u8],
-    root: MerkleNode,
+    root: Option<MerkleNode>,
```
oleganza: Maybe use MerkleNode::Leaf([u8; 32]) and avoid optional here? Or add ::Empty variant if that’d be cleaner.

zkvm/src/merkle.rs
bobg: Is there a Rust idiom that lets me pass a container to a function to be initialized? Like passing a write-iterator or something?

oleganza: There is &mut Vec and there is IntoIterator trait. Or do you specifically mean “i want to avoid memset”? Then you can use unsafe raw memory api in core::mem and core::ptr.

zkvm/src/merkle.rs
```diff
@@ -149,27 +134,35 @@ impl MerkleTree {
...
     }
 
     /// Returns the hash of a Merkle tree.
-    fn hash(&self) -> &[u8; 32] {
+    fn hash(&self) -> [u8; 32] {
         match self {
-            MerkleNode::Leaf(h) => h,
-            MerkleNode::Node(h, _, _) => h,
+            MerkleNode::Empty(h) => h.clone(),
+            MerkleNode::Leaf(h) => h.clone(),
```
oleganza: Is there a need to clone these hashes?

bobg: Adjusted.

PR #260, opened 2019-03-27
Musig: multimessage signature protocol design docs

cathieyun:
First pass on writing up the "joint" protocol that allows for multimessage signing (safe from russell's attack) and multikey (single message) signing. 

@oleganza I'm just looking for a review on "do I have the right idea" / "am I missing anything glaringly conceptual"? - sorry for the messiness, I can clean up typos and small mistakes when I get back!


musig/README.md
```diff
@@ -34,19 +34,63 @@ Ristretto base point in compressed form:
 B = e2f2ae0a6abc4e71a884a961c500515f58e30b6aa582dd8db6a65945e08d2d76
 ```
 
-### Multikey
+### MusigContext
+
+This is a trait with two functions:
+- `commit(&self, &mut transcript)`: takes a mutable transcript, and commits the internal context to the transcript.
+- `challenge(&self, &verification_key, &mut transcript, &nonce_commitment) -> Scalar`: takes a public key and mutable transcript, and returns the
```
oleganza: The addition of the `sum{R_i}` is the same for both contexts, so we can move this outside this context and perform inside the Party...::receive_commitments(), while the contexts are only doing context-specific things.

```suggestion
- `challenge(&self, &verification_key, &mut transcript) -> Scalar`: takes a public key and mutable transcript, and returns the 
```

musig/README.md
```diff
@@ -34,19 +34,63 @@ Ristretto base point in compressed form:
...
+- `Multikey::challenge(&self, &verification_key, &mut transcript, &nonce_commitment) -> Scalar`: 
+  Computes challenge `c_i = a_i * c`, where `a_i = H_agg(<L>, X_i)` and `c = H_sig(X, R, m)`.
+
+  For calculating `a_i`, `<L>` (the list of pubkeys that go into the aggregated pubkey)
+  has already been committed into `self.transcript`. Therefore this function simply clones `self.transcript`, 
+  commits the verification key (`X_i`) into the transcript with label "X_i", 
+  and then squeezes the challenge scalar `a_i` from the transcript with label "a_i".
+
+  For calculating `c`, the message `m` has already been committed to the input `transcript`, 
+  so the function only needs to commit `self.aggregated_key` with label "X" and input `hash_commitment` with label "R",
```
oleganza: AFAIK, the `MusigContext::commit()` has already committed X, and then (inside Party::receive_commitments) R was committed too, so we only need to squeeze the challenge in this method.

Also: the MusigContext should probably be a private trait, so the usage of these commit/challenge helpers is hardcoded correctly under the public API, so no one is supposed to compose usage of these in their protocols. This is strictly the internal organization of the Musig implementation.

musig/README.md
```diff
@@ -34,19 +34,63 @@ Ristretto base point in compressed form:
...
+
+### Multimessage (implements MusigContext)
+
+Fields:
+- pairs: `Vec<(VerificationKey, [u8])`
+
+Functions:
+- `Multimessage::new(Vec<(VerificationKey, [u8])) -> Self`: creates a new MultiMessage instance using the inputs.
+
+- `Multimessage::commit(&self, &mut transcript)`: It commits each of the pairs in `self.pairs` to the input `transcript`,
```
oleganza: mention the length prefix too?

`transcript.commit_u64(pairs.len())`

musig/README.md
```diff
@@ -34,19 +34,63 @@ Ristretto base point in compressed form:
...
+  by iterating through `self.pairs` and committing the `VerificationKey` with label "X" and the message with label "m".
+
+- `Multimessage::challenge(&self, &verification_key, &mut transcript, &nonce_commitment) -> Scalar`: 
+  Computes challenge `c_i = H(R, <S>, i)`.
+  It first commits the input `nonce_commitment` to the input `transcript` with label "R".
```
oleganza: Per notes above, the R is assumed to be committed by the caller (Party::receive_commitments), and pairs are committed by the caller too via Multimessage::commit().

Here we only need to figure out the `i`, fork the transcript, commit `i`, get challenge, and then in the non-forked transcript add domain separation with `commit("dom-sep", "Musig.multi-message-boundary")`.

Does this make sense?

musig/README.md
```diff
@@ -123,24 +167,52 @@ There are several paths to signing:
...
+    Function: `Signature::verify_multimessage(...)`
+
+    Input: 
+    - `&self`
+    - transcript: `&mut Transcript` - a transcript to which the signed message has already been committed.
+    - multimessage: `Multimessage` 
+
+    Operation:
+    - Iterate through each (pubkey, message) pair in `multimessage`. For each, create `c_i` by calling
+      `multimessage.challenge(...)` and passing in the verification key, transcript, and `self.R`.
```
oleganza: per notes above, how about this:

1. use `multimessage.commit(&mut t)` to commit the key
2. commit the R directly w/o using multimessage: `t.commit(R)`
3. use `multimessage.challenge(pk, &mut t)` to squeeze the per-point challenge.
4. Sum up `sum_i(X_i * c_i)` into `cX` and then check `s*G == cX + R`.

musig/README.md
cathieyun: sounds good!

musig/README.md
cathieyun: it doesn't seem to be a huge difference whether we commit `R` and the context in the `MusigContext::challenge` method (as I detailed in the first draft of the writeup), or outside it in the `PartyAwaitingCommitments::receive_commitments` method (as you are suggesting). The only thing I would be cautious of with your proposal, is making sure that the transcript has actually had `R` and the context committed to it before squeezing the challenge! (Whereas with the first draft, you don't have to worry about it because the method commits those for you). 

I'll change the writeup to your suggestion!

musig/README.md
cathieyun: added a note that musigcontext should be a private trait.

Also, nowhere in the party transitions was `MusigContext::commit()` being called in the design... only in `MusigContext::challenge()`. So, as I detailed here (https://github.com/interstellar/slingshot/pull/260#discussion_r271080929), the current design and your proposed design are both valid, but different in where the commitment is happening.

musig/README.md
cathieyun: done, thanks for making that clearer!

musig/README.md
```diff
@@ -98,13 +153,13 @@ There are several paths to signing:
...
 
 2. Make a Schnorr signature with one aggregated key (`Multikey`), derived from multiple public keys.
-    - Create an aggregated pubkey. For more information see the [key aggregation](#key-aggregation) section.
+    - Create a `Multikey`. For more information, see the [key aggregation](#key-aggregation) section.
 
     For each party that is taking part in the signing:
-    - Call `Party::new(transcript, privkey, multikey, pubkeys)`.
+    - Call `Party::new(transcript, privkey, multikey)`.
     - Get back `PartyAwaitingPrecommitments` and a `NoncePrecommitment`.
     - Share your `NoncePrecommitment`, and receive other parties' `NoncePrecommitment`s.
```
bobg: These blank lines don't show up in the rendered version of the document. Do you intend to include section dividers here? If so, you need to make them visible; if not, please remove these blank lines.

musig/README.md
```diff
@@ -123,24 +178,53 @@ There are several paths to signing:
...
+    Function: `Signature::verify_multimessage(...)`
+
+    Input: 
+    - `&self`
+    - transcript: `&mut Transcript` - a transcript to which the signed message has already been committed.
+    - multimessage: `Multimessage` 
+
+    Operation:
+    - Use `multimessage.commit(&mut transcript)` to commit the key
+    - Commit the `self.R` to the transcript with label "R".
```
bobg:
```suggestion
    - Commit `self.R` to the transcript with label "R".
```

musig/README.md
```diff
@@ -355,19 +440,20 @@ Fields:
...
+- context: `C`
+- transcript: `&mut transcript`
 
 Operation:
-- Verify that `s_i * G == R_i + c * a_i * X_i`.
-  `s_i` = share, `G` = [base point](#base-point), `R_i` = self.commitment, `c` = challenge, 
-  `a_i` = `multikey.factor_for_key(self.pubkey)`, `X_i` = self.pubkey.
+- Verify that `s_i * G == R_i + c_i * X_i`.
+  `s_i` = share, `G` = [base point](#base-point), `R_i` = self.commitment,
+  `c_i` = `multikey.challenge(self.pubkey, &mut transcript, nonce_sum)`, `X_i` = self.pubkey.
```
oleganza: do you mean context?

```suggestion
  `c_i` = `context.challenge(self.pubkey, &mut transcript, nonce_sum)`, `X_i` = self.pubkey.
```

musig/README.md
cathieyun: I did intend to have section dividers; i'll make this more explicit.

musig/README.md
cathieyun: yes, thanks for the catch!

musig/README.md
```diff
@@ -34,19 +34,73 @@ Ristretto base point in compressed form:
...
+
+  For calculating `c`: the message `m`, the nonce commitment sum `R`, and the aggregated key `X` 
+  have already been committed to the input `transcript`.
+  It then gets the challenge scalar `c` from the transcript with label "c".
+
+  Returns `c_i = a_i * c`.
+
 - `Multikey::aggregated_key(&self)`: returns the aggregated key stored in the multikey, `self.aggregated_key`.  
 
+- `Multikey::get_pubkeys(&self)`: returns the list of public keys, `self.public_keys`.
```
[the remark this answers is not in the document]
cathieyun: done!

musig/README.md
```diff
@@ -34,19 +34,73 @@ Ristretto base point in compressed form:
...
+  commits the verification key (`X_i`) into the transcript with label "X_i", 
+  and then squeezes the challenge scalar `a_i` from the transcript with label "a_i".
+
+  For calculating `c`: the message `m`, the nonce commitment sum `R`, and the aggregated key `X` 
+  have already been committed to the input `transcript`.
+  It then gets the challenge scalar `c` from the transcript with label "c".
+
+  Returns `c_i = a_i * c`.
+
 - `Multikey::aggregated_key(&self)`: returns the aggregated key stored in the multikey, `self.aggregated_key`.
```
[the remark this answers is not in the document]
cathieyun: done!

oleganza:
why the approval is still on, after newly added commits?

oleganza:
Ah, it’s not a merge into main.

cathieyun:
Yup, it was originally based off of `cathie/musig` (since it wasn't merged in yet). I just rebased it off of `main`, so now can you give it another 👍 ?

PR #261, opened 2019-03-28
zkvm/signature: add DeferredVerification type

### LINKED ISSUE #243: signature: deferred verification object
As we are going to move musig out of zkvm (#239), we couldn't use `PointOp`, and it's a poor match anyway: all we need is a subset of this struct that only has one base point:

```rust
/// Deferred signature verification.
#[derive(Clone, Debug)]
pub struct DeferredVerification {
    /// Weight for the Ristretto base point.
    pub static_point_weight: Scalar,

    /// Weights for arbitrary points.
    pub dynamic_point_weights: Vec<(Scalar, CompressedRistretto)>,
}
```

We can borrow the single and batchable verification code from zkvm and make a simple converter of this into PointOp to batch with other ZkVM instructions.

- [ ] Return `DeferredVerification` from `Signature::verify_deferred` method.
- [ ] Change `Signature::verify` method to call `verify_deferred().verify()`.
- [ ] Add `.verify` and `.verify_batch` akin to zkvm's impl of PointOp.
- [ ] In ZkVM add `impl From<DeferredVerification> for PointOp` that moves the weights to zkvm's structure.

vickiniu:
Changes signature verification to not be based on `PointOp`, since musig is moving out of zkvm. 

Fixes #243


vickiniu:
updated to use old version of single-party Schnorr key aggregation in tests (& export as a helper), with closures for multi-party MuSig when that's ready!

cathieyun:
LGTM! 

By "closures for multi-party MuSig", do you mean multi-message MuSig? (since the current setup is multi-party by nature of using a MuSig aggregated key?)

vickiniu:
ah, I did definitely mean multi-message MuSig! too much multi for me

PR #264, opened 2019-04-01
zkvm/src/blockchain: implement some of blockchain spec

bobg:
This implements some of the spec at https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md. It is purposely incomplete. We're merging it to `main` to support parallel development of aspects of the spec.

Data types included:
- [BlockHeader](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#block-header)
- [Block](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#block)
- [BlockchainState](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#blockchain-state)
- [BlockID](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#block-id)

Procedures:
- [BlockchainState::make_initial](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#start-new-network)
- [BlockchainState::apply_block](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#apply-block)
- [BlockchainState::apply_txlog](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#apply-transaction-log)
- [BlockHeader::make_initial](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#make-initial-block-header)
- [Block::validate](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#validate-block)


tessr:
If this build passes, can you update the PR description to describe which parts of the spec you've implemented (e.g. which types and which processes)?

bobg:
Depends on https://github.com/interstellar/slingshot/pull/255.

zkvm/src/blockchain/block.rs
```diff
@@ -0,0 +1,121 @@
...
+pub struct BlockHeader {
+    pub version: u64,
+    pub height: u64,
+    pub prev: BlockID,
+    pub timestamp_ms: u64,
+    pub txroot: [u8; 32],
+    pub utxoroot: [u8; 32],
+    pub nonceroot: [u8; 32],
+    pub refscount: u64,
+    pub ext: Box<[u8]>,
```
oleganza: Is there a reason for using a boxed slice? `Vec<u8>` would be more idiomatic.

zkvm/src/blockchain/block.rs
```diff
@@ -0,0 +1,121 @@
...
+        if self.header.version == 1 && self.header.ext.len() != 0 {
+            return Err(BCError::IllegalExtension);
+        }
+        if self.header.height != prev.height + 1 {
+            return Err(BCError::BadHeight);
+        }
+        if self.header.prev != prev.id() {
+            return Err(BCError::MismatchedPrev);
+        }
+        if self.header.timestamp_ms <= prev.timestamp_ms {
```
oleganza: There is a lot of conditions that are inverted and hard to read. How about having a helper method `check` (or `require`, or `ensure` or w/e) that uses the desired condition directly:

```rust
check(self.header.version >= prev.version, BCError::VersionReversion)?;
check(self.header.height == prev.height + 1, BCError::BadHeight)?;
check(self.header.prev == prev.id(), BCError::MismatchedPrev)?;
check(self.header.timestamp_ms > prev.timestamp_ms, BCError::BadBlockTimestamp)?;
...
```
 
Also: maybe it makes sense to factor out all BlockHeader rules into `BlockHeader::validate`, so headers can be validated w/o transactions, and we don’t have to repeat “self.header, self.header, self.header” many times?

zkvm/src/blockchain/block.rs
```diff
@@ -0,0 +1,121 @@
...
+            return Err(BCError::BadBlockTimestamp);
+        }
+        if self.header.refscount > prev.refscount + 1 {
+            return Err(BCError::BadRefscount);
+        }
+
+        let mut txlogs: Vec<TxLog> = Vec::new();
```
oleganza:
```suggestion
        let mut txlogs: Vec<TxLog> = Vec::with_capacity(self.txs.len());
```

and same for txids.

zkvm/src/blockchain/block.rs
```diff
@@ -0,0 +1,121 @@
...
+            if tx.header.mintime_ms > self.header.timestamp_ms
+                || self.header.timestamp_ms > tx.header.maxtime_ms
+            {
+                return Err(BCError::BadTxTimestamp);
+            }
+            if self.header.version == 1 && tx.header.version != 1 {
+                return Err(BCError::BadTxVersion);
+            }
+
+            // TODO(bobg): The API currently requires anticipating how many multipliers (generators) are needed,
```
oleganza: Based on conversation in the Dalek slack with Henry and Vicki, the best path forward is:

1. On the prover, measure how many mults you need from a proof, and insert that into Tx object. We can also have a “measurement” impl of ConstraintSystem trait that collects that w/o fully computing the r1cs proof, but that’s more like an optimization.
2. Verifier would have the required capacity in the Tx object immediately, so it can ensure necessary capacity of the generators or fail early if it’s beyond whatever limits.

zkvm/src/blockchain/block.rs
```diff
@@ -0,0 +1,121 @@
...
+            // _before_ knowing what's in the transaction.
+            // The value is 64 for each range proof,
+            // and there are at least as many range proofs as outputs in the tx.
+            // Guessing a value should _not_ be part of the tx-verifying API.
+            // Related: https://github.com/dalek-cryptography/bulletproofs/pull/263
+            let bp_gens = BulletproofGens::new(64 * 64, 1);
```
oleganza: Generators are expensive to compute and must be passed as an argument to whatever verification logic that wants to use them (as `&BulletproofGens`)

zkvm/src/blockchain/errors.rs
```diff
@@ -0,0 +1,34 @@
+use crate::errors::VMError;
+
+#[derive(Debug, Fail)]
+pub enum BCError {
```
oleganza: maybe this longer name is not bad, so we avoid clunky abbreviations? That’s my personal preference though.
```suggestion
pub enum BlockchainError {
```

zkvm/src/blockchain/state.rs
```diff
@@ -0,0 +1,78 @@
+use super::block::{Block, BlockHeader, BlockID};
+use super::errors::BCError;
+use crate::{Entry, TxLog, VMError, UTXO};
+use std::collections::{HashSet, VecDeque};
+
+#[derive(Clone)]
+pub struct BCState {
```
oleganza: I wish there was a nice single-word term. Meanwhile, `BlockchainState` looks milder to my eye than “B-C-State”.

zkvm/src/blockchain/block.rs
bobg: Much nicer, thanks!

zkvm/src/blockchain/block.rs
bobg: I'm not 100% sure but it sounds like this is not actionable feedback about this PR, so I'm resolving it.

zkvm/src/blockchain/block.rs
bobg: Same thing - sounds like this is future work. Resolving.

oleganza: this one is actual suggestion for PR.

zkvm/src/blockchain/block.rs
oleganza: Yeah, sorry, this is not actionable. Need to file an issue for this.

zkvm/src/blockchain/block.rs
oleganza: (see other APIs that take &BulletproofGens as argument - r1cs::Verifier, zkvm::Verifier)

bobg: Oh, are you suggesting that the call signature for Block::validate should include a `&BulletproofGens`? That wasn't clear.

oleganza: yeah. Sorry!

zkvm/src/blockchain/state.rs
```diff
@@ -0,0 +1,89 @@
...
+use super::block::{Block, BlockHeader, BlockID};
+use super::errors::BlockchainError;
+use crate::{Entry, TxLog, VMError, UTXO};
+
+#[derive(Clone)]
+pub struct BlockchainState {
+    pub initial: BlockHeader,
+    pub tip: BlockHeader,
+    pub utxos: HashSet<UTXO>,
+    pub nonces: VecDeque<([u8; 32], u64)>, // xxx need fast lookup by anchor
```
vickiniu: Maybe a [Map](https://doc.rust-lang.org/std/iter/struct.Map.html)?

zkvm/src/blockchain/state.rs
```diff
@@ -0,0 +1,89 @@
...
+        key_agg_fn: F,
+    ) -> Result<BlockchainState, BlockchainError>
+    where
+        F: Clone + Fn(&[VerificationKey]) -> Result<VerificationKey, VMError>,
+    {
+        let txlogs = b.validate(&self.tip, bp_gens, key_agg_fn)?;
+        let mut new_state = self.clone();
+
+        // Remove expired nonces.
+        while let Some(nonce_pair) = new_state.nonces.front() {
```
vickiniu: It seems like this will remove nonces from the set until we encounter one with timestamp >= blockheader timestamp. Is that the desired behavior? Seems like we would want to iterate through all nonces and remove ones with expired timestamp, but I might be misunderstanding

zkvm/src/blockchain/state.rs
bobg: Sure, but I think that's for a followup PR. The code that needs to do the lookup doesn't exist yet, and remember, we're just trying to get this landed so everyone else can start filling in parts of it.

zkvm/src/blockchain/state.rs
bobg: Great catch! This logic presumes that `new_state.nonces` is sorted by expiration time. Rather than fix anything, I added a comment for future implementers about that being a requirement.

zkvm/src/blockchain/state.rs
vickiniu: Makes sense to me!

vickiniu:
this lgtm, but looks like there are some test failures from expecting a `&Tx` in `verify_tx` ?

PR #267, opened 2019-04-04
zkvm/key: use MuSig key aggregation scheme

vickiniu:
Removes the `key_agg_fn` closure in `Verifier::verify_tx` to make key aggregation consistent between Multikey and single-party Schnorr aggregation. 

Sparked by [this conversation](https://i10r.slack.com/archives/CEMSC5YRK/p1554331172056000) since we want all transactions to be verifiable without this extra per-tx information about which key aggregation scheme to use.

### DIFF
--- a/zkvm/src/verifier.rs
+++ b/zkvm/src/verifier.rs
@@ -99,7 +99,8 @@ impl<'t> Verifier<'t> {
 
         if verifier.signtx_keys.len() != 0 {
             verifier.deferred_operations.push(
-                // TODO: change this to use multi-message context
+                /// TODO: use MuSig multi-message API, signing contract
+                /// IDs in addition to TxID for each key.
                 tx.signature
                     .verify(
                         &mut signtx_transcript,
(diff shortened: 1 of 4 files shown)


zkvm/src/verifier.rs
```diff
@@ -108,7 +101,12 @@ impl<'t> Verifier<'t> {
             verifier.deferred_operations.push(
                 // TODO: change this to use multi-message context
                 tx.signature
-                    .verify(&mut signtx_transcript, key_agg_fn(&verifier.signtx_keys)?)
+                    .verify(
+                        &mut signtx_transcript,
+                        Multikey::new(verifier.signtx_keys)
```
oleganza: Maybe add a comment that we'd need to sign the contract IDs per key (in addition to txid) via multi-message API?

```suggestion
                        // TODO: change this to a proper multi-message API (#176, #266) and use (Contract ID, pubkey) pairs.
                        Multikey::new(verifier.signtx_keys)
```

PR #268, opened 2019-04-04
zkvm: implement Taproot

### LINKED ISSUE #248: zkvm: redesign predicate tree as Taproot

debnil:
Implements the Taproot proposal to store smart contracts in place of the predicate tree. Closes #248 .


zkvm/src/merkle.rs
```diff
@@ -60,14 +60,12 @@ impl MerkleTree {
         Ok(result)
     }
 
-    /// Verifies the Merkle path for an item givne the path and the Merkle root.
-    pub fn verify_path<M: MerkleItem>(
-        label: &'static [u8],
+    /// Computes the Merkle root, given the Merkle path.
+    pub fn compute_root_from_path<M: MerkleItem>(
```
oleganza: This probably should not be a public API. For one reason: it takes the precomputed transcript object that's guaranteed to be in a consistent state only by this module (and it's done as an optimization). The user is not supposed to instantiate or clone one themselves.

```suggestion
    fn compute_root_from_path<M: MerkleItem>(
```

oleganza: Ok, i see why you need this public API. 

Suggestion: use `label: &'static [u8],` instead of a raw transcript like it's done in `verify_path`. Because the use of term "transcript" here is an implementation detail — we just use the sponge API as our hashing primitive, and the user is not even supposed to know that.

zkvm/src/predicate.rs
```diff
@@ -1,7 +1,7 @@
 //! Implementation of a predicate tree.
 //! Inspired by Taproot by Greg Maxwell and G'root by Anthony Towns.
 //! Operations:
-//! - disjunction: P = L + f(L,R)*B
+//! - taproot key: P = X + h(X, M)*B
```
oleganza:
```
//! Implementation of a predicate tree.
//! Inspired by Taproot by Greg Maxwell and G'root by Anthony Towns.
```
->
```
//! ZkVM-specific implementation of a Taproot design proposed by Greg Maxwell.
//! https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2018-January/015614.html
```

zkvm/src/predicate.rs
```diff
@@ -27,21 +29,38 @@ pub enum Predicate {
...
     precomputed_point: CompressedRistretto,
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct CallProof {
+    // Signing key
+    pub signing_key: VerificationKey,
+
+    // Bit pattern indicating neighbor position
+    pub positions: u32,
```
oleganza: `MerkleNeighbor` already encodes the position via Left/Right variants. This 32-bit pattern can be kept at the serialization level only. In this struct you can keep the neighbors and verification key only.

Later, to avoid reallocations, you can change the proof-verifying code to take iterator of `MerkleNeighbor`s, so we don't have to allocate a Vec of them, but just parse them on the fly from the serialized callproof.

zkvm/src/predicate.rs
```diff
@@ -27,21 +29,38 @@ pub enum Predicate {
...
+    blinding: [u8; 32],
+    // We precompute the tweaked aggregated signing key when composing it via `Predicate::tree()`,
     // so that we can keep `to_point`/`encode` methods non-failable across all types.
     precomputed_point: CompressedRistretto,
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct CallProof {
+    // Signing key
+    pub signing_key: VerificationKey,
```
oleganza: Maybe use "signing key" as a synonym for "private key" (because you can _sign_ with it)?
```suggestion
    pub verification_key: VerificationKey,
```

zkvm/src/predicate.rs
```diff
@@ -27,21 +29,38 @@ pub enum Predicate {
...
+
+    // Bit pattern indicating neighbor position
+    pub positions: u32,
+
+    // List of left-right neighbors, excluding the root and leaf hash
+    pub neighbors: Vec<MerkleNeighbor>,
+}
+
+// PRTODO: Rewrite MerkleItem for Program using BlindedProgram.
+pub enum BlindedProgram {
```
oleganza: The more apt name for this is `TaprootLeaf` because it's an enumeration of two kinds of leaves: `Program` and `BlindingFactor`.

oleganza: Or `PredicateLeaf`.

zkvm/src/predicate.rs
```diff
@@ -136,43 +102,61 @@ impl Predicate {
...
 
-    /// Creates a disjunction of a vector of predicates.
-    pub fn disjunction(preds: Vec<Predicate>) -> Result<Self, VMError> {
+    /// Creates a Tree from a vector of programs.
+    pub fn tree(key: Option<VerificationKey>, progs: Vec<Program>, blinding_key: [u8; 32]) -> Result<Self, VMError> {
+        let signing_key = match key {
+            Some(x) => x,
+            _ => VerificationKey(PedersenGens::default().B_blinding.compress()),
+        };
+        // PRTODO: Do I assume that progs already contains the per-program blinding factors?
```
oleganza: Not quite. The idea is to use some tree leaves as programs, and others - as blinding factors for them. Meaning, each leaf-program is replaced by a node with two leaves: a program and a dummy leaf acting as additional entropy to blind the program.

When constructing a tree, a single `blinding_key` can be used to generate blinding factors for each program. This can be done by creating a new transcript instance with label, say, `"ZkVM.taproot-blinding"`, commit N programs, and squeeze N blinding factors. Then, the first bit of the blinding factor can be used to determine the relative position of the program and a blinding factor: 0 for blinding-on-the-left, 1 for blinding-on-the-right.

```
N=4
progs = [p1, p2, p3, p4]
T = Transcript::new("ZkVM.taproot-blinding")
T.commit("n", 4)
T.commit("prog", p1)
T.commit("prog", p2)
T.commit("prog", p3)
T.commit("prog", p4)

f1 = T.challenge("blinding") // 1...
f2 = T.challenge("blinding") // 0...
f3 = T.challenge("blinding") // 0...
f4= T.challenge("blinding") // 1...

leaves = [p1, f1, f2, p2, f3, p3, p4, f4]
root = MerkleTree::root(leaves)
```

oleganza: The `zkvm::Verifier` will not know the difference between a neighbour being a hash of a blinding factor, and a neighbour being a hash of program. Also, the relative order is uniformly random, so even the fact of using this particular scheme remains hidden.

oleganza: The last piece: in multi-party contracts, Alice wants to be sure that Bob cannot use their blinding factor as a valid program (and vice versa). So we'd use a different leaf hashing function to make the resulting merkle node non-openable into a program.

Currently, `impl MerkleItem for Program` uses label `"program"` for committing a program. For blinding factor, we need a domain separator, like `"blinding"`. Then, Bob can show the blinding factor to Alice and she'll compute the node with a different label, and will be convinced that it's infeasible to find a program that (if committed with a proper `"program"` label) would open to the same node hash.

zkvm/src/predicate.rs
```diff
@@ -181,75 +165,46 @@ impl Into<CompressedRistretto> for Predicate {
...
+        SliceReader::parse(data, |r| {
+            let mut call_proof = CallProof::default();
+            call_proof.signing_key = VerificationKey(r.read_point()?);
+            call_proof.positions = r.read_u32()?;
+            let mut neighbors = vec![];
+            while r.len() > 0 {
+                neighbors.push(r.read_u8x32()?);
+            }
+            for i in 0..neighbors.len() {
+                if call_proof.positions & (1 << 31 - i) == 0 {
```
oleganza: would be great to specify the endianness and semantics of the `callproof.positions`. This code seems to use the highest bit for the lowest node. Maybe reverse this? Then, the highest unused bits can be zeroes, so the overall number is ≈proportional to the proof size?

zkvm/src/vm.rs
```diff
@@ -713,44 +713,31 @@ where
...
+        let call_proof = self.pop_item()?.to_data()?.to_call_proof()?;
         let contract = self.pop_item()?.to_contract()?;
         let predicate = contract.predicate;
 
-        // 0 = -P + h(prog) * B2
+        let signing_key = call_proof.signing_key;
+        let neighbors = call_proof.neighbors;
+
+        // Compute Merkle root M.
+        let transcript = Transcript::new(b"ZkVM.taproot");
```
oleganza: in addition to the above suggestion, let's also move the root computation into the `predicate.prove_taproot`? So this code does not have to assemble the proving, but can use a single function to prove the relation between `program`, `call_proof` and `predicate`.

zkvm/src/predicate.rs
```diff
@@ -27,19 +28,39 @@ pub enum Predicate {
...
     // so that we can keep `to_point`/`encode` methods non-failable across all types.
     precomputed_point: CompressedRistretto,
+
+    // Scalar that's added to the signing key - a hash of the merkle root and the pubkey
+    adjustment_factor: Scalar,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct CallProof {
+    // Tweaked verification key
```
oleganza: Why is this tweaked? In the context of `call`, this key is just a piece of extra data necessary to verify a commitment. And it's supposed to be the pure verification key X in the formula `P = X + hash(X,Root)*B`, not the predicate point itself (P).

zkvm/src/predicate.rs
```diff
@@ -27,19 +28,39 @@ pub enum Predicate {
...
+    blinding_key: [u8; 32],
+    // We precompute the tweaked aggregated signing key when composing it via `Predicate::tree()`,
     // so that we can keep `to_point`/`encode` methods non-failable across all types.
     precomputed_point: CompressedRistretto,
+
+    // Scalar that's added to the signing key - a hash of the merkle root and the pubkey
+    adjustment_factor: Scalar,
+}
+
+#[derive(Clone, Debug, Default)]
```
oleganza: Documentation:
```suggestion
/// Call proof represents a proof that a certain program is committed via the merkle tree into the predicate.
/// Used by `call` instruction. The program is not the part of the proof.
#[derive(Clone, Debug, Default)]
```

zkvm/src/predicate.rs
```diff
@@ -27,19 +28,39 @@ pub enum Predicate {
...
+pub struct CallProof {
+    // Tweaked verification key
+    pub verification_key: VerificationKey,
+
+    // List of left-right neighbors, excluding the root and leaf hash
+    pub neighbors: Vec<MerkleNeighbor>,
+}
+
+#[derive(Clone, Debug)]
```
oleganza:
```suggestion
/// PredicateLeaf represents a leaf in the merkle tree of predicate's clauses.
/// For secrecy, each program is blinded via a dummy neighbour called the "blinding leaf".
/// From the verifier's perspective, the hash of this node simply appears as part of a merkle proof,
/// but from the prover's perspective, some leafs are dummy uniformly random nodes.
#[derive(Clone, Debug)]
```

zkvm/src/predicate.rs
```diff
@@ -68,66 +83,20 @@ impl Predicate {
...
-    /// Creates a program with an empty blinding factor.
-    pub fn unblinded_program<T: Into<Program>>(x: T) -> Self {
-        Predicate::Program(x.into(), Vec::new())
-    }
-
     /// Downcasts the predicate to a verification key
     pub fn to_key(self) -> Result<VerificationKey, VMError> {
         match self {
             Predicate::Key(k) => Ok(k),
+            Predicate::Tree(t) => Ok(VerificationKey(t.precomputed_point)),
```
oleganza: Oops. We actually need two separate semantics (can fix this in the follow-up PR):

1. On the prover's end, we need to convert ::Key/::Tree into "key witness thing". Say, `to_key_witness()->Result<Multikey>` (`Multikey` from MuSig). And fail if it's not unsignable, or if the predicate is `::Opaque`.
2. On the verifier's end, we need a separate method `to_verification_key()` that converts all variants into a VerificationKey and does not fail.

oleganza: Filed here: https://github.com/interstellar/slingshot/issues/272

zkvm/src/predicate.rs
```diff
@@ -136,42 +105,43 @@ impl Predicate {
...
+        let mut t = Transcript::new(b"ZkVM.taproot");
+        t.commit_bytes(b"key", key);
+        t.commit_bytes(b"merkle", root);
+        t.challenge_scalar(b"h")
+    }
+
+    /// Verifies whether the current predicate is a commitment to a signing key `key` and Merkle root `root`.
+    /// Returns a `PointOp` instance that can be verified in a batch with other operations.
+    pub fn prove_taproot(&self, program: &Program, call_proof: &CallProof) -> PointOp {
+        let pred_leaf = PredicateLeaf::Program(program.clone());
```
oleganza: That's unfortunate clone() here. Can we impl MerkleItem for `&Program` and pass `&program` as a leaf to `compute_root_from_path` API? (We cannot open a blinding factor to a program anyway.)

zkvm/src/predicate.rs
```diff
@@ -136,42 +105,43 @@ impl Predicate {
...
-        t.challenge_scalar(b"f")
     }
 
-    fn commit_program(prog: &[u8], blinding: &[u8]) -> Scalar {
-        let mut t = Transcript::new(b"ZkVM.predicate");
-        t.commit_bytes(b"blinding", blinding);
-        t.commit_bytes(b"prog", &prog);
-        t.challenge_scalar(b"h")
+    /// Return predicate tree as predicate.
+    pub fn tree(
```
oleganza: let's remove this method so there's no too-thin wrappers. User can simply do `Predicate::Tree(PredicateTree::new())` and whenever it's too cumbersome, we can `impl From<PredicateTree> for Predicate` and do `PredicateTree::new().into()` instead.

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
     }
 
-    #[test]
-    fn valid_program_commitment() {
-        let prog = Program::build(|p| p.drop());
-        let pred = Predicate::unblinded_program(prog.clone());
-        let op = pred.prove_program_predicate(&bytecode(&prog), &[]);
-        assert!(op.verify().is_ok());
+    /// Returns the adjustment factor for signing
+    pub fn adjustment_factor(&self) -> Scalar {
```
oleganza: probably add a comment that instead of this API we'd rather return a "key witness" object like musig::Multikey and  the adjustment factor will be stored right there.

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
-        let prog2 = Program::build(|p| p.dup(1));
-        let pred = Predicate::unblinded_program(prog);
-        let op = pred.prove_program_predicate(&bytecode(&prog2), &[]);
-        assert!(op.verify().is_err());
+    /// Creates the call proof and returns that with the program at an index.
+    pub fn create_callproof_program(
+        &self,
+        prog_index: usize,
+    ) -> Result<(CallProof, Program), VMError> {
+        let poss_leaf = &self.leaves[2 * prog_index];
```
oleganza:
```suggestion
        let possible_leaf = &self.leaves[2 * prog_index];
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
+        let leaf_index = match poss_leaf {
+            PredicateLeaf::Blinding(_) => 2 * prog_index + 1,
+            PredicateLeaf::Program(_) => 2 * prog_index,
+        };
+        let tree = MerkleTree::build(b"ZkVM.taproot", &self.leaves);
+        let neighbors = tree.create_path(leaf_index).unwrap();
+        let call_proof = CallProof {
+            verification_key: self.key,
+            neighbors: neighbors,
```
oleganza:
```suggestion
            neighbors,
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
+        &self,
+        prog_index: usize,
+    ) -> Result<(CallProof, Program), VMError> {
+        let poss_leaf = &self.leaves[2 * prog_index];
+        let leaf_index = match poss_leaf {
+            PredicateLeaf::Blinding(_) => 2 * prog_index + 1,
+            PredicateLeaf::Program(_) => 2 * prog_index,
+        };
+        let tree = MerkleTree::build(b"ZkVM.taproot", &self.leaves);
+        let neighbors = tree.create_path(leaf_index).unwrap();
```
oleganza: `unwrap` is generally not good outside tests - use normal error handling over panicking.
```suggestion
        let neighbors = tree.create_path(leaf_index)?;
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
+            neighbors: neighbors,
+        };
+        let program = self.leaves[leaf_index].clone().to_program()?;
+        Ok((call_proof, program))
     }
 
-    #[test]
-    fn valid_disjunction1() {
-        let gens = PedersenGens::default();
+    fn commit_leaves(progs: &Vec<Program>, blinding_key: [u8; 32]) -> Vec<PredicateLeaf> {
```
oleganza: maybe this is clearer?
```suggestion
    fn create_merkle_leaves(progs: &Vec<Program>, blinding_key: [u8; 32]) -> Vec<PredicateLeaf> {
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
+            prog.encode(&mut buf);
+            t.commit_bytes(b"prog", &buf);
+        }
+
+        let mut leaves = Vec::new();
+        for prog in progs.iter() {
+            let blinding = t.challenge_scalar(b"blinding");
+            let blinding_leaf = PredicateLeaf::Blinding(blinding.to_bytes());
+            let program_leaf = PredicateLeaf::Program(prog.clone());
+            if blinding.to_bytes()[31] & 1 == 0 {
```
oleganza: Top bit is always 0 since scalars are encoded as little-endian and mod `2^252+...`, not `2^256`. Use the low bit instead:

```suggestion
            if blinding.to_bytes()[0] & 1 == 0 {
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
+        t.commit_bytes(b"key", &blinding_key);
+        for prog in progs.iter() {
+            let mut buf = Vec::new();
+            prog.encode(&mut buf);
+            t.commit_bytes(b"prog", &buf);
+        }
+
+        let mut leaves = Vec::new();
+        for prog in progs.iter() {
+            let blinding = t.challenge_scalar(b"blinding");
```
oleganza: Blinding factor does not need to be a scalar. Get raw 32-byte array `[u8;32]` instead.

zkvm/src/predicate.rs
oleganza: i'd use the lowest bit to keep it somewhat simpler (not that it matters - also if blinding was a scalar, top 4 bits would be zero, so choosing 8th-from-the-top bit is kinda weird):

```suggestion
            if blinding[0] & 1 == 0 {
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
+        // VerificationKey is a 32-byte array
+        // MerkleNeighbor is a 32-byte array
+        32 + 4 + self.neighbors.len() * 32
+    }
 
-        let pred = Predicate::disjunction(preds.clone()).unwrap();
-        let op = pred.prove_disjunction(&preds);
-        assert!(op.verify().is_ok());
+    pub fn parse(data: &[u8]) -> Result<Self, VMError> {
+        SliceReader::parse(data, |r| {
```
oleganza: Better to leave the decision of "should this blob be a callproof or callproof+otherstuff" to the user. Let's rename this method to `decode` and work on the SliceReader instance like other `decode` methods. Then, in the VM, when you perform `call` instruction, you'd treat the blob on the stack as a call proof and wrap parsing in `SliceReader::parse()` to indicate that such blob must be a callproof end-to-end.

oleganza: We need to know the number of neighbors, though. We can sacrifice one bit of the 32-bit `positions` number to indicate the size of the proof: for `n` neighbors, `n`th bit must be set to one.

```
Clauses         Neighbors         Bit pattern (lo->hi)
-----------------------------------------------------------
0               n/a               n/a
1               0                 10000000 00000000 00000000 00000000
2               1                 x1000000 00000000 00000000 00000000
3-4             2                 xx100000 00000000 00000000 00000000
5-8             3                 xxx10000 00000000 00000000 00000000
9-16            4                 xxxx1000 00000000 00000000 00000000
17-32           5                 xxxxx100 00000000 00000000 00000000
```

All-zero bit pattern is invalid.

Use `u32::leading_zeros()` to count leading zeroes and subtract that number from 31 to learn the number of neighbors: `n = 31-position.leading_zeros()`.

Also when building a tree, add check that number of items is `<= (1<<31)`.

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
+    }
+}
+
+impl MerkleItem for PredicateLeaf {
+    fn commit(&self, t: &mut Transcript) {
+        match self {
+            PredicateLeaf::Program(prog) => {
+                let mut buf = Vec::new();
+                prog.encode(&mut buf);
+                t.commit_bytes(b"predicate_leaf", &buf);
```
oleganza: It is paramount that blinding factors and programs must be domain-separated: the current implementation allows treating a blinding factor as a program, which is a security hole in multi-party contracts.

```suggestion
                t.commit_bytes(b"program", &buf);
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +151,195 @@ impl Into<CompressedRistretto> for Predicate {
...
+            }
+            PredicateLeaf::Blinding(bytes) => {
+                t.commit_bytes(b"predicate_leaf", &bytes.clone());
```
oleganza: see the comment above
```suggestion
                t.commit_bytes(b"blinding", &bytes.clone());
```

zkvm/src/predicate.rs
```diff
@@ -136,42 +105,43 @@ impl Predicate {
...
+    fn commit_taproot(key: &[u8], root: &[u8]) -> Scalar {
+        let mut t = Transcript::new(b"ZkVM.taproot");
+        t.commit_bytes(b"key", key);
+        t.commit_bytes(b"merkle", root);
+        t.challenge_scalar(b"h")
+    }
+
+    /// Verifies whether the current predicate is a commitment to a signing key `key` and Merkle root `root`.
+    /// Returns a `PointOp` instance that can be verified in a batch with other operations.
+    pub fn prove_taproot(&self, program: &Program, call_proof: &CallProof) -> PointOp {
```
oleganza: This method cannot take `Program`, it must take a bytecode — the Verifier is not going to parse the program only for it to be encoded for verification, and then parsed again during execution. 

```suggestion
    pub fn prove_taproot(&self, program: &[u8], call_proof: &CallProof) -> PointOp {
```

zkvm/src/vm.rs
```diff
@@ -713,44 +712,24 @@ where
     }
 
     fn call(&mut self) -> Result<(), VMError> {
-        // Pop program, blinding factor, contract, and predicate
-        let prog = self.pop_item()?.to_data()?;
-        let blinding = self.pop_item()?.to_data()?;
+        // Pop program, call proof, and contract
+        let program_data = self.pop_item()?.to_data()?;
+        let program = program_data.clone().to_program()?;
```
oleganza: The program is going to be interpreted during the VM execution. We need to preserve the bytecode as-is without parsing it here, as it's going to be parsed again by the VM.

oleganza:
@debnil Per my comments about not parsing the program prematurely, we'd probably need a `enum ProgramWitness` (or smth like that) that holds either `::Opaque(Vec<u8>)` (on the verifier's end) or `::Program(Program)`. This way, impls like `MerkleItem` can implement hashing both for prover and verifier.

zkvm/src/predicate.rs
debnil: This makes sense to me in theory. However, the only use of `program` in this function is as an argument to `compute_root_from_path`. That particular argument is an `Entry`, which requires the implementation of `MerkleItem`. This doesn't seem possible for a byte-array. This seems like it might require some larger changes.. what do you think?

zkvm/src/predicate.rs
debnil: Good catch -- I think this is just a hold-over comment from a previous incorrect implementation. From the construction of the `PointOp` [here](https://github.com/interstellar/slingshot/pull/268/files/abe92d6b5cf4c4944d10ba4a12920edeb5cf9bff#diff-7e106625822af74f4132c28bcb9fe28bR128), this is the pure verification key `X`.

zkvm/src/predicate.rs
debnil: The above all makes sense! One small error: the number of neighbors `n = positions.leading_zeros()`.

zkvm/src/vm.rs
debnil: The only reason the bytecode is cast to a program is because `prove_taproot` requires that, as documented above. Changes to resolve the above situation should resolve this too.

zkvm/src/predicate.rs
```diff
@@ -27,19 +29,45 @@ pub enum Predicate {
...
 }
 
+/// Represents a ZkVM predicate tree.
 #[derive(Clone, Debug)]
-pub struct PredicateDisjunction {
-    preds: Vec<Predicate>,
-    // We precompute the disjunction predicate when composing it via `Predicate::or()`,
+pub struct PredicateTree {
+    /// Vector of the programs and blinding factors, stored as Merkelized predicate leafs.
+    pub leaves: Vec<PredicateLeaf>,
```
oleganza: keeping these fields public allows holder of mut-borrow to modify the fields directly and getting other private fields out of sync. Let's make them private and expose necessary getter methods if needed?

zkvm/src/program.rs
```diff
@@ -120,56 +131,51 @@ impl Program {
...
-        for pred in preds.iter() {
-            prog.push(pred.as_opaque());
+    }
+
+    /// Encodes a program witness into a buffer.
+    pub fn encode(&self, buf: &mut Vec<u8>) {
+        match self {
+            ProgramWitness::Program(prog) => prog.encode(buf),
+            ProgramWitness::Opaque(bytes) => {
+                for i in bytes.iter() {
```
oleganza:
```suggestion
                buf.extend_from_slice(&bytes);
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,205 @@ impl Into<CompressedRistretto> for Predicate {
...
 
-    #[test]
-    fn valid_disjunction2() {
-        let gens = PedersenGens::default();
+    /// Serializes the call proof to a byte array
+    pub fn encode(&self, buf: &mut Vec<u8>) {
+        encoding::write_point(self.verification_key.as_compressed(), buf);
+
+        let num_neighbors = self.neighbors.len();
+        let mut positions: u32 = 1 << 31 - num_neighbors;
```
oleganza: This length indicator seems to be growing from high to low. As number of neighbors grows, this bit is going to collide with right/left bits growing from the lowest end. Please re-read https://github.com/interstellar/slingshot/pull/268#discussion_r275136553

What we need is:

**Encoding**

Set the `1<<neighbour_count` for count indicator. Set `positions | (1<<i)` for each neighbor.

**Decoding**

Read `leading_zeros()`, if 32, fail (proof for empty tree).

Then do `neighbors_count = 31 - leading_zeros()` to determine number of neighbours (if it's 0, it means tree contains one item, there are no neighbours and no bits are used for left/right indication).

Then, use `positions & (1 << i) == 0` check to distinguish left from right for i-th neighbor.

zkvm/src/types.rs
```diff
@@ -59,6 +59,12 @@ pub enum Data {
 
     /// An input object (claimed UTXO).
     Output(Box<Output>),
+
+    /// A call proof.
+    CallProof(CallProof),
```
oleganza: Not sure that there's value to hold the unparsed `Data::CallProof` variant. It does not contain any secret data, and encoding/decoding it on prover's end won't make a performance difference. For simplicity, i'd remove it and use `Data::Opaque` which must be parsed into CallProof both by prover and verifier VMs.

zkvm/src/types.rs
```diff
@@ -59,6 +59,12 @@ pub enum Data {
...
+
+    /// A program witness.
+    ProgramWitness(ProgramWitness),
```
oleganza: There's some redundancy here: `Data::Opaque` and `Data::Program` already support verifier's and prover's representations of the program, respectively. What if we kept only them, and had `data.to_program_witness()` method that converts `Data::Opaque->ProgramWitness::Opaque` and `Data::Program->ProgramWitness::Program`, and fails otherwise. Then it still would avoid premature parsing, and there would be no redundancy. 

We already do the similar dual-downcast of Opaque/Nonopaque types with `Data::to_predicate` and `Data::to_commitment` - check them out.

zkvm/src/types.rs
```diff
@@ -216,6 +228,15 @@ impl Data {
         }
     }
 
+    /// Downcast the data item to a `CallProof` type.
+    pub fn to_call_proof(self) -> Result<CallProof, VMError> {
```
oleganza: per comment above - we can probably parse the callproof on prover also, so no need for this method.

zkvm/src/types.rs
```diff
@@ -191,6 +199,10 @@ impl Data {
         match self {
```
oleganza: per comment above, let's remove `to_program` entirely and use `to_program_witness` instead where needed, which converts Data::Opaque to PW::Opaque and Data::Program to PW::Program, respectively.

zkvm/src/types.rs
```diff
@@ -228,6 +249,15 @@ impl Data {
         }
     }
 
+    /// Downcast the data item to a `ProgramWitness` type.
+    pub fn to_program_witness(self) -> Result<ProgramWitness, VMError> {
+        match self {
+            Data::Opaque(data) => Ok(ProgramWitness::Opaque(data)),
+            Data::ProgramWitness(pw) => Ok(pw),
```
oleganza: per comment above (keeping ::Opaque and ::Program only, w/o ::ProgramWitness), this becomes 
```suggestion
            Data::Program(p) => Ok(ProgramWitness::Program(p)),
```

oleganza:
Final note: `ProgramWitness` sounds like it's non-opaque because it's "witness" (which is unavailable in opaque variants by definition).

Let's find a better name for it, like `ProgramItem` (and therefore, `Data::to_program_item`). And its variants named `::Bytecode(Vec<u8>)` and `::Program(Program)`.

zkvm/src/predicate.rs
oleganza: AFAIU, `leading_zeros` is the number of higher zero-bits (when in BE, those are left-most bits which is extra confusing).

From rust docs:

```rust
let n = u8::max_value() >> 2; // BE: 0011 1111, LE: 1111 1100
assert_eq!(n.leading_zeros(), 2);
```

oleganza: (also, `<<` and `>>` operators use BE notation, so we shift towards higher bits using `<<`, while my table above is using lo->hi (LE) notation)

zkvm/src/predicate.rs
oleganza: image

debnil: Aha, I think the second comment explains my confusion! Thanks.

zkvm/src/predicate.rs
```diff
@@ -27,19 +29,45 @@ pub enum Predicate {
...
 #[derive(Clone, Debug)]
-pub struct PredicateDisjunction {
-    preds: Vec<Predicate>,
-    // We precompute the disjunction predicate when composing it via `Predicate::or()`,
+pub struct PredicateTree {
+    /// Vector of the programs and blinding factors, stored as Merkelized predicate leafs.
+    leaves: Vec<PredicateLeaf>,
+    /// Verification key for the tree.
+    key: VerificationKey,
+    blinding_key: [u8; 32],
```
oleganza:
```suggestion
    /// Random seed from which we derive individual blinding factors for each leaf program.
    blinding_key: [u8; 32],
```

zkvm/src/predicate.rs
```diff
@@ -136,42 +118,35 @@ impl Predicate {
...
+        t.challenge_scalar(b"h")
+    }
+
+    /// Verifies whether the current predicate is a commitment to a signing key `key` and Merkle root `root`.
+    /// Returns a `PointOp` instance that can be verified in a batch with other operations.
+    pub fn prove_taproot(&self, program_item: &ProgramItem, call_proof: &CallProof) -> PointOp {
+        let key = &call_proof.verification_key;
+        let neighbors = &call_proof.neighbors;
+        let root = MerkleTree::compute_root_from_path(b"ZkVM.taproot", program_item, neighbors);
+        let h = Self::commit_taproot(&key.into_compressed().to_bytes(), &root);
```
oleganza: every use of `commit_taproot` has to convert a verification key to byteslice. How about we do so inside that function instead and take `key:&VerificationKey` parameter?

```suggestion
        let h = Self::commit_taproot(&key, &root);
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
-    use super::*;
-    use bulletproofs::PedersenGens;
+impl PredicateTree {
+    /// Creates new predicate tree with a verification key and a list of programs
+    pub fn new(
+        key: Option<VerificationKey>,
+        progs: Vec<Program>,
+        blinding_key: [u8; 32],
+    ) -> Result<Self, VMError> {
+        let key = key.unwrap_or_else(|| Predicate::unsignable_key());
```
oleganza:
```suggestion
        // If the key is None, use a point with provably unknown discrete log w.r.t. primary basepoint.
        let key = key.unwrap_or_else(|| Predicate::unsignable_key());
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
 
-    #[test]
-    fn invalid_program_commitment() {
-        let prog = Program::build(|p| p.drop());
-        let prog2 = Program::build(|p| p.dup(1));
-        let pred = Predicate::unblinded_program(prog);
-        let op = pred.prove_program_predicate(&bytecode(&prog2), &[]);
-        assert!(op.verify().is_err());
+    /// Creates the call proof and returns that with the program at an index.
+    pub fn create_callproof_program(
```
oleganza: Currently it sounds like it returns a special kind of program called "callproof program". The primary use of this method is to return a callproof, and the program is returned as a convenience.

```suggestion
    pub fn create_callproof(
```
wdyt?

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
+        &self,
+        prog_index: usize,
+    ) -> Result<(CallProof, Program), VMError> {
+        let possible_leaf = self.get_leaf(2 * prog_index);
```
oleganza:
```suggestion
        // The `prog_index` is used over the list of the programs,
        // but the actual tree contains also contains blinding factors,
        // so we need to adjust the index accordingly.
        // Blinding factors are located randomly to the left or right of the program leaf,
        // so we simply pick the left leaf, check it, and if it's not the program, pick the right one instead.
        let possible_leaf = self.get_leaf(2 * prog_index);
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
+        let leaf_index = match possible_leaf {
+            PredicateLeaf::Blinding(_) => 2 * prog_index + 1,
+            PredicateLeaf::Program(_) => 2 * prog_index,
+        };
+        let leaves = self.get_leaves();
+        let tree = MerkleTree::build(b"ZkVM.taproot", leaves);
+        let neighbors = tree.create_path(leaf_index)?;
+        let verification_key = self.get_key().clone();
```
oleganza: No need for `get_key` method as we are accessing the private field within the method of the same type.

```suggestion
        let verification_key = self.key.clone();
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
+        let call_proof = CallProof {
+            verification_key,
+            neighbors,
+        };
+        let program = self.get_leaf(leaf_index).clone().to_program()?;
```
oleganza: Same for `get_leaf`. Cf. https://github.com/interstellar/slingshot/pull/268/files#r276778503

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
+        }
 
-        // dummy predicates
-        let preds = vec![
-            Predicate::Opaque(gens.B.compress()),
-            Predicate::Opaque(gens.B_blinding.compress()),
-        ];
+        let mut leaves = Vec::new();
+        for prog in progs.iter() {
+            let blinding = rand::thread_rng().gen::<[u8; 32]>();
```
oleganza: the role of the `blinding_key` is to construct the entire tree deterministically.

```suggestion
            let blinding = [0u8;32]; 
            t.challenge_bytes("blinding", &mut blinding);
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
+            let blinding_leaf = PredicateLeaf::Blinding(blinding);
+            let program_leaf = PredicateLeaf::Program(ProgramItem::Program(prog.clone()));
+            if blinding[0] & 1 == 0 {
```
oleganza:
```suggestion
            // sacrifice one bit of entropy in the blinding factor to make the position of the program
            // random and make the tree indistinguishable from non-blinded trees.
            if blinding[0] & 1 == 0 {
```

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
+        &self,
+        prog_index: usize,
+    ) -> Result<(CallProof, Program), VMError> {
+        let possible_leaf = self.get_leaf(2 * prog_index);
+        let leaf_index = match possible_leaf {
+            PredicateLeaf::Blinding(_) => 2 * prog_index + 1,
+            PredicateLeaf::Program(_) => 2 * prog_index,
+        };
+        let leaves = self.get_leaves();
+        let tree = MerkleTree::build(b"ZkVM.taproot", leaves);
```
oleganza: No need for `get_leaves` method, as we can access private fields on `self` directly.

```suggestion
        let tree = MerkleTree::build(b"ZkVM.taproot", &self.leaves);
```

zkvm/src/program.rs
```diff
@@ -120,56 +131,69 @@ impl Program {
...
+        match self {
+            ProgramItem::Program(prog) => Ok(prog),
+            ProgramItem::Bytecode(_) => return Err(VMError::TypeNotProgram),
+        }
+    }
+
+    /// Downcasts a program item into a vector of bytes.
+    pub fn to_bytes(self) -> Result<Vec<u8>, VMError> {
+        match self {
+            ProgramItem::Program(_) => return Err(VMError::TypeNotProgram),
```
oleganza: If you are sure we don't need this method to force-covert non-opaque program to bytes, then it's better to rename it to something sounding more like "downcast", not encoding. E.g. `to_bytecode`. And clarify that the ::Program is not encoded automatically - for this one must use `encode`.

zkvm/src/vm.rs
```diff
@@ -696,44 +696,25 @@ where
...
-            preds.insert(0, self.pop_item()?.to_data()?.to_predicate()?);
-        }
-        let mut contract = self.pop_item()?.to_contract()?;
-        let p = &contract.predicate;
-        self.delegate
-            .verify_point_op(|| p.prove_disjunction(&preds))?;
-
-        contract.predicate = preds.remove(k as usize);
-        self.push_item(contract);
+        self.continue_with_program(program_data.to_program_item()?)?;
```
```suggestion
        self.continue_with_program(program_item)?;
```

zkvm/src/vm.rs
```diff
@@ -696,44 +696,25 @@ where
     }
 
     fn call(&mut self) -> Result<(), VMError> {
-        // Pop program, blinding factor, contract, and predicate
-        let prog = self.pop_item()?.to_data()?;
-        let blinding = self.pop_item()?.to_data()?;
+        // Pop program, call proof, and contract
+        let program_data = self.pop_item()?.to_data()?;
+        let program_item = program_data.clone().to_program_item()?;
```
```suggestion
        let program_item = self.pop_item()?.to_data()?.to_program_item()?;
```

and let's remove unneeded `program_data `

zkvm/src/program.rs
```diff
@@ -120,56 +131,69 @@ impl Program {
...
-    {
-        choose_fn(PredicateTree {
-            prog: self,
-            pred: pred,
-        })?;
+        pred_tree: PredicateTree,
+        prog_index: usize,
+    ) -> Result<&mut Program, VMError> {
+        let (call_proof, program) = pred_tree.create_callproof_program(prog_index)?;
+        self.push(Data::Opaque(call_proof.to_bytes())).call();
```
oleganza: why are there two `call` instructions?

oleganza: i think tests are not using `choose_call` API

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
+        call_proof.neighbors = neighbors;
+        Ok(call_proof)
+    }
+
+    /// Serializes the call proof to a byte array
+    pub fn encode(&self, buf: &mut Vec<u8>) {
+        encoding::write_point(self.verification_key.as_compressed(), buf);
+
+        let num_neighbors = self.neighbors.len();
+        let mut positions: u32 = 1 << num_neighbors;
```
oleganza: can we copy my description of the bit pattern and the table here as a big descriptive comment?

zkvm/src/predicate.rs
```diff
@@ -181,75 +156,221 @@ impl Into<CompressedRistretto> for Predicate {
...
+        for n in &self.neighbors {
+            match n {
+                MerkleNeighbor::Left(l) => encoding::write_bytes(l, buf),
+                MerkleNeighbor::Right(r) => encoding::write_bytes(r, buf),
+            }
+        }
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut buf = Vec::new();
```
oleganza:
```suggestion
        let mut buf = Vec::with_capacity(self.serialized_length());
```

zkvm/src/predicate.rs
```diff
@@ -68,66 +90,26 @@ impl Predicate {
...
-    pub fn unblinded_program<T: Into<Program>>(x: T) -> Self {
-        Predicate::Program(x.into(), Vec::new())
-    }
-
     /// Downcasts the predicate to a verification key
     pub fn to_key(self) -> Result<VerificationKey, VMError> {
         match self {
             Predicate::Key(k) => Ok(k),
+            Predicate::Tree(t) => {
+                let ovk = VerificationKey::from_compressed(t.precomputed_point);
```
oleganza: you can use `.ok_or(VMError::InvalidPoint)` which does the same job as your match statement.

PR #269, opened 2019-04-04
.circleci: test go code using circle

tessr:


tessr:
I believe this needs to end up on main before I can build anything on circle with it.

vickiniu:
Do we need to add Circle to our gh repo? or is it just not turned on until `.circleci/config.yml` lands?

tessr:
We do need to add it to get it to show up as a "check," but we can still run builds here: https://circleci.com/gh/interstellar/slingshot/

.circleci/config.yml
```diff
@@ -0,0 +1,25 @@
+# Check https://circleci.com/docs/2.0/language-go/ for more details
+version: 2
+jobs:
+  build:
+    docker:
+      # specify the version
+      - image: circleci/golang:1.9
```
bobg: Should this not be at least 1.11 (for new modules behavior)?

.circleci/config.yml
```diff
@@ -0,0 +1,25 @@
...
+    ####   /go/src/github.com/circleci/go-tool
+    ####   /go/src/bitbucket.org/circleci/go-tool
+    working_directory: /go/src/github.com/interstellar/slingshot
+    steps:
+      - checkout
+
+      # specify any bash command here prefixed with `run: `
+      - run: go vet ./slidechain/...
+      - run: test `gofmt -s -l ./slidechain | grep -vF /vendor/ | wc -l` -eq 0
+      - run: go test -v ./slidechain/... -timeout 60m
```
bobg: Please add a trailing newline.

.circleci/config.yml
tessr: good call!

tessr:
Cross-posting from Slack, but: I'm giving up on using Circle, for now, for a few reasons: 

1. no easy support for rust nightlies. we’d have to build and maintain our own docker container and update every time we updated the nightly
2. performance doesn’t seem to be better. builds often queue for a long time for seemingly no reason. the page hasn't even been rendering correctly? it seems like they've had a few incidents over the last few days, which explains some of this, but it's not encouraging. 

If we still want to move away from Travis in a few weeks, we can investigate something self-hosted.

PR #274, opened 2019-04-16
utreexo: new Rust crate

bobg:
Based on the Go implementation at github.com/bobg/utreexo.


utreexo/Cargo.lock
```diff
@@ -0,0 +1,6 @@
+# This file is automatically @generated by Cargo.
```
oleganza: Libraries generally do not put Cargo.lock under version control. This is only useful for applications to lock down all dependencies. Libraries can be used with differently locked down applications, so they only specify versions per SemVer rules to allow upgrades.

utreexo/src/update.rs
```diff
@@ -0,0 +1,45 @@
+use super::hash::Hash;
+use super::proof::{Proof, ProofStep};
+use super::utreexo::Utreexo;
+
+use std::collections::HashMap;
+
+pub struct Update {
+    pub(crate) u: Utreexo,
```
oleganza: nitpick: could we have more spelled out names, because `u.u.` and `hh` in the code are kinda hard to read :-)

eupn:
Didn't know that there is a crate here, so I've ended up doing the same work 🤦‍♂: https://github.com/eupn/utreexo-rs

I will maintain this crate on crates.io so it may be unnecessary to embed utreexo crate into a project but use it as a Cargo.toml dependency.

oleganza:
Closed since we are using implementation in #308

PR #275, opened 2019-04-17
zkvm/blockchain: implement BlockchainState::apply_txlog

vickiniu:
Changes UTXO representation to match contract ID and implements `apply_txlog`, removing inputs and adding outputs to the UTXO set.

### DIFF
--- a/zkvm/src/blockchain/state.rs
+++ b/zkvm/src/blockchain/state.rs
@@ -3,13 +3,13 @@ use std::collections::HashSet;
 
 use super::block::{Block, BlockHeader, BlockID};
 use super::errors::BlockchainError;
-use crate::{Entry, TxLog, VMError, UTXO};
+use crate::{ContractID, Entry, TxLog, VMError};
 
 #[derive(Clone)]
 pub struct BlockchainState {
     pub initial: BlockHeader,
     pub tip: BlockHeader,
-    pub utxos: HashSet<UTXO>,
+    pub utxos: HashSet<ContractID>,
 
     pub initial_id: BlockID,
 }
@@ -47,17 +47,14 @@ impl BlockchainState {
             match entry {
                 // Remove input from UTXO set
                 Entry::Input(input) => {
-                    let utxo = input.as_utxo();
-                    if self.utxos.contains(&utxo) {
-                        self.utxos.remove(&utxo);
-                    } else {
+                    if !self.utxos.remove(&input) {
                         return Err(VMError::InvalidInput);
                     }
                 }
 
                 // Add output entry to UTXO set
                 Entry::Output(output) => {
-                    self.utxos.insert(UTXO::from_output(output));
+                    self.utxos.insert(output.id());
                 }
                 _ => {}
             }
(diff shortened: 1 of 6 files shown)


zkvm/src/contract.rs
```diff
@@ -118,6 +119,11 @@ impl ContractID {
         &self.0
     }
 
+    /// Returns the contractID as a UTXO
+    pub fn as_utxo(&self) -> UTXO {
+        UTXO(self.0)
```
bobg: In ZkVM, utxo and contractID are synonyms. That doesn't mean we should have two distinct types! Can you replace uses of the UTXO type with the ContractID type?

zkvm/src/blockchain/state.rs
```diff
@@ -45,19 +45,83 @@ impl BlockchainState {
     fn apply_txlog(&mut self, txlog: &TxLog) -> Result<(), VMError> {
         for entry in txlog.iter() {
             match entry {
-                Entry::Input(contractID) => {
-                    // xxx check contractID is in self.utxos
-                    // xxx remove contractID
+                // Remove input from UTXO set
+                Entry::Input(input) => {
+                    let utxo = input.as_utxo();
+                    if self.utxos.contains(&utxo) {
```
bobg: Is there a more Rust-idiomatic way to do this whole if/else block? Something like:

`self.utxos.remove_or_die(&utxo)?`

The error returned by `remove_or_die` would have to be converted to a `VMError::InvalidInput`. Is there an idiom for that? If not, maybe this code is good the way it is.

zkvm/src/blockchain/state.rs
```diff
@@ -45,19 +45,83 @@ impl BlockchainState {
...
+                Entry::Output(output0.clone()),
+                Entry::Output(output1.clone()),
+            ])
+            .unwrap();
+        state
+            .apply_txlog(&vec![Entry::Input(output0.id())])
+            .unwrap();
+
+        // Check that output0 was consumed
+        assert_eq!(state.utxos.contains(&output0.id().as_utxo()), false);
```
bobg: Is there a reason to prefer this over `assert!(!state.utxos.contains(...))`?

zkvm/src/contract.rs
vickiniu: Makes sense! That's even simpler :)

zkvm/src/blockchain/state.rs
vickiniu: Good question! There might very well be such a thing

zkvm/src/blockchain/state.rs
vickiniu: Not really! I'm happy to go with that

bobg: `!(!`

zkvm/src/blockchain/state.rs
```diff
@@ -47,17 +47,14 @@ impl BlockchainState {
             match entry {
                 // Remove input from UTXO set
                 Entry::Input(input) => {
-                    let utxo = input.as_utxo();
-                    if self.utxos.contains(&utxo) {
-                        self.utxos.remove(&utxo);
-                    } else {
+                    if !self.utxos.remove(&input) {
```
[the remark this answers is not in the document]
vickiniu: I think for `HashSet`, `remove` returns a `bool`! https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.remove

PR #276, opened 2019-04-17
zkvm/blockchain: execute tx list

vickiniu:
Implements the [Execute tx list](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#execute-transaction-list) procedure from the blockchain spec!

### DIFF
--- a/zkvm/Cargo.toml
+++ b/zkvm/Cargo.toml
@@ -18,6 +18,7 @@ rand = "0.6"
 subtle = "2"
 curve25519-dalek = { version = "1.0.1", features = ["serde"] }
 serde = { version = "1.0", features=["derive"] }
+rayon = "1.0"
 
 
 [dependencies.bulletproofs]
--- a/zkvm/Cargo.toml
+++ b/zkvm/Cargo.toml
@@ -18,7 +18,6 @@ rand = "0.6"
 subtle = "2"
 curve25519-dalek = { version = "1.0.1", features = ["serde"] }
 serde = { version = "1.0", features=["derive"] }
-rayon = "1.0"
 
 
 [dependencies.bulletproofs]
--- a/zkvm/src/blockchain/state.rs
+++ b/zkvm/src/blockchain/state.rs
@@ -78,7 +78,7 @@ impl BlockchainState {
         // Check that, for the current block version, this tx version is
         // supported. For block versions higher than 1, we do not yet know
         // what tx versions to support, so we accept all.
-        if block_version == 1 && block_version != tx.header.version {
+        if block_version == 1 && tx.header.version != 1 {
             return Err(BlockchainError::VersionReversion);
         }
 
--- a/zkvm/src/blockchain/state.rs
+++ b/zkvm/src/blockchain/state.rs
@@ -3,7 +3,7 @@ use std::collections::HashSet;
 
 use super::block::{Block, BlockHeader, BlockID};
 use super::errors::BlockchainError;
-use crate::{ContractID, Entry, TxLog, VMError};
+use crate::{ContractID, Entry, Tx, TxID, TxLog, VMError, Verifier};
 
 #[derive(Clone)]
 pub struct BlockchainState {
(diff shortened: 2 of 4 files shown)


zkvm/src/blockchain/state.rs
```diff
@@ -60,4 +60,29 @@ impl BlockchainState {
...
         Ok(())
     }
+
+    /// Executes a list of transactions, returning their tx IDs and tx logs.
+    pub fn execute_txlist(
+        txs: Vec<Tx>,
+        version: u64,
+        timestamp_ms: u64,
+    ) -> Result<Vec<(TxID, TxLog)>, BlockchainError> {
+        let bp_gens = BulletproofGens::new(256, 1);
```
bobg: I think we want to avoid allocating our own generators inside the API. Our functions should instead accept generators from the caller and pass them where they're needed. (Double-check with @oleganza and/or @cathieyun.)

zkvm/src/blockchain/state.rs
```diff
@@ -60,4 +60,29 @@ impl BlockchainState {
...
+
+        txs.iter()
```
bobg: Perhaps for a later PR: this step can be parallelized (as noted in the spec). An easy way to achieve this might be with [the rayon crate](https://rust-lang-nursery.github.io/rust-cookbook/concurrency/parallel.html).

zkvm/src/blockchain/state.rs
vickiniu: that makes sense to me!

zkvm/src/blockchain/state.rs
vickiniu: Thanks for linking `rayon` — this looks great! Updated to add rayon and process in parallel

zkvm/src/blockchain/state.rs
```diff
@@ -62,6 +63,30 @@ impl BlockchainState {
...
     }
+
+    /// Executes a list of transactions, returning their tx IDs and tx logs.
+    pub fn execute_txlist(
+        txs: Vec<Tx>,
+        bp_gens: BulletproofGens,
+        version: u64,
+        timestamp_ms: u64,
+    ) -> Result<Vec<(TxID, TxLog)>, BlockchainError> {
+        txs.par_iter()
```
oleganza: i'm worried about this dependency on a specific multi-threading library. What if we strive to decouple optimizations from the logic, and use generics/iterators so we can push the decision on how to assemble all the verification as high up as possible.

E.g. would be good to wrap the body of this loop into a method (maybe part of zkvm::Verifier?) that takes the external timestamp and the int "max_tx_version". And then one can wrap this into a short iterator or parallel iterator, whichever is needed.

zkvm/src/blockchain/state.rs
```diff
@@ -62,6 +63,30 @@ impl BlockchainState {
...
+            .map(|tx| {
+                if tx.header.mintime_ms > timestamp_ms || tx.header.maxtime_ms < timestamp_ms {
+                    return Err(BlockchainError::BadTxTimestamp);
+                }
+                if version == 1 && version != tx.header.version {
```
oleganza: `version` here i guess is the block version, maybe rename it accordingly? This is very different from tx version, so comparison between them is semantically incorrect. The goal here is to say "if the block version is current, then tx version must be among currently supported to prevent squatting future versions; but if it's in the future, we are running an outdated software, so we need to allow higher tx versions."

Currently the set of supported block versions is simply {1} and the set of supported tx versions is also {1}, but those are ("type-wise") completely different 1's. It's conceivable that some other changes in the protocol would warrant bumping block versions to 2,3,4, but still keeping tx version capped at 1 for all of them.

zkvm/src/blockchain/state.rs
vickiniu: That makes sense to me! I took out the body of the loop to only expose that, and to execute a tx list, the caller can wrap that in their own iterator, as you mentioned :)

zkvm/src/blockchain/state.rs
vickiniu: Did a name change and added a comment — lmk if that makes sense

zkvm/src/blockchain/state.rs
```diff
@@ -62,6 +63,29 @@ impl BlockchainState {
...
+        timestamp_ms: u64,
+    ) -> Result<(TxID, TxLog), BlockchainError> {
+        if tx.header.mintime_ms > timestamp_ms || tx.header.maxtime_ms < timestamp_ms {
+            return Err(BlockchainError::BadTxTimestamp);
+        }
+
+        // Check that, for the current block version, this tx version is
+        // supported. For block versions higher than 1, we do not yet know
+        // what tx versions to support, so we accept all.
+        if block_version == 1 && block_version != tx.header.version {
```
oleganza: Per my previous comment, let's not compare tx version with block version. Ideally, they should be of different types and not comparable with each other ;-)

We want to say "if block version is known, the tx version must be known (per block version, so blkver=1 may allow only txver=1, but blkver=2 may allow txver={1,2})".

```suggestion
        if block_version == 1 && tx.header.version != 1 {
```

zkvm/Cargo.toml
```diff
@@ -18,6 +18,7 @@ rand = "0.6"
 subtle = "2"
 curve25519-dalek = { version = "1.0.1", features = ["serde"] }
 serde = { version = "1.0", features=["derive"] }
+rayon = "1.0"
```
oleganza: do we need `rayon`?

```suggestion
```

zkvm/src/blockchain/state.rs
vickiniu: got it! makes sense :)

PR #279, opened 2019-04-18
zkvm/blockchain: implement `apply tx list` procedure

vickiniu:
Implements [apply transaction list](https://github.com/interstellar/slingshot/blob/main/zkvm/docs/zkvm-blockchain.md#apply-transaction-list) from blockchain spec

### DIFF
--- a/zkvm/src/blockchain/state.rs
+++ b/zkvm/src/blockchain/state.rs
@@ -62,6 +62,13 @@ impl BlockchainState {
 
         Ok(())
     }
+
+    fn apply_txlist(&mut self, txlist: &[TxLog]) -> Result<(), VMError> {
+        for txlog in txlist.iter() {
+            self.apply_txlog(txlog)?;
+        }
+        Ok(())
+    }
 }
 
 #[cfg(test)]
--- a/zkvm/src/blockchain/state.rs
+++ b/zkvm/src/blockchain/state.rs
@@ -1,4 +1,5 @@
 use bulletproofs::BulletproofGens;
+use std::borrow::Borrow;
 use std::collections::HashSet;
 
 use super::block::{Block, BlockHeader, BlockID};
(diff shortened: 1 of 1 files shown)


zkvm/src/blockchain/state.rs
```diff
@@ -62,6 +62,13 @@ impl BlockchainState {
 
         Ok(())
     }
+
+    fn apply_txlist(&mut self, txlist: &[TxLog]) -> Result<(), VMError> {
```
oleganza:
```suggestion
    fn apply_txlogs(&mut self, txlist: &[TxLog]) -> Result<(), VMError> {
```
wdyt?

Also, would be better to take not a memory-allocated slice of logs, but an iterator. If the user is verifying txs on the fly, they can apply them w/o allocating a storage for them. That's rust-idiomatic way of taking lists of things as an argument when all you need is to iterate over the list.

OTOH, maybe you don't need such method at all - simply use `for .. in .. {}` where you need it (that's probably one or two places anyway).

vickiniu: That makes sense! Updated to change name + take iterator as argument

zkvm/src/blockchain/state.rs
```diff
@@ -62,6 +63,17 @@ impl BlockchainState {
 
         Ok(())
     }
+
+    fn apply_txlogs<I>(&mut self, txlist: I) -> Result<(), VMError>
+    where
+        I: Iterator,
```
oleganza: IntoIterator is even better, since the user can send their Vec or &Vec directly w/o doing `.iter()`.
```suggestion
        I: IntoIterator,
```

oleganza:
This PR adds 12 lines for a job of a one-liner iterator. Maybe we can just loop and apply where we need to, w/o intermediate generic function that does it?

vickiniu:
Sure, that makes sense to me! Closing this PR :)

PR #280, opened 2019-04-18
zkvm/blockchain: compute utxo and tx root

vickiniu:
Makes changes to use an ordered `Vec` for the UTXO set (per the spec change in #278), and implements computing the utxo and tx root.

### DIFF
--- a/zkvm/src/blockchain/block.rs
+++ b/zkvm/src/blockchain/block.rs
@@ -37,18 +37,14 @@ impl BlockHeader {
         BlockID(result)
     }
 
-    pub fn make_initial(
-        timestamp_ms: u64,
-        refscount: u64,
-        utxos: Option<&Vec<ContractID>>,
-    ) -> BlockHeader {
+    pub fn make_initial(timestamp_ms: u64, refscount: u64, utxos: &Vec<ContractID>) -> BlockHeader {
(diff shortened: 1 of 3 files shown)


zkvm/src/blockchain/block.rs
```diff
@@ -117,3 +121,15 @@ impl Block {
         Ok(txlogs)
     }
 }
+
+impl Root {
+    /// Computes the Merkle txroot
+    pub fn tx(txids: &[TxID]) -> Root {
+        Root(MerkleTree::root(b"transaction_ids", txids))
```
oleganza:
```suggestion
        Root(MerkleTree::root(b"ZkVM.txroot", txids))
```

zkvm/src/blockchain/block.rs
```diff
@@ -117,3 +121,15 @@ impl Block {
...
+    }
+
+    /// Computes the Merkle utxoroot
+    pub fn utxo(utxos: &[ContractID]) -> Root {
+        Root(MerkleTree::root(b"utxos", utxos))
```
oleganza:
```suggestion
        Root(MerkleTree::root(b"ZkVM.utxoroot", utxos))
```

Issue #146, opened 2019-02-27
musig: multi-signer API


oleganza:
For multi-key/multi-party signing we need an API that handles both individual signatures (for `delegate`) and aggregated per-tx signature (`signtx`). 

1. Need to support arbitrarily nested levels of partial signing: at the tx level, at the multiple-parties-per-predicate level, and at arbitrary nested levels when one of the "parties" is themselves a multisig predicate.
2. For `signtx` we need to figure out how to verify or specify which inputs we are spending and which outputs we are expecting, in order to ensure that signing a specific txid is safe. Maybe we need to support "partial transaction execution" or something like this to make this as straightforward as possible.

Related issues: 
* Removing signing secrets from `Predicate::Key` witness: https://github.com/interstellar/slingshot/pull/142
* MuSig protocol implementation: #92.

PR #282, opened 2019-04-20
zkvm: integrate multi-message API

### LINKED ISSUE #176: musig: multi-message API for mitigating Russell's attack
### Background

See appendices A.2/A.3 in the [MuSig paper](https://eprint.iacr.org/2018/068.pdf).

### TL;DR

If you have two UTXOs using the same key, and you did not check ALL the utxos being using in the signed message, then your partial signature for one UTXO can be reused to also sign off inclusion of another UTXO.

### Solution for aggregation-by-verifier

The MuSig paper's appendix A.3 describes a solution:

1. The total signed message is formed with ordered pairs `(pubkey, item)` where `item` is a specific message signed by the given pubkey (e.g. a unique utxo id or contract id - see also #172).
2. Each party generates their own Fiat-Shamir challenge corresponding to their index. 
3. The per-party F-S challenges effectively cover the "delinearization" factor applied to each pubkey, so there is no need to apply it to pubkeys in prior steps.
4. The verifier needs to see each individual pubkey and individual message for this to work.

### Key aggregation by provers

In the cases when the verifier _does not_ see individual keys, that is, when they were compressed under n-of-n rule offline into a single predicate, each signer is supposed to be fully aware of a message they are signing.

This works fine in case of `delegate` (provided the contract's anchor is unique and covered by signature: #172).

For `signtx`, the signer must be "aware of" not just `txid`, but the specific pair of `(pubkey, item)` which is normally a `(predicate, utxo_id)`. They usually would ignore `txid` and only verify proofs of certain outputs to belong to the txid that they sign, since the txid itself and other signers' outputs are not relevant to the given signer.

### API

The multi-message API is inherently different from single-message API, since each key's delinearization factor is effectively an individual fiat-shamir challenge bound to the specific (key-submessage) pair.

This should be composable over the regular musig protocol for nested Multikeys via API for cumulative multiplicative adjustment factor (#236).

oleganza:
1. Merges `Output` type into `Contract` type, and stores precomputed `ContractID` within the contract.
2. Adds multi-message API from Musig to be used with `signtx` with the use of contract ID.
3. Adds contract ID to the transcript in `delegate` instruction.
4. Added `UnsignedTx` type returned by the zkvm::Prover. See also #146.
5. All tx-related types are placed into `tx.rs` module: TxHeader, UnsignedTx, Tx, VerifiedTx, TxID, TxEntry, TxLog.

Closes #176


token/src/token.rs
```diff
@@ -145,22 +143,34 @@ mod tests {
...
-            ))
-        })
+                }
+                None
+            })
+            .collect();
+
+        let mut signtx_transcript = Transcript::new(b"ZkVM.signtx");
+        signtx_transcript.commit_bytes(b"txid", &utx.txid.0);
+        let sig = Signature::sign_multi(
```
vickiniu: Will this `sign_multi` call also work if we have only one privkey, or do we need to call `sign_single` in that case?

cathieyun: sign_multi will still work with one privkey. It uses the "multimessage API" (A.3 / defends against russel's attack), which is different from the "single message API" (normal musig), so calling sign_multi and sign_single are not interchangeable. (the verification for the two protocols is different.

oleganza: Multi-message signature with 1 key-message pair is inherently different (domain-separated) from the single message signature. In case of `signtx` instruction, we need to use multi-message API even if there's only one input, to keep things consistent.

vickiniu: got it, that makes sense! thanks :)

PR #286, opened 2019-04-23
musig: update design doc

cathieyun:
Update the docs to match the code implementation


musig/README.md
```diff
@@ -464,3 +486,14 @@ Operation:
...
+
+### Changing challenge definition for Multikey case
+
+In order to have a consistent multi-party signer protocol between the Multikey and Multimessage cases, we changed the definition of the hash function `H_agg` in the Multikey case from `c_i = H_agg(<L>, X_i)` to `c_i = H_agg(<L>, i)`. This way, the `.challenge(...)` functions in the Multikey and Multimessage cases are dependent on the public key's index `i`, instead of the public key itself. The safety of the protocol is not changed, since the hash function `H_agg` also takes `<L>`, the encoding of the multiset of the the public keys, as an input, so the hash output is bound to a specific public key in the list both when it takes in `i` and `X_i`.
```
oleganza:
```suggestion
In order to have a consistent multi-party signer protocol between the Multikey and Multimessage cases, we changed the definition of the hash function `H_agg` in the Multikey case from `c_i = H_agg(<L>, X_i)` to `c_i = H_agg(<L>, i)`. This way, the `.challenge(...)` functions in the Multikey and Multimessage cases are dependent on the public key's index `i`, instead of the public key itself. The safety of the protocol is not changed, since the hash function `H_agg` also takes `<L>`, the encoding of the multiset of the the public keys, as an input, so the hash output is bound to a specific public key in the list both when it takes in `i` and `X_i`.

```

PR #287, opened 2019-04-23
zkvm: implement guaranteed optimization for cleartext constraints

### LINKED ISSUE #252: zkvm: guaranteed optimization for constant constraints
Like we cannot soft-fork more efficient arithmetic on non-secret expressions (aka "numbers"), we cannot do the same for constraints (aka "booleans"). So maybe it makes sense to have a clean "guaranteed optimization" for constant constraints too (that are created by doing `eq` on two constant expressions):

- [x] Change `eq` to create new variant `Constraint::Cleartext(bool)` if both arguments are _constant expressions_. Boolean is `true` if they are equal, `false` otherwise.
- [x] Change `and` in case one of the constraints is `Cleartext`: return the other argument as-is if it's true, or return `Cleartext(false)` if it's false.
- [x] Change `or` in case one of the constraints is `Cleartext`: return the other argument as-is if it's false, or return `Cleartext(true)` if it's true.
- [x] Change `not` for `Cleartext` argument: invert the boolean.
- [x] Change `verify` for `Cleartext` argument: fail execution if it's `false`, and simply drop the argument if it's `true`.

When flattening non-cleartext constraint, the `Cleartext()` case is guaranteed to not show up, as it's always optimized away, but compiler still requires matching on all enum variants. To make it clean, convert `true` to const-expression of `0` (not a typo), and `false` to const-expression of `1`. Plus, add a comment, that this is never supposed to be executed.

oleganza:
Closes #252

Partly depends on https://github.com/dalek-cryptography/bulletproofs/pull/271 — if that PR 271 is merged, we can replace `PartialEq` implementation for `Expression` with a simple `#[derive(PartialEq)]`.

### DIFF
--- a/zkvm/src/constraints.rs
+++ b/zkvm/src/constraints.rs
@@ -471,7 +471,8 @@ impl Add for Expression {
     }
 }
 
-// TBD: Remove this in once PartialEq is impl'd for r1cs::Variable and use #[derive(PartialEq)] instead.
+// TBD: Remove this once PartialEq is impl'd for r1cs::Variable and use #[derive(PartialEq)] instead.
+// See https://github.com/dalek-cryptography/bulletproofs/pull/271
 impl PartialEq for Expression {
     fn eq(&self, other: &Expression) -> bool {
         match (self, other) {
@@ -533,6 +534,7 @@ impl Into<CompressedRistretto> for Commitment {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use merlin::Transcript;
 
     #[test]
     fn expression_arithmetic() {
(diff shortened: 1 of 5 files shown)


zkvm/src/constraints.rs
```diff
@@ -92,11 +104,77 @@ impl Constraint {
...
         self,
         cs: &mut CS,
     ) -> Result<(r1cs::LinearCombination, Option<Scalar>), r1cs::R1CSError> {
         match self {
+            // equivalent to (a - b) == 0
+            Constraint::Cleartext(true) => {
+                panic!(
+                    "Cleartext constraint should be optimized away early in `verify` instruction"
+                );
+                //Ok((Scalar::zero().into(), Some(Scalar::zero())))
```
cathieyun: remove commented out code?

zkvm/src/constraints.rs
```diff
@@ -92,11 +104,77 @@ impl Constraint {
...
+            }
+            // equivalent to (a - b) == not-zero, but we choose and fix 1.
+            Constraint::Cleartext(false) => {
+                panic!(
+                    "Cleartext constraint should be optimized away early in `verify` instruction"
+                );
+                //Ok((Scalar::one().into(), Some(Scalar::one())))
```
cathieyun: same as above

oleganza:
Yeah, having tests sounds like a good idea :-) Will add.

The difference between Const Expressions and Cleartext Constraints is the same as between Expressions and Constraints, and between Numbers and Booleans.

Expression can evaluate to any scalar, Constraint can evaluate to any boolean. Boolean operations are only applicable to Constraints, and not to Expressions. So having Const Expression for arithmetic operations does not translate in any way to boolean operations that need to work on Constraint objects.

PR #288, opened 2019-04-23
musig+zkvm: spec updates

oleganza:
This adds specification for signature protocols used in ZkVM with proper cross-references.

### DIFF
--- a/musig/docs/musig-spec.md
+++ b/musig/docs/musig-spec.md
@@ -3,11 +3,180 @@
 This is a technical specification for single-message and multi-message signature protocols
 implemented with [Ristretto](https://ristretto.group) and [Merlin transcripts](https://merlin.cool).
 
+* [Definitions](#definitions)
+* [Single-message signature](#single-message-signature)
+* [Multi-message signature](#multi-message-signature)
+* [Multi-party signing](#multi-party-signing)
+
+## Definitions
+
+### Scalar
+
+A _scalar_ is an integer modulo [Ristretto group](https://ristretto.group) order 
+`|G| = 2^252 + 27742317777372353535851937790883648493`.
+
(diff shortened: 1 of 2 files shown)


musig/docs/musig-spec.md
```diff
@@ -0,0 +1,182 @@
...
+TBD: spec for offline key aggregation.
+
+
+## Multi-party signing
+
+When the prover consists of several independent parties, each of the above signature protocols must be
+amended with a multi-round computation of a nonce commitment to prevent malicious
+choice of the nonce by one of the signers.
+
+TBD: spec.
```
cathieyun: Maybe link to the `Signer` state transitions here for the multi-party signing spec?

musig/docs/musig-spec.md
```diff
@@ -0,0 +1,182 @@
...
+    ```
+9. Prover sends `s` to the verifier.
+10. Verifier checks the relation:
+    ```
+    s·B  ==  R + sum{c[i]·X[i]}
+    ```
+
+## Key aggregation
+
+TBD: spec for offline key aggregation.
```
cathieyun: Link to the key aggregation notes?
https://github.com/interstellar/slingshot/tree/7c9e5474e133720492261f7b8171c075b0de41dd/musig#key-aggregation

oleganza:
I'd like the spec to focus on the specifics on the protocol precise enough to make a compatible implementation, but not too tied into language specifics. And also to separate concepts. Like, separate multi-round nonce commitment verification/aggregation from the single/multi-message schnorr logic (since nonce exchange is private business of the prover(s)).

Design doc was originally made to help sketch the implementation. It could still be a useful documentation to the _code_, but for the spec i'd have something more focused and concise.

PR #296, opened 2019-04-26
zkvm: remove import/export instructions

oleganza:
We probably will use normal `issue`/`retire` instructions for pegging the external assets. If not, we could bring those back any time, but for now let's clean up the codebase from undesigned features.


zkvm/docs/zkvm-spec.md
```diff
@@ -1290,70 +1269,6 @@ Merges and splits `m` [wide values](#wide-value-type) into `n` [values](#value-t
 Immediate data `m` and `n` are encoded as two [LE32](#le32)s.
```
```suggestion
```

PR #297, opened 2019-04-26
zkvm: stellar-zkvm pegging protocol draft

oleganza:


oleganza:
Update: @jonjove has pointed out a possible attack on user's peg-out transaction, destroying their funds, and offered a possible solution.

#### Attack

1. User prepares a temp Stellar transaction that acts as _uniqueness anchor_.
2. User forms a withdrawal Stellar transaction that spends and closes that temp account, but also pays from the custodian's account the $V being exported.
3. User retires mapped $V on the sidechain, annotating it with the temp account's ID and sequence number.
4. User asks the custodian to co-sign the withdrawal transaction.
5. Custodian checks the proof of publication of the retirement tx, co-signs the withdrawal.
6. Meanwhile, an attacker sends $X to user's account, to make it sufficiently close to the trustline limit. Of course, if required $X is ≥ $V, then attacker does not actually deprive the user of $V. The attack relevant if the user's remaining trustline limit is less than 2x$V.
7. If the user does not bump that limit before publishing the withdrawal tx, the temp account's sequence number will be consumed, tx included in the ledger, but transfers will not happen. Since the retired funds are tied to a specific seqnum (for anti-replay reasons), user cannot try again to withdraw those funds.

#### Proposed solution

1. The spec should include creation of an offer that pays a dummy asset for $V amount. This "reserves" a buffer of $V for the trustline, so the attacker sending money to the account cannot squeeze it more than to that much.
2. The withdrawal tx has one more operation, preceding the payout of $V from custodian: destroying that offer, which clears room for receiving $V, even if trustline is maxed out.

I think there might be a simpler option: the user may pay to their temporary account, which is originally set up with max trustline, that's impossible to max out without paying the user more than $V.

oleganza:
Another point raised by @jonjove is that assets should not have `AUTH_REQUIRED` and also be `AUTH_IMMUTABLE` so that export process cannot be interrupted mid-way.

PR #304, opened 2019-05-10
zkvm: non-copyable Program, Expression and Constraint

### LINKED ISSUE #301: zkvm: make recursion expensive and/or visible to prevent DoS
### Problem

To prevent a class of DoS attacks on ZkVM, we need to limit amount of cheap recursion that the users can perform, so that execution of the transaction's program is proportional to the program's length.

Currently it's conceivable that a relatively short program can duplicate a program string fixed number of times and execute it. Wrapping such program in another, similar to it, allows increasing the cost of execution exponentially while growing the overall transaction only linearly. 

### Some ways to solve the problem

1. Limit on the stack depth.
2. Limit a number of executed instructions (runlimit, like in TxVM).
3. Non-copyable programs as immediate data to instructions `delegate` and `call`, instead of taking a string from stack.
4. Non-copyable Program type created with `program:x` instead of `push:x`. `dup` refuses to copy it. `delegate`/`call` stay the same. 

The first two are more general, but also more ad-hoc: any limit could turn out to be insufficient in some cases, and too limiting in others.

Third option seems inelegant and may interfere with some cases where the `delegate`/`call` instruction needs to be executed within another contract, but the program must be provided later.

Fourth option clearly encodes programs as "statically typed" first-class items, so they can be passed around as any other items, but only executed once. At the same time, we avoid any artificial limits and hopefully can have bounded and reasonable worst-case cost per byte.

### Prior discussion

image

### LINKED ISSUE #306: zkvm: disallow copying of Expressions and Constraints
For background, see a similar issue with programs: #301

The Constraint type represents a tree of Constraints that get flatten into a set of R1CS constraints when `verify` instruction is used. With the repeated use of `dup` and `and`/`or` instructions, it is conceivable to construct an exponentially large constraint tree with many multipliers and linear constraints, in a linearly growing program. Even if we can estimate the validation cost from the r1cs proof length, the proof could be invalid and we'd be still DoSed with just a horribly big transient R1CS instance before we even verify the proof. 

Options to deal with it:

1. Disallow copying of Constraints. Then, the CS cost is proportional to the length of the program. For the same reasons we should probably also disallow copying of Expressions.
2. Try to commit to number of linear constraints and multipliers upfront, and then check that we don't run out of them during program execution. This seems to be complicated and brittle.

oleganza:
This adds a non-copyable type `Program` which is separate from `Data` type. Instructions `call` and `delegate` expect Program type, instruction `dup` fails if used with a Program.

Also, the `Expression` and `Constraint` are made non-copyable and non-droppable. Variables are left to be copyable, so they can participate in constraints and then be used in the construction of Values.

Rationale is covered in #301 & #306.

Closes #301
Closes #306

### DIFF
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -503,9 +503,10 @@ Output is a serialized [contract](#contract-type):
       Output  =  Anchor || Predicate  ||  LE32(k)  ||  Item[0]  || ... ||  Item[k-1]
       Anchor  =  <32 bytes>
    Predicate  =  <32 bytes>
-        Item  =  enum { Data, Value }
+        Item  =  enum { Data, Program, Value }
         Data  =  0x00  ||  LE32(len)  ||  <bytes>
-       Value  =  0x01  ||  <32 bytes> ||  <32 bytes>
+     Program  =  0x01  ||  LE32(len)  ||  <bytes>
+       Value  =  0x02  ||  <32 bytes> ||  <32 bytes>
 ```
 
 ### UTXO
@@ -1315,14 +1316,16 @@ Fails if the `prevoutput` is not a [data type](#data-type) with exact encoding o
 _items... predicate_ **output:_k_** → ø
 
 1. Pops [`predicate`](#predicate) from the stack.
-2. Pops `k` items from the stack.
+2. Pops `k` [portable items](#portable-type) from the stack.
 3. Creates a contract with the `k` items as a payload, the predicate `pred`, and anchor set to the [VM’s last anchor](#vm-state).
 4. Adds an [output entry](#output-entry) to the [transaction log](#transaction-log).
 5. Updates the [VM’s last anchor](#vm-state) with the [contract ID](#contract-id) of the new contract.
 
 Immediate data `k` is encoded as [LE32](#le32).
 
-Fails if VM’s [last anchor](#vm-state) is not set.
+Fails if:
+* VM’s [last anchor](#vm-state) is not set,
+* payload items are not [portable](#portable-type).
 
 
 #### contract
@@ -1330,14 +1333,16 @@ Fails if VM’s [last anchor](#vm-state) is not set.
 _items... pred_ **contract:_k_** → _contract_
 
 1. Pops [predicate](#predicate) `pred` from the stack.
-2. Pops `k` items from the stack.
+2. Pops `k` [portable items](#portable-type) from the stack.
 3. Creates a contract with the `k` items as a payload, the predicate `pred`, and anchor set to the [VM’s last anchor](#vm-state).
 4. Pushes the contract onto the stack.
 5. Update the [VM’s last anchor](#vm-state) with the [contract ID](#contract-id) of the new contract.
 
 Immediate data `k` is encoded as [LE32](#le32).
 
-Fails if VM’s [last anchor](#vm-state) is missing.
+Fails if:
+* VM’s [last anchor](#vm-state) is not set,
+* payload items are not [portable](#portable-type).
 
 
 #### log
--- a/zkvm/docs/zkvm-spec.md
+++ b/zkvm/docs/zkvm-spec.md
@@ -150,15 +150,14 @@ categories: [copyable types](#copyable-types) and [linear types](#linear-types).
 
 ### Copyable types
 
-Copyable types can be freely created, copied ([`dup`](#dup)), and destroyed ([`drop`](#drop)).
+Copyable types can be freely created, copied ([`dup`](#dup)), and destroyed ([`drop`](#drop)):
 
 * [Data](#data-type)
 * [Variable](#variable-type)
-* [Expression](#expression-type)
-* [Constraint](#constraint-type)
 
-Note: [program type](#program-type) is not copyable to avoid denial-of-service attacks
-via repeated execution of the same program.
+Note: the [program type](#program-type) is not copyable to avoid denial-of-service attacks
+via repeated execution of the same program that can be scaled exponentially while
+growing the transaction size only linearly.
 
 
 ### Linear types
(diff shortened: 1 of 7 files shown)


cathieyun:
Can you add a test showing that a proof trying to duplicate a program fails? (The tests checking that normal transaction succeed will check that normal program outputting succeeds, but there is no failure case check here).

oleganza:
@cathieyun added tests for Program, Expression and Constraint. Expanded the scope of this PR to extend to Expressions and Constraints.

zkvm/docs/zkvm-spec.md
```diff
@@ -147,44 +147,57 @@ transaction’s applicability to the blockchain.
 The items on the ZkVM stack are typed. The available types fall into two 
 categories: [copyable types](#copyable-types) and [linear types](#linear-types).
 
+
 ### Copyable types
 
-Copyable types can be freely created, copied ([`dup`](#dup)), and destroyed ([`drop`](#drop)).
+Copyable types can be freely created, copied ([`dup`](#dup)), and destroyed ([`drop`](#drop)):
```
bobg:
```suggestion
Copyable types can be freely created, copied (with [`dup`](#dup)), and destroyed (with [`drop`](#drop)):
```

zkvm/docs/zkvm-spec.md
```diff
@@ -147,44 +147,57 @@ transaction’s applicability to the blockchain.
...
 or [output](#output-structure) their contents themselves.
 * [Variables](#variable-type), [expressions](#expression-type) and [constraints](#constraint-type) have no meaning outside the VM state
-and its constraint system and therefore cannot be meaningfully ported between transactions.
+and its constraint system, and therefore cannot be meaningfully ported between transactions.
 
 
 ### Data type
 
-A _data type_ is a variable-length byte array used to represent signatures, proofs and programs.
+A _data type_ is a variable-length byte array used to represent [commitments](#pedersen-commitment), [scalars](#scalar), signatures and proofs.
```
bobg: Repeating some feedback from an earlier round of review on this spec: "type" is being used in a confusing way here. You can talk about "the data type" or "an instance of type data" but not "a data type." Consider:

> An object of the _data_ type is a variable-length byte array...

(Same thing with "program type" below.)

zkvm/docs/zkvm-spec.md
```diff
@@ -936,6 +944,14 @@ Pushes a [data](#data-type) `x` containing `n` bytes.
 Immediate data `n` is encoded as [LE32](#le32)
 followed by `x` encoded as a sequence of `n` bytes.
 
+#### program
+
+**program:_n_:_x_** → _program_
+
+Pushes a [program](#program-type) `x` containing `n` bytes.
+Immediate data `n` is encoded as [LE32](#le32)
+followed by `x` encoded as a sequence of `n` bytes.
```
bobg:
```suggestion
followed by `x`, as a sequence of `n` bytes.
```

I found the word "encoded" slightly confusing here.

zkvm/docs/zkvm-spec.md
```diff
@@ -1298,29 +1314,33 @@ Fails if the `prevoutput` is not a [data type](#data-type) with exact encoding o
 _items... predicate_ **output:_k_** → ø
 
 1. Pops [`predicate`](#predicate) from the stack.
-2. Pops `k` items from the stack.
+2. Pops `k` [portable items](#portable-types) from the stack.
```
bobg: If the top `k` items includes a non-portable item, is that an error or is it skipped? I know you intend it to be an error, but I think it's a little ambiguous as written.

I think the original wording ("Pops `k` items from the stack") plus the new "fails if" text below is fine.

zkvm/docs/zkvm-spec.md
```diff
@@ -1298,29 +1314,33 @@ Fails if the `prevoutput` is not a [data type](#data-type) with exact encoding o
...
+* payload items are not [portable](#portable-types).
 
 
 #### contract
 
 _items... pred_ **contract:_k_** → _contract_
 
 1. Pops [predicate](#predicate) `pred` from the stack.
-2. Pops `k` items from the stack.
+2. Pops `k` [portable items](#portable-types) from the stack.
```
bobg: Same comment.

zkvm/docs/zkvm-spec.md
```diff
@@ -1362,45 +1382,47 @@ _contract(P) proof prog_ **call** → _results..._
 7. Places the [payload](#contract-payload) on the stack (last item on top).
 8. Set the `prog` as current.
 
-Fails if the top two items are not [data](#data-type) or the third from top is not a [contract](#contract-type).
+Fails if:
+1. `prog` is not a [program type](#program-type),
```
bobg: Same comment about "a program type," "a data type," and "a contract type." I'd much prefer "a program," "a data item," and "a contract."

zkvm/docs/zkvm-spec.md
```diff
@@ -1362,45 +1382,47 @@ _contract(P) proof prog_ **call** → _results..._
...
-_contract prog sig_ **delegate** → _results..._
+_contract(P) prog sig_ **delegate** → _results..._
 
-1. Pop [data](#data-type) `sig`, [data](#data-type) `prog` and the [contract](#contract-type) from the stack.
-
-2. Place the [contract payload](#contract-payload) on the stack (last item on top), discarding the contract.
-
-3. Instantiate the [transcript](#transcript):
+1. Pop [data](#data-type) `sig`, [program](#program-type) `prog` and the [contract](#contract-type) from the stack.
+2. Reads the [predicate](#predicate) `P` from the contract.
```
bobg:
```suggestion
2. Read the [predicate](#predicate) `P` from the contract.
```

zkvm/docs/zkvm-spec.md
oleganza: How about renaming it into `String` in some other PR?

bobg: Sounds great. :+1:

PR #311, opened 2019-05-27
zkvm: Implement Encodable trait

### LINKED ISSUE #305: zkvm: internal Encodable trait
Organize `serialized_len`/`encode` methods under `Encodable` trait, so we can have a blanket implementation of `encode_to_vec()` to use in a few places instead of a noisy three-liner.

cyberbono3:
Resolves #305 
Implemented Encodable trait for encode() and serialized_length()

### DIFF
--- a/zkvm/src/contract.rs
+++ b/zkvm/src/contract.rs
@@ -73,6 +73,11 @@ impl Encodable for Contract {
         }
         size
     }
+    fn encode_to_vec(&self) -> Vec<u8> {
+        let mut buf = Vec::with_capacity(self.serialized_length());
+        self.encode(&mut buf);
+        buf
+    }
 }
 impl Contract {
     /// Creates a contract from a given fields
@@ -83,8 +88,7 @@ impl Contract {
             payload,
             anchor,
         };
-        let mut buf = Vec::with_capacity(contract.serialized_length());
-        contract.encode(&mut buf);
+        let mut buf = contract.encode_to_vec();
         contract.id = ContractID::from_serialized_contract(&buf);
         contract
     }
@@ -216,6 +220,9 @@ impl Encodable for PortableItem {
             PortableItem::Value(_) => 1 + 64,
         }
     }
+
+    fn encode_to_vec(&self) -> Vec<u8> {}
+
 }
 
 impl PortableItem {
--- a/zkvm/src/contract.rs
+++ b/zkvm/src/contract.rs
@@ -1,10 +1,10 @@
 use crate::constraints::Commitment;
 use crate::encoding;
+use crate::encoding::Encodable;
 use crate::encoding::SliceReader;
 use crate::errors::VMError;
 use crate::merkle::MerkleItem;
 use crate::predicate::Predicate;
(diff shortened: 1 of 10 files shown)


zkvm/src/program.rs
```diff
@@ -47,6 +47,26 @@ macro_rules! def_op {
     );
 }
 
+pub(crate) trait Encodable {
+   // Returns the serialized length of a datatype who implement this trait.
+   fn encode(&self, buf: &mut Vec<u8>);
+   /// Encodes a datatype who im0lements this trait into a buffer.
+   fn serialized_length(&self) -> usize;
```
oleganza: let's add another implemented method to factor out the common encode-to-vec pattern:

```rust
fn encode_to_vec(&self) -> Vec<u8> {
    let mut buf = Vec::with_capacity(self.serialized_length());
    self.encode(&mut buf);
    buf
}
```

Then, every time we do `buf=Vec::with_capacity(serialized_length) ... encode(&mut buf)` we can call that method instead.

zkvm/src/program.rs
```diff
@@ -47,6 +47,30 @@ macro_rules! def_op {
     );
 }
 
+pub(crate) trait Encodable {
+    // Returns the serialized length of a datatype who implement this trait.
```
oleganza:
```suggestion
    // Encodes receiver into bytes appending them to a provided buffer.
```

zkvm/src/program.rs
```diff
@@ -47,6 +47,30 @@ macro_rules! def_op {
...
+    fn encode(&self, buf: &mut Vec<u8>);
+    /// Encodes a datatype who im0lements this trait into a buffer.
```
oleganza:
```suggestion
    /// Returns precise length in bytes for the serialized representation of the receiver.
```

zkvm/src/program.rs
```diff
@@ -47,6 +47,30 @@ macro_rules! def_op {
...
+    fn serialized_length(&self) -> usize;
+    /// Encodes a datatype who im0lements this trait into a bytecode array
```
oleganza:
```suggestion
    /// Encodes the receiver into a newly allocated vector of bytes.
```

zkvm/src/contract.rs
```diff
@@ -63,8 +83,7 @@ impl Contract {
             payload,
             anchor,
         };
-        let mut buf = Vec::with_capacity(contract.serialized_length());
-        contract.encode(&mut buf);
+        let mut buf = contract.encode_to_vec();
```
oleganza:
```suggestion
        let buf = contract.encode_to_vec();
```

buffer is not mutable in this scope

zkvm/src/program.rs
```diff
@@ -47,6 +47,30 @@ macro_rules! def_op {
     );
 }
 
+pub(crate) trait Encodable {
```
oleganza: let's move this into `encoding.rs`?

cyberbono3: done

PR #315, opened 2019-07-04
zkvm: guaranteed optimization for `range`

### LINKED ISSUE #310: zkvm: guaranteed optimization for `range` instruction with const expressions
Current optimizations for constant Expressions and cleartext Constraints cover all arithmetic operations except for the `range:n` instruction. But this is the most costly of all of them, and has `n`-bits parameter to tune down that cost.

Suggestion:

1. Specify a guaranteed optimization for `range` instruction over _constant expressions_ such that the check is performed on the cleartext scalar bits: all bits in positions [64;255] must be zero.
2. Remove `:n` parameter, making the range work on full 64-bit integer range. This simplifies the design and provides better confidentiality for when instruction is used with actually secret data: amounts and time intervals.

cyberbono3:
Resolves #310 

Description:
Modified add_range_proof in  zkvm/src/vm.rs
Added in_range()  in zkvm/src/scalar_witness.rs

### DIFF
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -143,11 +143,7 @@ impl Encodable for Instruction {
             Instruction::Add => write(Opcode::Add),
             Instruction::Mul => write(Opcode::Mul),
             Instruction::Eq => write(Opcode::Eq),
-            Instruction::Range(n) => {
-                write(Opcode::Range);
-                let bit_width: BitRange = *n;
-                program.push(bit_width.into());
-            }
+            Instruction::Range => write(Opcode::Range),
             Instruction::And => write(Opcode::And),
             Instruction::Or => write(Opcode::Or),
             Instruction::Not => write(Opcode::Not),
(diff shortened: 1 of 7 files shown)


zkvm/src/vm.rs
```diff
@@ -672,20 +673,26 @@ where
...
+                    Err(VMError::InvalidBitrange))
+                }
+            },
             Expression::LinearCombination(terms, assignment) => {
-                (r1cs::LinearCombination::from_iter(terms), assignment)
+                spacesuit::range_proof(
+                    self.delegate.cs(),
+                    r1cs::LinearCombination::from_iter(terms),
+                    ScalarWitness::option_to_integer(assignment)?,
+                    bitrange,
```
oleganza: I'd hard-code 64-bit range here, by passing `BitRange::new(64)` and removing `:n` parameter from the `range` instruction. This would require update to the spec too, and the according tweaks in program.rs and ops.rs

zkvm/src/ops.rs
```diff
@@ -185,7 +181,7 @@ impl Encodable for Instruction {
             Instruction::Program(progitem) => 1 + 4 + progitem.serialized_length(),
             Instruction::Dup(_) => 1 + 4,
             Instruction::Roll(_) => 1 + 4,
-            Instruction::Range(_) => 1 + 1,
+            Instruction::Range => 1 + 1,
```
oleganza: since `range` does not have an argument anymore, the instruction takes 1 byte, not 2.

```suggestion
            Instruction::Range => 1,
```

oleganza:
Please read the error messages. The `Instruction::Range` needs to be updated too - there's no longer an argument to it.

oleganza:
I pushed a couple of fixes on the `opt-range` branch, so you can simply merge it into your branch in this PR. Patch: https://github.com/stellar/slingshot/commit/ee8430dcaa9d3ed79bc59efa6b634106e744cbf2

PR #320, opened 2019-07-11
zkvm: spec for pin/signpin instructions

oleganza:
This patch introduces a new type: **Pin**, a **p**ortable **i**de**n**tifier, an object containing unique 32 byte string for use with a new `signpin` instruction. Pin allows binding a signed program to any contract instance that carries a corresponding Pin object. 

### Changes summary

1. New `pin` instruction creates a new unique pin.
2. New `signpin` instruction delegates contract to a signed program, with signature tied to the contract's Pin (on top of the payload stack).
3. Existing `delegate` instruction is renamed to `signid`, since the signature covers the contract ID.
4. Instructions `signtx`, `signid`, `signpin` are grouped together in the instruction set.

### Rationale

Pinned signatures are more flexible than `signid` (née `delegate`), yet safe against replay attacks since ZkVM guarantees that pins cannot be duplicated. Pins allow modeling complex logical states of a contract, decoupled from the concrete instantiation of the contract.

The motivating use-case is a _payment channel_: pinned signature can be applied to any force-closed state of the contract and simultaneously bring the contract to the latest state.

### Alternatives

One alternative is to simply allow signature bound to a program and a pubkey. This is even more flexible (enabling oracles to sign predicates usable across independent contracts), but one must not reuse pubkey between independent domains.

Another alternative is `signtag` (#321): to sign with a string tag that must be pre-committed to by being stored within a contract. So if one reuses the keys, but keeps the tags separate, signatures cannot be maliciously or accidentally replayed. This is similar to `signpin`, but does not require a new Pin type. User must choose a unique tag themselves w/o any support.


oleganza:
Closing this in favor of #320.

PR #325, opened 2019-07-18
zkvm: use newer merlin api

bobg:
Also eliminates a couple of unused imports to silence all compiler warnings.


oleganza:
Looks good, but misses `cargo fmt`

PR #341, opened 2019-08-09
schnorr: new crate factored out of musig

### LINKED ISSUE #239: musig|schnorr: factor out two separate crates
MuSig module implements the entire signing/verification protocol, and has a lot of moving parts and its own spec. It's time to move it into its own crate.

- [x] Move `VerificationKey` and `Signature` types into `slingshot/schnorr` crate, that focuses on verification and single-key-signing only.
- [x] Move the code into a crate `slingshot/musig`, link from it to `schnorr`, and link zkvm with both schnorr and musig.
- [ ] Move musig-related errors under `MuSigError` enum.
- [ ] Change remaining `Option<>` return types into `Result<>` using appropriate `MuSigError::*` variant.
- [ ] Re-export ristretto base point, `Scalar`, `CompressedRistretto` type (so verification keys can be made out of raw points, to interop with Keytree)
- [ ] Keytree can link with this crate to use its `VerificationKey` type to return as derived leafs.

PR: #299.

oleganza:
Closes #239

In the next episodes:

- [ ] integrate into musig
- [x] batch verification trait (per #342)
- [ ] optional hex-encoded signature for human-readable serializer
- [ ] add VRF support


oleganza:
cc @hdevalence

oleganza:
I'm going to merge this and do integration in the subsequent PRs.

PR #362, opened 2019-09-19
demo: dockerfile

brahman81:


Dockerfile.zkvm-demo
```diff
@@ -0,0 +1,51 @@
...
+        libpq-dev \
+        libsqlite3-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+ENV PATH "/root/.cargo/bin:$PATH"
+
+# accept defaults
+RUN curl https://sh.rustup.rs -sSf > rustup-init.sh && \
+    bash rustup-init.sh -y && \
+    rustup install nightly-2019-08-19 && \
```
oleganza: can we read from `./demo/rust-toolchain` to install an appropriate nightly version?

brahman81: This is now complete.

PR #384, opened 2019-11-19
node: a simple p2p node implementation [WIP]

oleganza:
Work in progress on making a multi-node version of the demo app. Each node connects to peers, and also provides a local HTTP server as a UI.


oleganza:
replaced by #393

PR #407, opened 2020-03-11
zkvm: allow dropping Program/Expression/Constraint

oleganza:
Fixes https://github.com/stellar/slingshot/issues/406

The Program/Expression/Constraint are safe to drop types, but not safe to copy due to DoS risk. This was not implemented since we used a single category of "copyable types" for both copying and dropping. This patch introduces another category of "droppable types".

Specifically, Expression should be allowed to be dropped because of the `range` instruction that leaves it on stack. If one needs an inequality check, the left-on-stack expression must be dropped to clear the stack. In turn, `range` cannot consume the expression because then it cannot be used in downstream expressions, and we cannot allow copying for expressions for the DoS reasons.

Constraint and Program are also permitted to be dropped as there is no reason to prohibit that.

### DIFF
--- a/zkvm/src/ops.rs
+++ b/zkvm/src/ops.rs
@@ -32,7 +32,7 @@ pub enum Instruction {
     ///
     /// Drops `x` from the stack.
     ///
-    /// Fails if `x` is not a _copyable type_.
+    /// Fails if `x` is not a _droppable type_.
     Drop,
 
     /// _x(k) … x(0)_ **dup:_k_** → _x(k) ... x(0) x(k)_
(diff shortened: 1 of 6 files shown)


zkvm/src/ops.rs
```diff
@@ -322,18 +322,16 @@ pub enum Instruction {
...
+    /// _qty_ **fee** → _widevalue_
     ///
-    /// 1. Pops an 8-byte _string_ `qty` from the stack and decodes it as _LE64_ integer.
-    /// 2. Pops _value_ from the stack.
-    /// 3. Checks that value.qty is unblinded commitment to `qty` and value.flv is unblinded commitment to zero.
-    /// 4. Adds a _fee entry_ to the _transaction log_.
-    ///
-    /// The checks are performed with group elements directly, without any changes to a _constraint system_.
+    /// 1. Pops an 4-byte _string_ `qty` from the stack and decodes it as _LE32_ integer.
+    /// 2. Checks that `qty` is less or equal to `2^24`.
```
SanJue: 2. Checks that `qty` and accumulated fees are less or equal to `2^24`.

zkvm/src/ops.rs
```diff
@@ -322,18 +322,16 @@ pub enum Instruction {
...
+    /// 3. Pushes _wide value_ `–V`, with quantity variable constrained to `-qty` and with flavor constrained to 0.
+    ///    Both variables are allocated from a single multiplier.
+    /// 4. Adds a _fee entry_ to the _transaction log_ with the quantity `qty`.
     ///
     /// Fails if:
-    /// * value.qty does not match unblinded `qty`,
-    /// * value.flv is not an unblinded commitment to 0.
+    /// * `qty` is exceeding `2^24`.
```
SanJue: `qty` or accumulated fees are exceeding `2^24`

zkvm/src/ops.rs
oleganza: 👍

zkvm/src/ops.rs
oleganza: 👍

PR #415, opened 2020-03-29
p2p: impl AsyncRead/AsyncWrite for the cybershake channel

p0lunin:
⚠️ This change is incompatible with the previous version of cybershake.

This implements Tokio `AsyncRead` and `AsyncWrite` traits for the cybershake channel (`Incoming`, `Outgoing` structs), allowing users writing and reading plaintext data, while it's encrypted behind the scenes. Encryption happens in authenticated chunks (up to 4Kb). Each chunk has 18 bytes of overhead (u16 length prefix + 16-byte authentication tag) and chunks are finalized on each flush() or when 4Kb limit is reached.

Hello message within the cybershake is done via raw `write/read` methods without unnecessary framing.

### Limitations

Current `send_message`/`receive_message` functions are deprecated and implemented on top of the new traits with their own u32 length prefix per message. There's a hardcoded limit of 10Mb message in these interfaces. We will replace their use with Tokio Codecs anyway.

Currently we require user to pin+box the raw reader/writer objects (e.g. TCP socket halves), but in the future we'll see how to remove this requirement and use pin-project or whatever we have to do in async Rust.

Operations on ciphertext and plaintext are serial (not pipelined). That is: writing end does not accept plaintext until the ciphertext buffer is fully sent; reading end does not receive/decrypt another ciphertext buffer until the current one is fully consumed by the user. Not clear whether we should even worry about it.


p2p/src/cybershake.rs
```diff
@@ -46,11 +46,16 @@ use merlin::Transcript; // TODO: change for raw Strobe.
...
 
+use futures::task::{Context, Poll};
 use serde::{Deserialize, Serialize};
+use std::io::{Read, Write};
+use std::pin::Pin;
 
 /// The current version of the protocol is 0.
 /// In the future we may add more versions, version bits or whatever.
 const ONLY_SUPPORTED_VERSION: u64 = 0;
+const PLAINTEXT_BUF_SIZE: u16 = 4096;
```
oleganza, line 57: This constant should contain "max" in the name since the actual chunk could be smaller when flushed.
```suggestion
/// This is the maximum amount of bytes that we are collecting
/// before encrypting and sending into the underlying socket.
/// This is not a limit imposed by the receiver.
/// We can receive chunks up to 2**16-1 bytes long (where the length fits in u16).
const PLAINTEXT_MAX_BUF_SIZE: u16 = 4096;
```

p2p/src/cybershake.rs
```diff
@@ -46,11 +46,16 @@ use merlin::Transcript; // TODO: change for raw Strobe.
...
+const CIPHERTEXT_BUF_SIZE: u16 = PLAINTEXT_BUF_SIZE + 16; // 16 - auth tag
```
oleganza, line 58: same as above

```suggestion
const CIPHERTEXT_MAX_BUF_SIZE: u16 = PLAINTEXT_MAX_BUF_SIZE + 16; // 16 - auth tag
```

p2p/src/cybershake.rs
```diff
@@ -218,73 +240,255 @@ where
...
+            Poll::Ready(Ok(n)) => { n }
+            Poll::Ready(Err(e)) => return Poll::Ready(Err(e)),
+        }
+    };
+}
+
 impl<W: AsyncWrite + Unpin> Outgoing<W> {
+    /// Send a message with any length.
     pub async fn send_message(&mut self, msg: &[u8]) -> Result<(), Error> {
+        let mut written = 0;
```
oleganza, line 256: Now that we decouple framing of messages from encryption, we probably can use something like Tokio Codecs to implement message framing outside of the cybershake module. See Henry's excellent write up on how to use codecs: https://www.zfnd.org/blog/decoding-bitcoin-messages-with-tokio-codecs/

oleganza: In this PR, though, we may keep this API and refactor it later.

p2p/src/cybershake.rs
p0lunin: But this constant mean the size of the buffer, not mean size of chunks.

p2p/src/cybershake.rs
```diff
@@ -218,73 +240,255 @@ where
...
+            let n = ready!(poll);
+            self.ciphertext_sent += n;
+            if self.ciphertext_sent == self.ciphertext_buf.len() {
+                self.ciphertext_sent = 0;
+                self.ciphertext_buf.clear();
+                if self.plaintext_needs_flushing {
+                    self.cipher_buf();
+                    self.plaintext_needs_flushing = false;
+                }
+                return Poll::Ready(Ok(()));
```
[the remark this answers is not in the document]
p0lunin: Ok, I'll fixed it, and think about how reduce nested scopes.

p2p/src/cybershake.rs
```diff
@@ -218,73 +240,255 @@ where
...
+    }
 
-        // Write the length prefix and the ciphertext.
-        self.writer
-            .write(&encode_u64le(ciphertext.len() as u64)[..])
-            .await?;
-        self.writer.write(&ciphertext[..]).await?;
-        self.writer.flush().await?;
-        Ok(())
+    pub fn flush_write(&mut self, cx: &mut Context) -> Poll<Result<(), io::Error>> {
```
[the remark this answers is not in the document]
p0lunin: Ok, I'll rename this method.

This method should be called only if we have text in ct. And it should be a private, it my mistake.

p2p/src/cybershake.rs
```diff
@@ -218,73 +240,255 @@ where
...
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<Result<usize, io::Error>> {
+        let me = self.get_mut();
 
-        // length must include IV prefix (16 bytes)
-        if len < 16 {
-            return Err(Error::ProtocolError);
+        if me.plaintext_needs_flushing {
```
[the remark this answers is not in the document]
p0lunin: In that case no more need in two buffers. We can write plaintext in buffer, and then write ct in the same buffer. But i think it will be more slowly, because we need to wait while ct will be sent.

p2p/src/cybershake.rs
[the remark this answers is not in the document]
p0lunin: Duplicated

p2p/src/cybershake.rs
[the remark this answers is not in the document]
p0lunin: Duplicated

p2p/src/cybershake.rs
oleganza: fair.

p2p/src/cybershake.rs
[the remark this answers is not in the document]
oleganza: Yeah, that'd need one buffer. I think this microlag is okay to have in this initial version, after all, it's a backpressure. Double-buffering does not really reduce average latency, but allows smoothing the peaks if we are doing some sort of continuous streaming, which we certainly do not do here.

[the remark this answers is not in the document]
p0lunin: Reworked

p2p/src/cybershake.rs
```diff
@@ -218,73 +236,246 @@ where
...
-        // Write the length prefix and the ciphertext.
-        self.writer
-            .write(&encode_u64le(ciphertext.len() as u64)[..])
-            .await?;
-        self.writer.write(&ciphertext[..]).await?;
-        self.writer.flush().await?;
-        Ok(())
+    fn flush_pending_ciphertext(&mut self, cx: &mut Context) -> Poll<Result<(), io::Error>> {
+        loop {
+            let poll = self
```
oleganza, line 295: i'd make this method safely callable with nothing to send. So instead of `loop`, it'd be `while sent < buf.len()` and _after_ the loop we can check `if sent > 0` and then do cleanup.

oleganza: oops, and check that the flushing flag is set.

p2p/src/cybershake.rs
```diff
@@ -218,73 +236,246 @@ where
...
         self.kdf.append_u64(b"seq", self.seq);
         let mut key = [0u8; 32];
         self.kdf.challenge_bytes(b"key", &mut key);
 
         let ad = encode_u64le(self.seq);
 
         let ciphertext = Aes128PmacSiv::new(GenericArray::clone_from_slice(&key))
-            .encrypt(&[&ad], msg)
-            .map_err(|_| Error::ProtocolError)?;
+            .encrypt(&[&ad], &self.buf)
```
oleganza, line 279: You can use `encrypt_in_place` to avoid copying, and make sure that the buffer had 16-byte (IV_SIZE) padding before the plaintext data. Also leave 2 bytes for the length prefix.

p2p/src/cybershake.rs
```diff
@@ -218,73 +236,246 @@ where
...
+        cx: &mut Context<'_>,
+        buf: &[u8],
+    ) -> Poll<Result<usize, io::Error>> {
+        let me = self.get_mut();
 
-        // length must include IV prefix (16 bytes)
-        if len < 16 {
-            return Err(Error::ProtocolError);
+        if me.flushing {
+            ready!(me.flush_pending_ciphertext(cx));
```
oleganza, line 320: ...then, here i'd make flushing unconditional: if we have some CT to send, we need to flush it. The necessary checks will be inside the method.

p2p/src/cybershake.rs
```diff
@@ -218,73 +236,246 @@ where
...
         }
-        // Check the message length and fail before changing any of the remaining state.
-        if (len - 16) > self.message_maxlen {
-            return Err(Error::MessageTooLong(len - 16));
+
+        if me.buf.len() + buf.len() > PLAINTEXT_BUF_SIZE as usize {
```
oleganza, line 323: 1. check if buf.len() is zero and return immediately with Ready(Ok(0)).
2. Change `>` for `>=` - if we fit perfectly in the frame's size, we should send it.

p2p/src/cybershake.rs
```diff
@@ -218,73 +236,246 @@ where
...
+            // plaintext_buf has BUF_SIZE size, so subtract with overflow will be never.
+            let size_to_write = PLAINTEXT_BUF_SIZE as usize - me.buf.len();
+            if let Err(err) = Write::write(&mut me.buf, &buf[..size_to_write]) {
```
oleganza, line 326: why should we fail to write in the buffer? BTW, maybe change all the uses of Write::write on the Vec into `extend_from_slice`, which does not fail and also more clearly tells us that there's no fancy backend involved.

p2p/src/cybershake.rs
p0lunin: implemented

p2p/src/cybershake.rs
p0lunin: implement

p2p/src/cybershake.rs
p0lunin: Checks were moved into `flush_pending_ciphertext()`

p2p/src/cybershake.rs
p0lunin: 1. It will be return anyway in 
```rust
} else {
   Poll::Ready(Write::write(&mut me.buf, buf))
}
```
2. Ok.

p2p/src/cybershake.rs
p0lunin: Fixed

p2p/src/cybershake.rs
```diff
@@ -68,19 +73,33 @@ pub struct PublicKey {
...
     seq: u64,
     kdf: Transcript,
+    buf: Vec<u8>,
+    stage: WriteStage,
+    ciphertext_length: [u8; 2],
+    ciphertext_sent: usize,
+}
+
+enum WriteStage {
+    ReadPlaintext,
```
oleganza, line 86: that's a misleading name since we are "writing plaintext". Maybe try simply {Pt, CtLen, Ct}?

p0lunin: WriteStage were removed

p2p/src/cybershake.rs
```diff
@@ -218,63 +239,134 @@ where
...
         self.kdf.challenge_bytes(b"key", &mut key);
 
         let ad = encode_u64le(self.seq);
 
-        let ciphertext = Aes128PmacSiv::new(GenericArray::clone_from_slice(&key))
-            .encrypt(&[&ad], msg)
-            .map_err(|_| Error::ProtocolError)?;
+        let tag = Aes128PmacSiv::new(GenericArray::clone_from_slice(&key))
+            .encrypt_in_place_detached(&[&ad], &mut self.buf[TEXT_START_POSITION..])
+            .map_err(|_| unimplemented!())
```
oleganza, line 282: maybe do `.expect("never fails because we have just one header")`?

p2p/src/cybershake.rs
```diff
@@ -218,63 +239,134 @@ where
...
+            Poll::Ready(Ok(buf.len()))
         }
+    }
 
-        let mut ciphertext = Vec::with_capacity(len);
-        ciphertext.resize(len, 0u8);
-        self.reader.read_exact(&mut ciphertext[..]).await?;
+    fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
+        let me = self.get_mut();
+        if me.buf.len() == 0 {
```
oleganza, line 338: i think that's a bug. WriteZero error is intended when we try to write >0, but got Ok(0) from the underlying writer. But if we are flushing an empty buffer, that should be a no-op.

p2p/src/cybershake.rs
```diff
@@ -286,6 +378,121 @@ impl<W: AsyncRead + Unpin> Incoming<W> {
...
+        let ad = encode_u64le(seq);
+
+        let siv_tag = GenericArray::clone_from_slice(&self.buf.as_slice()[..16]);
+        Aes128PmacSiv::new(GenericArray::clone_from_slice(&key))
+            .decrypt_in_place_detached(
+                &[&ad],
+                &mut self.buf.as_mut_slice()[16..self.ciphertext_length as usize],
+                &siv_tag,
+            )
+            .unwrap();
```
oleganza, line 399: If decryption fails (normally, because auth tag is invalid), we should return a ProtocolError and not panic.

p2p/src/cybershake.rs
```diff
@@ -286,6 +378,121 @@ impl<W: AsyncRead + Unpin> Incoming<W> {
...
+        }
+
+        self.plaintext_length = pt_len;
+        self.ciphertext_length = 0;
+        self.ciphertext_read = 0;
+    }
+
+    fn read_buffer(&mut self, buf: &mut [u8]) -> Result<usize, io::Error> {
+        match Read::read(
+            &mut &self.buf[self.plaintext_read..self.plaintext_length],
```
oleganza, line 415: Can we please not use the Read trait here? It is a mutable trait that's supposed to update offset of the backend, but we are doing it manually here via the `self.plaintext_read` offset. Passing a mutable reference to our buffer makes me wonder what's going to be changed. It'd be much clearer if we checked for conditions separately:

1. If the requested buf is longer than what we have left - read till the end and return that much.
2. If it's shorter or equal - read as much as requested.

p2p/src/cybershake.rs
```diff
@@ -218,63 +239,134 @@ where
...
-        let ad = encode_u64le(seq);
+impl<W: AsyncRead + Unpin> Incoming<W> {
+    pub async fn receive_message(&mut self) -> Result<Vec<u8>, Error> {
+        let mut buf = vec![0u8; 4096];
 
-        let plaintext = Aes128PmacSiv::new(GenericArray::clone_from_slice(&key))
-            .decrypt(&[&ad], &ciphertext)
-            .map_err(|_| Error::ProtocolError)?;
+        let len = self.read(&mut buf).await.map_err(|e| match e.kind() {
+            io::ErrorKind::InvalidData => Error::ProtocolError,
```
oleganza, line 364: Why not wrap this kind into our Error::IoError, like all other io::ErrorKind variants?

p2p/src/cybershake.rs
```diff
@@ -286,6 +378,121 @@ impl<W: AsyncRead + Unpin> Incoming<W> {
...
+        if me.plaintext_length != 0 {
+            return Poll::Ready(me.read_buffer(buf));
+        }
+
+        if me.ciphertext_length == 0 {
+            loop {
+                let poll = me
+                    .reader
+                    .as_mut()
+                    .poll_read(cx, &mut me.buf[me.ciphertext_read as usize..2]);
```
oleganza, line 448: This line calls for a better code organization: currently it reads "we may have read some bytes, but we want 2 more", which safe in the overall context (because we expect 16 bytes of auth tag anyway), but locally does not look correct. Maybe better to have an enum for reading that encodes states properly:

enum ReadState{
Len(n:usize) // contains number of bytes read so far. We need to read `2-n`.
Body(n: usize, k:usize) // total length, and number of _body_ bytes read, ignoring the length prefix that we already have.
}

oleganza: Then we don't need `ciphertext_read`/`ciphertext_length` variables. 

When we are done, we revert to Len(0) state (aka "expecting to read the length prefix, but have 0 bytes so far").

p2p/src/cybershake.rs
p0lunin: Fixed

p2p/src/cybershake.rs
p0lunin: Changed returning value from Err to Ok

p2p/src/cybershake.rs
p0lunin: We cannot return any other error than `io::Error` from `poll_* methods. We can return `io::Error::Custom(io::ErrorKind::InvalidData, "protocol error")` as example

p2p/src/cybershake.rs
p0lunin: I choose io::ErrorKind::InvalidData as marker which signalize that error has been occured when we deciphered buffer.

p2p/src/cybershake.rs
p0lunin: Fixed

p2p/src/cybershake.rs
p0lunin: Added enum ReadState

p2p/src/cybershake.rs
```diff
@@ -286,6 +374,105 @@ impl<W: AsyncRead + Unpin> Incoming<W> {
     }
 }
 
+impl<W: AsyncRead + Unpin> Incoming<W> {
```
oleganza, line 377:
```suggestion
impl<R: AsyncRead + Unpin> Incoming<R> {
```

p2p/src/cybershake.rs
oleganza: let's kill our Error and use io::Error with InvalidData kind for version mismatch and encryption failures.

p2p/src/cybershake.rs
oleganza: let's kill our Error and use io::Error with InvalidData kind for version mismatch and encryption failures.

p2p/src/cybershake.rs
p0lunin: fixed

p2p/src/cybershake.rs
p0lunin: removed

p2p/src/cybershake.rs
p0lunin: Done

PR #421, opened 2020-04-16
p2p: encode messages using Tokio Codecs instead of serde/bincode

p0lunin:
This is an initial step towards providing robust encoding/decoding of p2p messages. Here we encode `PeerMessage` as a length-prefixed blob, and provide the `::Custom` variant as a copied buffer.

Later we will extend it to a generic type over custom message `T` that gets decoded together with `PeerMessage`, without extra copying.


p2p/src/codec.rs
```diff
@@ -0,0 +1,268 @@
...
+        match item {
+            PeerMessage::Hello(u) => {
+                dst.put_u8(0); // Message type
+                const HELLO_MESSAGE_LEN: u64 = 2;
+                dst.put_u64(HELLO_MESSAGE_LEN);
+                dst.put_u16(u);
+            }
+            PeerMessage::Peers(p) => {
+                dst.put_u8(1); // Message type
+                let body = p.into_iter().fold(BytesMut::new(), |mut bytes, peer| {
```
oleganza, line 25: can we fold using the current `dst`, so we don't allocate another buffer? Also the code might be more straightforward: first write the length prefix, then each peer address.

p2p/src/codec.rs
```diff
@@ -0,0 +1,268 @@
...
+                dst.put_u8(2); // Message type
+                dst.put_u64(data.len() as u64);
+                dst.put(data.as_slice());
+            }
+        }
+        Ok(())
+    }
+}
+
+fn encode_peer_addr(peer: PeerAddr, buf: &mut BytesMut) {
```
oleganza, line 42: i'd put this function close to decoding function, so it's easier to review their consistency.

p2p/src/codec.rs
```diff
@@ -0,0 +1,268 @@
...
+                        return Err(io::Error::new(
+                            io::ErrorKind::InvalidData,
+                            format!("Unknown command: {}", command_type),
+                        ))
+                    }
+                }
+                self.state = DecodeState::Len(command_type);
+                self.decode(src)
+            }
+            DecodeState::Len(m_type) => {
```
oleganza, line 107: ~i don't see much value in cross-cutting of entities by factoring out a length prefix for multiple messages. E.g. for the Hello message there's now unnecessary length prefix because the message is of fixed-length. The only factored out piece we need is the tag for the message type, while each message type should bundle all the logic together. It may even make sense to have sub-Decoders for structuring this state machine cleaner.~

as i was writing the above thoughts, i began to see the value in having length prefix for the entire body, so we can simply wait once till all the bytes we need arrive, so we don't have to track all the intermediate states (which would be much nicer in a async/await function, but we'd need async fn in trait, which forces boxing of futures, so... nope).

p2p/src/codec.rs
```diff
@@ -0,0 +1,268 @@
...
+        m => Err(io::Error::new(
+            io::ErrorKind::InvalidData,
+            format!("Unknown message type: {}", m),
+        )),
+    }
+}
+
+fn decode_peer_addr(buf: &mut BytesMut) -> Result<Option<PeerAddr>, io::Error> {
+    if buf.is_empty() {
+        return Ok(None);
```
oleganza, line 160: I don't think this is locally correct (but seems to be correct within the scope of the module). This helper is supposed to be running in a context where we are expecting to have all the bytes that we need (like the socket address parsing helpers below), while `Ok(None)` means that we (1) need more bytes and (2) will correctly remember the state of already received bytes. But the whole point of having a big fat length prefix for all message types is that we don't have to remember all intermediate states while waiting for more bytes. Instead, we can just parse them and fail fast. Our messages are quite short and we are not feeding portions of messages to a downstream system, so mandating buffering for the entire message is not a design or performance concern.

oleganza: aha, it's Ok(None) in case we run out of bytes, which is when we finish parsing the list. So i guess, it's okay to leave it as-is.

p2p/src/codec.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+    let addr = read_socket_addr(buf)?;
+    let key = buf.split_to(32);
```
oleganza, line 163: `split_to` panics. So if the length prefix was intentionally bogus, we'd be crashing the app on user-supplied input. This is a DoS vulnerability. Need to check the remaining length of the buffer after we've checked the variable-length socket addr.

p0lunin: Oh, i moved in read_socket_addr this check. I'll move it here.

p2p/src/codec.rs
p0lunin: Since it is not clear at first sight, I'll move this check into `read_message_body`.

p2p/src/codec.rs
p0lunin: Done

p2p/src/codec.rs
p0lunin: Changed: now we reserve 4 bytes for length prefix, than write body, then write length prefix at reserved place.

p2p/src/codec.rs
```diff
@@ -0,0 +1,284 @@
...
+    let id = PeerID(PublicKey::from(CompressedRistretto::from_slice(
+        key.as_ref(),
+    )));
+    Ok(PeerAddr { id, addr })
+}
+
+fn read_socket_addr(buf: &mut BytesMut) -> Result<SocketAddr, io::Error> {
+    // We check if !buf.is_empty() in MessageDecoder::decode, so it have at least 1 byte
+    let ipv = buf.get_u8();
```
oleganza, line 187: I'd keep functions as robust locally as possible, even if we do overlapping checks - those are not expensive, but later allowing a crash may be devastating. Let's add a check for enough bytes.

To make all these checks not so mouthful, i'd introduce `fn check_length(&mut BytesMut, n: usize, label: &str) -> Result<(), io::Error>` that simply errs with UnexpectedEof and message "need {n} bytes for {label}, but only {buf.len} is available in the buffer". Then we can do one-liners like `check_length(buf, 1)?;` before any attempt to read.

p0lunin: Done.

p2p/src/codec.rs
```diff
@@ -0,0 +1,277 @@
...
+        ip, port, flowinfo, scope_id,
+    )))
+}
+
+fn check_length(buf: &mut BytesMut, len: usize, label: &str) -> Result<(), io::Error> {
+    if buf.len() < len {
+        Err(io::Error::new(
+            io::ErrorKind::UnexpectedEof,
+            format!(
+                "Expected {} bytes for {}`, but found {}",
```
oleganza, line 224: bad backtick

oleganza, line 224:
```suggestion
                "Expected {} bytes for {}, but found {}",
```

PR #432, opened 2020-05-02
CI: no optimizations for bench builds

oleganza:
This should make CI builds faster.


oleganza:
Moved `cargo bench` builds from 10 minutes down to 2 minutes.

PR #467, opened 2020-06-09
zkvm: rename mintime to locktime, remove maxtime

### LINKED ISSUE #454: zkvm: remove maxtime
We should probably remove the `maxtime` (expiration time) field from TxHeader. The reason for it in the first place was due to symmetry with `mintime` and ability to reliably kick out low-fee transactions from the mempool, so they don't bounce in the network indefinitely.

Unfortunately, the problem with maxtime is that it makes mempool logic more complicated and opens some DoS vectors. The more you want to avoid DoS, the less useful maxtime is, as nodes would require it to be farther in the future.

Bitcoin uses a simple principle "every tx accepted in mempool can be mined" (aside of intentional double spending attempts). This means, that tx that incurred the cost of relay to the network cannot just disappear w/o paying a fee. This principle greatly simplifies the mempool logic. Bouncing issue can be mitigated in two ways: from the UX perspective, CPFP (child-pays-for-parent) can be used; and to prevent DoS-y bouncing, kicked out TxIDs can be remembered in a bloom filter to avoid their reappearance and re-relay without appropriately increased fee (via CPFP).

oleganza:
Closes #454

### DIFF
--- a/accounts/src/lib.rs
+++ b/accounts/src/lib.rs
@@ -20,7 +20,7 @@ and then be able to obtain a utxo proof in order to be able to spend the funds.
 
     Confirms payment details.
     Selects utxos to cover the payment amount
-    Forms a transaction with maxtime=min(sender's exptime, receiver exptime).
+    Forms a transaction.
     Send back the ReceiverReply that allows constructing a contract ID.
 
                               ReceiverReply ------->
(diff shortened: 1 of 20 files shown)


oleganza:
Closing this as relative timelocks require extra state handling compared with dealing with maxtime. See #454.

